PR #8, opened 2015-04-29
Fixes for native binaries build system

aelmahmoudy:


n8fr8:
Thanks for this PR! Will review and let you know if I have any questions.

aelmahmoudy:
Thanks.

Could you explain the vercode numbering scheme that is used in Orbot now ? The reason is that we build two versions of Orbot, one for pre android-16 compile without PIE, and the other one compile with PIE). Hence we give the version compile without PIE a lesser vercode than the one compiled with PIE.

n8fr8:
When you say "we build", can you tell me a little more about that? Are you redistributing Orbot as part of another app, or under another brand? Just curious and excited to hear about your work.

Otherwise, yes, I hadn't considered your use of the vercode for this purpose. I will take a look at VLC.

aelmahmoudy:
Oh, sorry, I mean "we" on F-Droid repository

n8fr8:
Ah, thanks! The "royal we" as they say!

aelmahmoudy:
Erm, actually I meant F-Droid team, not just myself.

By the way, how could I build meek-client & obfs4proxy ? Is there some sort of cross-platform "go" compiler that I should use ?

n8fr8:
Yes, all the Go build information for PT's is here: https://github.com/guardianproject/pluto

I haven't integrated into our Makefile yet but I am planning on it as some point.

PR #11, opened 2015-06-10
The big start stop makeover

eighthave:
Ok, this is a mega pull request for sure, so I imagine there will be some discussion on it.  Its working well for me, but I'm sure there will be some kind of regression since it changes a lot.  The only thing I've noticed is that I can't get the % messages from tor to reliably show on the bottom of the main screen.  My guess is that it is something related to the incoming messages taking up too much time on the UI thread.

In any case, there are lots of comments in the commits themselves.  The overall gist of this is that the entire start/stop sequence of Orbot was reworked to make sure there is a single, clear point that is `OFF`, `ON`, `STARTING`, and `STOPPING`.  Then that is broadcast to any app that is listening.  Also, this adds the `ACTION_START` request for starting `TorService` in the background without any user action.  There is a pref to disable it.  `ACTION_START` also serves as a way to query Orbot for its status info, since it will reply with the status.


eighthave:
This is related to these things:
* https://github.com/guardianproject/NetCipher/pull/25
* https://dev.guardianproject.info/projects/orbot/wiki/Orbot_Intent_API
* https://dev.guardianproject.info/issues/3117
* https://dev.guardianproject.info/issues/1800

eighthave:
Also addresses: https://dev.guardianproject.info/issues/3612

n8fr8:
I haven't looked at the code but do we now report the actual ports being used, and not just the default constants?

eighthave:
yup, that's what I just added.  It was easy to do with the new structure

eighthave:
Ok, looked at it with fresh eyes.  I found better spots to announce `STATUS_STARTING` and `STATUS_ON`, and that let me drive the UI updates entirely from the `TorService` status messages.  This fixes the regressions we discussed in IRC today.

n8fr8:
Testing these updates now

n8fr8:
Even if I disable the background starts, it still starts in the background using the START_TOR intent from Lightning Browser.

n8fr8:
Also, it seems the Activity UI does not reconnect to the background service to get the last status now, so if the Activity is destroyed, but the Service is still active, when you return to the Activity it shows "OFF" even though it is actually "ON"

eighthave:
Ok, all fixed.  For the first, I added a check of the "Allow Background Starts" pref, and removed an errant `finish()`.  For the second, I made `OrbotMainActivity` ask `TorService` for status in `onResume()`.  More details in the commit comments.

I also added some logic to `OrbotMainActivity` to detect if the tor daemon is running but not `TorService`.  If that is the case, `TorService` is started.

n8fr8:
I noticed there was still a problem with the activity not getting the latest status from the service when it is recreated... so the Service is running fine, and the Daemon as well, but the activity is destroyed and recreated. To fix this, I added status to the data transfer broadcast events, to make sure whenever those come in, it passes the latest status... otherwise, do you think that this scenario should be handled some other way?

n8fr8:
I have also now added a requestStatus() method the activity can call no resume, that just prompts the service to send the status broadcast back out. This handles the case I described above. Will merge here, and push.

PR #15, opened 2015-07-09
Update OpenSSL to v1.0.2a to fix build for x86_64 arch.

aelmahmoudy:
This was done in previous pull that you accepted. But for some reason in commit 2f8c130f02528d358115ea07f05ea9d8a9a42d48 you reverted that change (although the commit message says otherwise)


n8fr8:
How strange. Sometimes git still baffles me.

PR #44, opened 2016-06-26
Added more exit country codes

SpotComms:


n8fr8:
I appreciate the contribution, however... there are not meaningful Tor exit node presence in most of these countries. There was a reason we chose a more limited list.

PR #46, opened 2016-08-24
Fix 'Transparent Proxying' (not) enabling in rooted devices after 'Request Root Access' checkbox changes

lucabertoni:
Hi,
I modified the code to let the checkbox "Transparent Proxying" be enabled or disabled after the checkbox "Request Root Access" changes its value.
Before the checkbox were enabled only if the flow went into the "if" statement, and this happend only if the device wasn't rooted.
Setting "Request Root Access" to "unchecked" before didn't disable "Transparent Proxying" checkbox, now it disable it.

Thanks,
Luca


n8fr8:
We no longer support root features in Orbot

PR #54, opened 2016-09-30
Use low priority notification

Nutomic:
The notification is nice to have, but I don't need to see it all the time. A low priority notification makes much more sense imo.


n8fr8:
Unfortunately many people like seeing the notification all the time, to ensure that Tor is actually running. Perhaps this could be implemented as a preference?

Nutomic:
Would also be good for me. I'm on vacation right now, but will look into it after.

PR #64, opened 2016-11-29
fix for bug https://github.com/n8fr8/orbot/issues/63

arrase:
fix for: https://github.com/n8fr8/orbot/issues/63


arrase:
ouch! .... there more problems to solve ....

    private void updateStatus(String torServiceMsg) {
    	if (torStatus == null)
    		return; //UI not init'd yet

arrase:
i think that solves the issue without big changes

arrase:
@n8fr8 consider this fix

n8fr8:
This fix was merged

n8fr8:
Thanks!

PR #80, opened 2017-08-03
Expose IsolateDestAddr

SpotComms:


n8fr8:
Looks good. Will merge and test for the next release.

SpotComms:
Cool. Just to clarify, if two apps are connected via the SOCKS port they'll end up on different circuits. But if two apps use the HTTP proxy, they'll end up on the same circuit. Yes?

n8fr8:
I don't think so. The HTTP proxy is just a front-end for the SOCKS proxy, so all behaviors will be consistent.

PR #83, opened 2017-08-11
Try to improve hidden service reachability during sleep

cheahjs:
The issue at the moment is that while the device is sleeping for long periods of time, it is possible for the HS to become unreachable as a result of Tor detecting a clock jump of length greater than `NUM_JUMPED_SECONDS_BEFORE_WARN` (100 seconds) upon waking up, which then closes all circuits. Another issue is that if the device was woken up by incoming network traffic, the device only stays awake for about a second before going back to sleep, which isn't enough time for Tor to rebuild the intro circuits, and thus the HS is no longer reachable until Tor is able to rebuild the circuits.

I attempt to improve this situation in two ways:

1. Increase `NUM_JUMPED_SECONDS_BEFORE_WARN` from 100 seconds to 600 seconds to avoid triggering the clock-jumped-close-all-circuits code every time the device wakes up from sleep.
2. Add a new command (`MARKCONNFORWAKELOCK`) and event (`WAKELOCK`) to the control port to allow Tor to synchronously signal Orbot to hold wake lock on behalf of Tor (since it isn't possible to hold a wake lock from native code). A wake lock is acquired at the start of a event callback, then released when libevent returns from its event loop when there are no active events. This prevents the device from sleeping when Tor still has work to do.

### DIFF
--- a/app/src/fullperm/AndroidManifest.xml
+++ b/app/src/fullperm/AndroidManifest.xml
@@ -22,6 +22,7 @@
     <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
     <uses-permission android:name="android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS" />
     <uses-permission android:name="android.permission.READ_PHONE_STATE" />
+    <uses-permission android:name="android.permission.WAKE_LOCK" />
 
     <application
         android:name=".OrbotApp"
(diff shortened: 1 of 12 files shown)


n8fr8:
Thanks, we are reviewing and discussing on the Tor project trac system here: https://trac.torproject.org/projects/tor/ticket/23289#ticket

h-2:
I tried to add to the ticket at torproject, but the registration CAPTCHA failed like 10 times so I gave up.

> Increase NUM_JUMPED_SECONDS_BEFORE_WARN from 100 seconds to 600 seconds to avoid triggering the clock-jumped-close-all-circuits code every time the device wakes up from sleep

I suggest using 900 seconds as that is the renewal interval for web-socket connections and some other things (maybe 910 just to be sure). If I understand the proposed changes correctly you could piggy-back on the wake-ups by other services if they happen within the interval, right?

PR #97, opened 2017-11-02
Expose *ConnectionPadding

SpotComms:
This exposes 'ConnectionPadding' and 'ReducedConnectionPadding' options as recommended by the Tor man page.

> This option should be offered via the UI to mobile users for use where bandwidth may be expensive.

It also enables 'ReducedConnectionPadding' by default as to save battery and keep idle data usage low.

### DIFF
--- a/orbotservice/src/main/java/org/torproject/android/service/OrbotConstants.java
+++ b/orbotservice/src/main/java/org/torproject/android/service/OrbotConstants.java
@@ -36,4 +36,7 @@ public interface OrbotConstants {
 
 	public final static String PREF_ISOLATE_DEST = "pref_isolate_dest";
 
+	public final static String PREF_CONNECTION_PADDING = "pref_connection_padding";
+	public final static String PREF_REDUCED_CONNECTION_PADDING = "pref_reduced_connection_padding";
+
 }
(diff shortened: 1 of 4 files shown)


orbotservice/src/main/java/org/torproject/android/service/OrbotConstants.java
```diff
@@ -36,4 +36,7 @@
 
 	public final static String PREF_ISOLATE_DEST = "pref_isolate_dest";
 
+	public final static String PREF_CONNECTION_PADDING = "pref_connection_padding";
+	public final static STRING PREF_REDUCED_CONNECTION_PADDING = "pref_reduced_connection_padding";
```
Unpublished: Must be **String**

orbotservice/src/main/java/org/torproject/android/service/TorService.java
```diff
@@ -672,6 +672,15 @@ private boolean updateTorConfigFile () throws FileNotFoundException, IOException
         extraLines.append("TestSocks 0").append('\n');
         extraLines.append("WarnUnsafeSocks 1").append('\n');
 
+	if(prefs.getBoolean(OrbotConstants.PREF_CONNECTION_PADDING, false))
+	{
+		extraLines.append("ConnectionPadding 1").appends('\n');
```
Unpublished: appends -> append

orbotservice/src/main/java/org/torproject/android/service/TorService.java
```diff
@@ -672,6 +672,15 @@ private boolean updateTorConfigFile () throws FileNotFoundException, IOException
...
+	}
+	if(prefs.getBoolean(OrbotConstants.PREF_REDUCED_CONNECTION_PADDING, true))
+	{
+		extraLines.append("ReducedConnectionPadding 1").appends('\n');
```
Unpublished: same

SpotComms:
@Unpublished I've fixed those, thanks!

PR #98, opened 2017-11-10
Add lzma and zstd

goapunk:
We started conducting some tests on Tor's data usage (https://code.briarproject.org/akwizgran/briar/issues/45) and while at it I added support for lzma and zstd to see if it makes a difference. I thought this could be useful here too.
(In case you already got something prepared on your own feel free to close this MR :) )
Signed-off-by: goapunk <noobie@goapunks.net>

### DIFF
--- a/external/Makefile
+++ b/external/Makefile
@@ -101,6 +101,8 @@ endif
 	assets assets-clean \
 	openssl-static openssl-static-clean \
 	libevent libevent-clean \
+	lzma lzma-clean \
+	zstd zstd-clean \
 	tor tor-clean \
 	polipo polipo-clean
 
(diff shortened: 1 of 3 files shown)


external/Makefile
```diff
@@ -216,14 +268,13 @@ tor/Makefile: tor/configure
 				--disable-asciidoc \
 				--enable-static-libevent --with-libevent-dir=$(EXTERNAL_ROOT) \
 				--enable-static-openssl --with-openssl-dir=$(EXTERNAL_ROOT) \
-				--disable-linker-hardening --disable-gcc-hardening --disable-tool-name-check \
-				--disable-lzma --disable-zstd
+				--disable-linker-hardening --disable-gcc-hardening --disable-tool-name-check --disable-systemd
```
goapunk: On Fedora I need to disable systemd support, otherwise it won't compile.

n8fr8:
Fanastic, will review and merge. FYI, the plan is to break out the building of tor into its own library project and gradle.

n8fr8:
Thanks again, and keep us posted on the data measurements!

goapunk:
>FYI, the plan is to break out the building of tor into its own library project and gradle.

Great idea! Making the Tor build process reproducible is one of our upcoming tasks. I'll keep you posted about any positive results.
>Thanks again, and keep us posted on the data measurements!

Will do :)

PR #105, opened 2017-12-09
Update version strings in help->about

syphyr:
libevent is version 2.0.22 (stable)
polipo is version 1.1.2 (final release)
OpenSSL is version 1.0.2k
Tor is version 0.3.1.9 (stable)


n8fr8:
So you think the solution is to just remove the version numbers, and just put the URLs? It is easier to maintain that way.

n8fr8:
Thanks!

PR #110, opened 2017-12-26
Add new preference: pref_open_proxy_on_all_interfaces

BjarniRunar:
If set, this configures Tor to listen on 0.0.0.0 instead of 127.0.0.1.

The primary use-case for this, is to share Orbot with tethered devices or others on the same LAN. This is particularly useful when trying to access the Internet using a mobile plan that restricts or blocks tethered devices.

Without this patch, I would have had no useful Internet access from my laptop this Christmas...!  Filed in the hopes it will prove useful to others as well. Apologies in advance if this is the wrong venue for this sort of thing. Thanks for Orbot!

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -128,6 +128,8 @@
   <string name="pref_entrance_node_dialog">Enter Entrance Nodes</string>
   <string name="pref_allow_background_starts_title">Allow Background Starts</string>
   <string name="pref_allow_background_starts_summary">Let any app tell Orbot to start Tor and related services</string>
+  <string name="pref_open_proxy_on_all_interfaces_title">Open Proxy on All Interfaces</string>
+  <string name="pref_open_proxy_on_all_interfaces_summary">Allow Wifi peers, tethered devices and anyone else who can connect to your IP, to access Tor</string>
 
   <string name="button_proxy_all">Proxy All</string>
   <string name="button_proxy_none">Proxy None</string>
--- a/app/src/main/res/xml/preferences.xml
+++ b/app/src/main/res/xml/preferences.xml
@@ -28,6 +28,12 @@ android:key="pref_allow_background_starts"
 android:summary="@string/pref_allow_background_starts_summary"
 android:title="@string/pref_allow_background_starts_title"/>
 
+<CheckBoxPreference
+android:defaultValue="false"
+android:key="pref_open_proxy_on_all_interfaces"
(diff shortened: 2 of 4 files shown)


BjarniRunar:
Addendum: I have gotten some questions from elsewhere about whether listening on 0.0.0.0 is desirable, or whether it would be better to scope this to only listen on selected addresses.

I decided against this for the following reasons:

1. Simplicity: this is a trivial, easy to review and understand change
2. UX: a single check-box is easier to use (and explain, and understand) than anything which would let the user select interfaces

Regarding security concerns:

This only exposes the SOCKS5 port, which in turn only grants access to the Tor network - which is public anyway. The risk that someone will abuse this should be roughly equivalent to the risk of running a Tor bridge node; it may cost you some bandwidth, but not much else. I am assuming that this part of Tor is battle-tested reasonably secure.

I do agree is unfortunate that we are listening on the mobile network's IP address, as that probably goes against user expectations, but I suspect the practical risks here are relatively minor. All the same, I deliberately tried to make the explanation on the check-box sound a bit scary so people would think twice before enabling this.

Potential future enhancements:

* Expose Polipo as well
* Add to the home-screen information about which IP addresses Tor is currently available on, to simplify configuration of clients. As it is, the user needs to figure out the IP of their Android device, which is probably beyond most users' capabilities (requires listening for network state changes).
* Figure out which IP address belongs to the cellular network and avoid listening on that, to limit the exposure somewhat (this requires not listening on 0.0.0.0 and probably restarting and reconfiguring Tor every time the Android network state changes).

n8fr8:
Thanks, I agree this is a helpful feature, with low risk.

BjarniRunar:
Hooray! I'm glad you liked it. :)

Malsasa:
@BjarniRunar hello, finally I find the transparent proxy feature origin here. I really think this feature could be really useful and I'm glad to find it worked on your system above. But it does not work on mine (specs. below), as I have enabled "Open Proxy on All Interfaces" on Orbot, then share hotspot from phone, but on connected laptop Firefox doesn't go through Tor (check.torproject.org says "Sorry"). I use unconfigured Ubuntu 19.04 on laptop and without dnscrypt-proxy. Please tell me if my report lacks something and I need your guidance. Thank you. 

- Orbot 16.1.2 (from F-Droid)
- Tor 0.4.1.5 (bundled)
- Coolpad E580 smartphone
- Not rooted
- Android 6.0
- Kernel version 3.18

PR #126, opened 2018-02-26
VPN App Selection UI Improvements

bitmold:
The old VPN App Selection UI's GridView (pictured below) was off centered and its cells would often overlap. Additionally, it's TextView's would sometimes be truncated at the bottom. 
old-grid-view

I've centered it and each app's TextViews 2 lines for the app name which works well for almost every application installed on my device. (Extremely long apps that a user wouldn't launch from the launcher use an ellipsis). 

This is what the UI looks like after these changes:
new-new-ui

Additionally, a ProgressDialog was used without any text. This caused it to appear uncentered. I replaced it with a ProgressBar (which is a View)  and it is centered (and I believe more in line with UI standards on Android). Pictured below is the before and after 

old-pb
new-pb

### DIFF
--- a/app/src/main/java/org/torproject/android/ui/AppManagerActivity.java
+++ b/app/src/main/java/org/torproject/android/ui/AppManagerActivity.java
@@ -11,13 +11,11 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.StringTokenizer;
 
-import org.torproject.android.BuildConfig;
 import org.torproject.android.service.OrbotConstants;
 import org.torproject.android.R;
 import org.torproject.android.service.util.TorServiceUtils;
 import org.torproject.android.service.vpn.TorifiedApp;
 
-import android.app.ProgressDialog;
 import android.content.Context;
 import android.content.Intent;
 import android.content.SharedPreferences;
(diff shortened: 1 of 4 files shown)


bitmold:
I just noticed switching from `ProgressDialog` to `ProgressBar` fixes a bug in the VPN app selection screen too. 

With the old `ProgressDialog` if you rotate the screen a ton of times while it is loading the app will crash with an `IllegalArgumentException` that looks something like this: `java.lang.IllegalArgumentException: View=DecorView@e4f8729[] not attached to window manager`

The error occurs because Activity is killed during screen rotation, however the `AsyncTask` responsible for displaying the `ProgressDialog` and loading the apps is still executing and contains a reference to the dead Activity (passed into the `ProgressDialog`'s constructor). This error goes away when we switch to the `ProgressBar` approach used in this PR since it is a member variable of the AsyncTask's outer class.

bitmold:
There is a screenshot on [Orbot's play store page](https://play.google.com/store/apps/details?id=org.torproject.android) of the VPN app selection screen:

image

If this PR were to be merged, it would be good to replace the screenshot with the UI improvements introduced in this PR.

PR #127, opened 2018-02-27
Some UI String Fixes

bitmold:
Things Addressed Here:

- The License dialog was read from the file without any newlines causing it to be crammed and have its hyperlinks 404.
- Some tiny string fixes (..,. appeared in some spots, and also Wi-Fi was referred to by an incorrect variation of Wi-Fi in a good number of places in the app)

The about dialog had a lot of broken URLs and words stuck together because it is partially populated by the [LICENSE](https://github.com/n8fr8/orbot/blob/master/LICENSE) file which was read in `OrbotMainActivity`'s `readFromAssets(Context, String)`. The issue here is that the file is read with `BufferedReader`'s [`readLine()`](https://docs.oracle.com/javase/7/docs/api/java/io/BufferedReader.html#readLine()) which terminates strings at a line feed or carriage return. The fix was to simply append a linefeed after every line the `BuferedReader` reads (these newlines are later transformed into `<br/>` elements when the About dialog is getting configured). 

It's easy to see what I mean via these screenshots:

license-bad

You can see in the first URL the "If" from the next sentence is squished against it causing users to have a 404 when they click on it.

license-better

With the linefeed fix we no longer have broken URLs and it's a bit easier to read too.

I also changed Wifi into Wi-Fi

### DIFF
--- a/app/src/main/java/org/torproject/android/OrbotMainActivity.java
+++ b/app/src/main/java/org/torproject/android/OrbotMainActivity.java
@@ -521,14 +521,14 @@ public class OrbotMainActivity extends AppCompatActivity
             .show();
         }
 
-    public static String readFromAssets(Context context, String filename) throws IOException {
+    private static String readFromAssets(Context context, String filename) throws IOException {
         BufferedReader reader = new BufferedReader(new InputStreamReader(context.getAssets().open(filename)));
 
         // do reading, usually loop until end of file reading
         StringBuilder sb = new StringBuilder();
         String mLine = reader.readLine();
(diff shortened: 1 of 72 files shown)


bitmold:
Additionally on the [play store](https://play.google.com/store/apps/details?id=org.torproject.android) there is a typo on the screenshot with for bridges. There is an ellipsis with a comma between the second and third period. I searched the project for that string, and it still occurs in the Dutch translation so I added that fix to this PR. 

It would be good to post a corrected screenshot on the playstore. 

image

PR #142, opened 2018-04-13
Revert "Update BUILD"

staneleigh:
Reverts n8fr8/orbot#137


bitmold:
I'm curious as to why this is being reverted, and from there why `/libs/x86/pdnsd` is even being generated if `/libs/armeabi/pdnsd` is used for both x86 and armeabi

goapunk:
This doesn't seem to make much sense, I made a copy&paste error in the original commit which was corrected by @bitmold . You'd either have to remove the whole bock (e.g. line 26-29) or leave it as it is....

PR #143, opened 2018-04-13
App sorting for TorifiedApps is case sensitive

bitmold:
Lowercase apps are buried at the end of the sorted uppercase apps. This matches the behavior used by every Android launcher that I've used so far. Aside from a smoother UX, it's possible that a user might think Orbot is broken/not trust it if they think that the app they want to use the VPN with is missing.

unsorted


n8fr8:
What if we use " compareToIgnoreCase" instead?

bitmold:
That's a good call, I did not know that method existed!

PR #160, opened 2018-05-19
Update BUILD Instructions for armeabi-v7a

bitmold:
There was nothing in the BUILD documentation file to describe zipping up the armeabi-v7a pdnsd shared library. 

Additionally, after zipping it, it was not ignored so I added it to `orbotservice`'s `.gitignore`


n8fr8:
I think we are going to move to not zipping them anymore. It seems to not be needed now, and the APK compression might actually work better if they are not already zipped.

bitmold:
Oh, that's cool and simplifies everything! Good to know!

PR #166, opened 2018-07-08
Build x86

herman2018:


n8fr8:
We now support x86 and ARM in the same release build.

PR #171, opened 2018-09-05
torrc: remove deprecated DNSListenAddress and obsolete WarnUnsafeSocks

Unpublished:
While replacing DNSListenAddress with DNSPort (which warns if listening on 0.0.0.0) I noticed
https://github.com/n8fr8/orbot/blob/a608a964964506e52878ef55e5b500fd54af90fb/orbotservice/src/main/java/org/torproject/android/service/TorService.java#L721-L722

which doesn't seem necessary as pdnsd 
https://github.com/n8fr8/orbot/blob/a608a964964506e52878ef55e5b500fd54af90fb/orbotservice/src/main/res/values/pdnsd.xml#L4-L31
already communicates between VPN network and device localhost therefore another direct connection between VPN network and DNS is not required/used.

@n8fr8 Can you double check?


n8fr8:
Yes, I can see how that might be redundant and not needed. I will test!

PR #209, opened 2019-05-08
Request To Opt Out of Personalized Learning on Text Inputs

bitmold:
Android allows you to **request** that the keyboard should not track user input by setting the IME option `EditorInfo.IME_FLAG_NO_PERSONALIZED_LEARNING`. 

IME's will not necessarily respect this, but perhaps a good FOSS one will. Google's popular Gboard **claims** to respect this. 

Anyway, if this flag is respected this minimizes the app's footprint on the device it runs on. 

There are two primary changes here: 

- The `SettingsPreferences` `Activity` iterates over all of its `Preference`s and automatically sets the flag on each instance of an `EditTextPreference`. 
- For all of the other text input (which takes place exclusively in dialogs) I've defined a subclass of `android.support.v7.widget.AppCompatEditText` that sets the flag called `NoPersonalizedLearningEditText`. I had to go through and then explicitly declare this View be used in each of the relevant dialog's XML file. This is less than ideal, but there were not many changes that need to be made.
- While searching for `<EditText>` XML elements, I found an unused layout file `layout_settings.xml` that had an `<EditText>` but since it seemed very old + was unused I deleted it.


app/build.gradle
```diff
@@ -67,11 +67,9 @@ android {
 }
 
 dependencies {
-    implementation 'com.android.support.constraint:constraint-layout:1.1.3'
```
bitmold: I'm not sure why this is in the diff since this was merged in via a previous PR. Sorry about this noise.

PR #213, opened 2019-05-12
Remove unused resources.

sisbell:
Removes unused resources. Also removes unused ids from layout file.


bitmold:
In addition to the reduction of APK size being worthwhile, I'm happy to see this done from the perspective of people new to hacking on Orbot having less material to sift through.

n8fr8:
Agreed!

bitmold:
word! ya that was why I opened #203 and #204 too

PR #222, opened 2019-06-04
Rename hidden services to onion services

### LINKED ISSUE #132: References to Hidden Services should be changed to Onion Services
From https://www.torproject.org/docs/onion-services.html.en:

> Tor makes it possible for users to hide their locations while offering various kinds of services, such as web publishing or an instant messaging server. Using Tor "rendezvous points," other Tor users can connect to these onion services, **formerly known as hidden services**, each without knowing the other's network identity. This page describes the technical details of how this rendezvous protocol works. For a more direct how-to, see our configuring onion services page.

pgerber:
As described in #132, the Tor Project has deprecated the use of the name *hidden service* and has adopted *onion service* as a replacement.

Closes #132

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -108,7 +108,7 @@
     <string name="hidden_service_request">An app wants to open hidden server port %1$s to the Tor network. This is safe if you trust the app.</string>
     <string name="found_existing_tor_process">found existing Tor process&#8230;</string>
     <string name="something_bad_happened">Something bad happened. Check the log</string>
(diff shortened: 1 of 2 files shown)


bitmold:
I am happy to see this getting done! 

The string with key `hidden_service_request` should be updated on your branch, it still uses "hidden" instead of "onion"

bitmold:
Also, I'm not sure how translations are handled, but this MR only effects English variants of Orbot; many non-Enlgish users would still see hidden services in their respective locales

n8fr8:
at some point we can do an outreach to our translation community (Localization Lab), and ask them to review all of these strings

PR #225, opened 2019-06-07
String resource missed from #222

bitmold:
The string resource `hidden_service_request` was missed in the "hidden service" to "onion service" rename in #222


pgerber:
FYI, I skipped `hidden_service_request` intentionally.

Mostly because:
a) changing text resource keys means there will be a missing translation (instead of a less appropriate *Hidden Service*) until the new key has been translated into all languages and
b) there are still many uses of *Hidden Service* within the code and I figured it made sense to just change all user-facing strings and leave any internal strings, variables, packages, etc. untouched.

Don't think there is a right or wrong. Let's just see what the maintainer prefers.

bitmold:
I didn't change the name of the key  (it still uses "hidden") but there was the phrase "hidden server" in the value of the string that I changed to "onion server"

pgerber:
My apologies, should have read your comment more carefully or checked what you actually changed.

Issue #237, opened 2019-07-01
Upgrade to Gradle 5.x


sisbell:
I'm unable to build with later versions of the JDK.  I get the following error

`Could not determine java version from '13'.`

I looks like we need to upgrade gradle version to 5.x for JDK versions 11+.

eighthave:
FYI, an alternate way to solve this, which lets you use older gradle versions, is to set `JAVA_HOME` to the one included in Android Studio, which is still Java8, e.g.
```console
$ export JAVA_HOME=/opt/android-studio-3.3/jre
```

PR #240, opened 2019-07-01
Separate Constants For VPN, TOR and MAIN_APP

sisbell:
Took Orbot constants and broke it apart based upon module usage. This will help with encapsulation and provided some additional cohesion for the VPN and the Tor Service modules.

Specifically, I introduced VpnConstants and VpnPrefs to contain Vpn specific information. The app also has a MainConstants, containing constants it needs.

I moved some methods from TorServiceUtils to VpnUtils to help with layering. The sharedPrefs method is duplicated but this assists with the separation for now.

The other thing to note is that the dnsPort is now set as a pref, rather than directly as a field on TorService. This removes a direct dependency.

Note this merge also includes #237 (Unfortunately I need this to build in my environment)

### DIFF
--- a/app/src/main/java/org/torproject/android/MainConstants.java
+++ b/app/src/main/java/org/torproject/android/MainConstants.java
@@ -1,0 +1,15 @@
+package org.torproject.android;
+
+public interface MainConstants {
+
+    //EXIT COUNTRY CODES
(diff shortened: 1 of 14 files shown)


sisbell:
I fixed merge conflicts with most recent master.

PR #241, opened 2019-07-07
Fixed crash I introduced with the NoPersonalizedLearningEditText

bitmold:


n8fr8:
hadn't tested that yet, thanks!

PR #247, opened 2019-08-06
Launcher icon was out of date on low resolution devices.

bitmold:


bitmold:
here's an example of the wrong icon being used in the app switcher

bad orbot graphic

PR #249, opened 2019-08-08
Conditional keystore.properties

bitmold:
Let project still build without a keystore.properties file. People cloning the repo don't need this file if they want to build Orbot locally.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -8,15 +8,19 @@ def keystorePropertiesFile = rootProject.file("keystore.properties")
 def keystoreProperties = new Properties()
 
 // Load your keystore.properties file into the keystoreProperties object.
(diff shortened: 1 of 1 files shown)


bitmold:
honestly i'm not sure if this is the correct approach but many of us are just trying to build Orbot and not sign it for release. at the same time, I'm not sure how this impacts the entities that **Are** trying to sign release builds, my suspicion is that this won't get in their way tho

bitmold:
one thing i should have clarified upon opening the PR is that the gradle sync completely fails without the file being present, even if it's not being used

n8fr8:
Thanks for this. We are just trying to put in a bit more automation / CI into our work, and using tools like fastlane for building and publishing split APK releases, etc, so we can make smaller and targeted APK releases. 

This does look good, and I'll tweak it as needed for releases!

PR #252, opened 2019-09-05
manually use `aapt singleCrunch` for reproducible builds

eighthave:
PNG crunching is not a deterministic process, especially the way aapt does it.  This makes the builds not reproducible.  The easy solution to this is to pre-crunch the PNGs and commit them to git.  This also uses exiftool to strip any metadata from the PNG.  This then disables running the crunch as part of the gradle build process.

https://medium.com/@duhroach/smaller-pngs-and-android-s-aapt-tool-4ce38a24019d


eighthave:
submitted here also https://trac.torproject.org/projects/tor/ticket/31650

eighthave:
The PNG changes should be quite low risk, I've run this process on a bunch of apps already with no troubles so far.  It does also disable aapt crunching in _build.gradle_, that's been safe so far in my experience on the apps where I've added that.

Issue #274, opened 2019-10-19
Minimum sdk version (16) conflicts with uiautomator library (required 18)


05nelsonm:
Getting an error upon build stating:

`ERROR: Manifest merger failed : uses-sdk:minSdkVersion 16 cannot be smaller than version 18 declared in library [com.android.support.test.uiautomator:uiautomator-v18:2.1.2]`

PR #275, opened 2019-10-19
fixes issue #274 by bumping minSdkVersion from 16 to 18

05nelsonm:


n8fr8:
We aim to support down to Android 4.1 / SDK 16 when possible. "Bumping" to support uiautomator tools is not a good reason to cut out the many less well off users in the world who need Orbot.

05nelsonm:
> We aim to support down to Android 4.1 / SDK 16 when possible. "Bumping" to support uiautomator tools is not a good reason to cut out the many less well off users in the world who need Orbot.

Ya know I thought of this lastnight and couldn't agree more. I was stuck in "get it building" mode instead of thinking more about the users for whom it serves a more free internet.

Will close.

PR #277, opened 2019-10-28
Bugfix, cleanup, and funding

eighthave:
* fix bug where `startForegroundService()` was still called even if `Prefs.persistNotifications()` was false

* increase modular separation of Orbot's _TorService_ from other aspects, like Prefs and _TorVpnService_.

* add _FUNDING.yml_ file to add a "Sponsor" button on GitHub

### DIFF
--- a/FUNDING.yml
+++ b/FUNDING.yml
@@ -1,0 +1,4 @@
+liberapay: GuardianProject
+patreon: guardianproject
+github:
+    - eighthave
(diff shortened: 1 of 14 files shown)


command32:
keepass as it is still has no.way to change a password without exposing it
to clipboard

n8fr8:
Much of this is good, though regarding Foreground Services, there were changes in O that require them, if an Activity is not in the foreground itself. This means onboot, or starting from an intent cal,l etc. With that in mind, I was going to remove the persist notification preference, as it really doesn't make sense from O onward.

eighthave:
Google says:
>   The system allows apps to call Context.startForegroundService() even while the app is in the background. However, the app must call that service's startForeground() method within five seconds after the service is created.

But if `Prefs.persistNotifications()` is false, then it looks like Orbot never calls `startForeground()`, which I think gives you an exception.  So in that case, it should only call `startForegroundService()` if `startForeground()` will definitely be called within 5 seconds.

n8fr8:
Ah right that is a bug then

PR #310, opened 2020-04-08
Fixes #300 Where Locale is set to Zulu when settings are opened

### LINKED ISSUE #300: context menu in Tor log in Zulu
When marking text in the Tor log that's shown when swiping left, the options "Copy", "Share" etc. are shown in Zulu as "Kopisha", "Yabelana" and so on.
aaa

bitmold:
The locale to the app was set to isiZulu (`zu`) when the Settings screen is opened even if the user didn't explicitly set another language. This is because of a line in `Languages.java` in a loop for each Locale: `config.locale = locale` is assigned to the Activity's Configuration object. 

This assignment is not necessary and occurs for each Locale supported by the app. The reason the locale is always set to isiZulu because `zu` is the last locale alphabetically included in Orbot. 

Bug #300 says that only context menu text is changed to isiZulu because Orbot doesn't have any isiZulu localization, so only the Android context menu changes. If more isiZulu localization was included in Orbot, other parts of the app would have been effected too. This bug is for every language supported in Orbot, any time you open the settings and do not explicitly set a locale.

This PR removes the assignment statement causing the bug, and also removes some code that isn't used by the app anymore.

### DIFF
--- a/app/src/main/java/org/torproject/android/settings/Languages.java
+++ b/app/src/main/java/org/torproject/android/settings/Languages.java
@@ -71,9 +71,8 @@ public class Languages {
         DisplayMetrics ignored = new DisplayMetrics();
         activity.getWindowManager().getDefaultDisplay().getMetrics(ignored);
         Resources resources;
-        Set<Locale> localeSet = new LinkedHashSet<Locale>();
+        Set<Locale> localeSet = new LinkedHashSet<>();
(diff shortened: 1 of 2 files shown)


n8fr8:
Wow, I really hope this is finally the solution to this hilarious bug. THANK YOU!

bitmold:
I'm confident it is :+)

Issue #289, opened 2019-12-16
Horrendous white-space bug keeps Orbot from loading bridges


ghost:
Hello
There is this bug that has been living happily in Orbot's code since it's earliest version and it is really, really surprising how it has evaded scrutiny right until the latest version.

Here's the short description: Orbot is very, very fragile when it comes to handling whitespaces that might exist in the bridge list fed into the "Bridges" Input box.

I'm almost certain that not a single one of Orbot's devs have *ever* tried using bridges **via email** -- the process is easy: 
  * log into your email, 
  * send "get transport obfs4" to "bridges@torproject.org"
  * when you receive the bridge list, Copy them all at one,
  * in Orbot, choose Settings -> Bridges, Paste the bridge list into the input box
  * restart Orbot, and watch as it **FAILS** to load the torrc file due to some unrecognised character!

I wish it was just that!

Orbot's "torrc" would also fail to parse in these conditions too:
  * when there is any whitespace after the '\n' separating the bridges; e.g:
  this would successfully parse:
  ```
obfs4 38.229.33.152:50614 AEDC774CF21732E86F233869469CD8E5EEFC7765 cert=pcaZ00YFYW72mrFAcXQAEzMJNxV+RhYMIDJHhhoGWdRzlQqKm3JE1xBR+Uis6bGqbAPaDQ iat-mode=1
obfs4 38.229.33.139:42141 4047D875CC00F1CDDFD432814AB4A2723BFC412D cert=b7ejF+izmtLL/fuMo0IMIjGCPmeUfB9B3uat/KyHGRldNWcoLZ50VNhcMZoETy0iHUfKSQ iat-mode=1
obfs4 13.81.48.139:43981 C128E91BCF541FE0F6323F51D6B3C67326E441F4 cert=GlYLwHazlIfTYpJETZRkjx2PNlcJAKajCV6CoqyR8/Xceph5EpstUK176KvTBJmaG+UnFA iat-mode=0
```

   ....... but this won't:
   ```
obfs4 38.229.33.152:50614 AEDC774CF21732E86F233869469CD8E5EEFC7765 cert=pcaZ00YFYW72mrFAcXQAEzMJNxV+RhYMIDJHhhoGWdRzlQqKm3JE1xBR+Uis6bGqbAPaDQ iat-mode=1

[...]

PR #311, opened 2020-04-09
Fixes bugs in #289 pertaining to parsing bridges

bitmold:
Fixes the problems addressed in #289 where you could possibly get an empty String when parsing custom bridges or Orbot would throw an IndexOutOfBoundsException. A deeper explanation of things can be found on the page for #289. 

One thing that remains uncertain is some commented out code in OrbotService that seems like it used to handle the parsing:
```
/**
for (String bridgeConfigLine : bridgeListLines) {
  if (!TextUtils.isEmpty(bridgeConfigLine)) {
  extraLines.append("Bridge ");
  extraLines.append(bridgeConfigLine.trim());
  extraLines.append("\n");
 }
}**/
```

This old parsing implementation would put all bridges into the torrc file VS the current implementation at the time of fixing #289 which randomly selects one of the custom bridges for some reason. @n8fr8 do you know why one bridge is selected instead of using all of them? It almost seems deceptive that a user could enter multiple bridges and only one of them would actually be used whenever Orbot is run. I'm pretty ignorant about how tor actually pulls bridges from the torrc file so maybe it's possible that tor would just end up selecting one at random anyway? 

Anyway, I'm quite confident that this fixes the parsing problems from #289

### DIFF
--- a/orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
+++ b/orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
@@ -25,8 +25,6 @@ import android.content.IntentFilter;
 import android.content.SharedPreferences;
 import android.content.pm.PackageManager;
 import android.database.Cursor;
-import android.net.ConnectivityManager;
-import android.net.NetworkInfo;
 import android.net.Uri;
 import android.os.Build;
 import android.os.IBinder;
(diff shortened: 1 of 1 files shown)


orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
```diff
@@ -1628,10 +1622,8 @@ else if (proxyPass != null)
 
                 if (bridgeList != null && bridgeList.length() > 5) //longer then 1 = some real values here
                 {
-                    String[] bridgeListLines = bridgeList.trim().split("\\n");
-
-
-                    int bridgeIdx = (int)Math.round(Math.random()*((double)bridgeListLines.length));
+                    String[] bridgeListLines = parseBridgesFromSettings(bridgeList);
+                    int bridgeIdx = (int)Math.floor(Math.random()*((double)bridgeListLines.length));
```
bitmold, line 1626: This random used to have the chance to cause an index out of bounds exception prior to this PR. I'm not sure though why we select one bridge instead of loading them all into torrc

n8fr8: This was meant for use with built-in bridges, to reduce the chance that a censor-state would block all the bridges at once if it saw traffic to them. It also reduces load on the built-in bridges, so that Tor users don't all connect to them in one fixed order. We need a smarter way to implement this than was done before, and this should really only be done with built-in bridges and not user entered ones, anyhow.

Issue #309, opened 2020-04-07
Implement MOAT as we have done with Onion Browser iOS


n8fr8:
https://github.com/OnionBrowser/OnionBrowser/issues/124

n8fr8:
You can see in the main activity, how we handle the result of scanning a bridge QR code, and call setNewBridges()

https://github.com/guardianproject/orbot/blob/master/app/src/main/java/org/torproject/android/OrbotMainActivity.java#L925

Getting access to the obfs4proxy binary that is used to run Meek, is a bit more complex. You need to use the TransportManager class as shown here:
https://github.com/guardianproject/orbot/blob/master/orbotservice/src/main/java/org/torproject/android/service/OrbotService.java#L608

More soon

bitmold:
Is anyone actively working on this? I have a lot of time right now and would love to help out with this

bitmold:
There's a way `fetchCaptcha()` in `MoatActivity` can cause the application to crash. 

If a `JsonException` is thrown in the `Response.Listener` for the request being built in that method the exception is caught and this call `displayError(e, response)`. In that method an `AlertDialog` is shown and if the `MoatActivity` is no longer running Orbot will crash. Since the network request happens asynchronously, the the `Context`may not necessarily remain, such as when an `Activity` is in the process of being destroyed. 

There's a really good explanation of this kind of crash here http://dimitar.me/android-displaying-dialogs-from-background-threads/  
```
2020-04-27 00:42:06.073 3877-3877/org.torproject.android E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.torproject.android, PID: 3877
    android.view.WindowManager$BadTokenException: Unable to add window -- token android.os.BinderProxy@e009c9a is not valid; is your activity running?
        at android.view.ViewRootImpl.setView(ViewRootImpl.java:907)
        at android.view.WindowManagerGlobal.addView(WindowManagerGlobal.java:387)
        at android.view.WindowManagerImpl.addView(WindowManagerImpl.java:95)
        at android.app.Dialog.show(Dialog.java:342)
        at androidx.appcompat.app.AlertDialog$Builder.show(AlertDialog.java:1009)
        at org.torproject.android.ui.onboarding.MoatActivity.displayError(MoatActivity.java:426)
        at org.torproject.android.ui.onboarding.MoatActivity.access$800(MoatActivity.java:60)
[...]

PR #318, opened 2020-04-22
Implemented MoatActivity

tladesignz:
Implements issue #309.

I updated to the latest dev environment.
Also fixed all warnings in OrbotMainActivity.

Please advice, if I missed something, esp. regarding strings. I only added original English for now. (Strings are mostly taken from Tor Browser desktop!)


n8fr8:
it functions! need some more UX/UI but great job!

Issue #324, opened 2020-04-28
Copy UX/UI from Onion Browser for bridge request, display


n8fr8:

n8fr8:
Both Bridge Configuration and Custom Bridges activities/views

tladesignz:
I didn't implement QR code acquisition from an image, yet, as that seems, that Zxing would be needed as a dependency for that. And I remember, that it's horrible as a dependency because you need to strip it down to get rid of the app parts.

Is there a better way or is there a cleaned Zxing library to use as a Gradle dependency?

I also moved the e-mail request part into `CustomBridgesActivity`, although Onion Browser doesn't have it. I should add it there, too, I guess.

Also, I didn't do a strings clean-up and no translation, as I'm not sure how the workflow is in that regard.

n8fr8:
We do it in Orbot from the main menu already by send an intent request to a third-party QR scanning app:
https://github.com/guardianproject/orbot/blob/master/app/src/main/java/org/torproject/android/OrbotMainActivity.java#L472

IntentIntegrator integrator = new IntentIntegrator(OrbotMainActivity.this);
            integrator.initiateScan();

We then look for the onActivityResult response here:
https://github.com/guardianproject/orbot/blob/master/app/src/main/java/org/torproject/android/OrbotMainActivity.java#L891

You can use this same flow for this

tladesignz:
@n8fr8, sure, I copied that already. I was talking about an image from the gallery, not from the camera. Anyhow, nevermind.

bitmold:
So what specifically are the requirements of this ticket. I don't have an iOS device so I'm not able to see the UI/UX of Onion Browser either, could that be posted here?

As for zxing, there's evidently an android core/core module in this repository https://github.com/zxing/zxing/releases I can look into trying to integrate that into Orbot at some point if y'all think that's worthwhile. I'd be annoyed if integrating taking a pic of a QR code/scanning one from the gallery significantly increases our APK size although I never cared much for the 3rd party integration either just because people might not have access to that app and would have to put their trust into something outside of Orbot. In any case this is probably worth revisiting just because "Barcode Scanner" (the zxing app) had its last release in February 2019 and there's no longer going to be any updates to it (this is described in the link above)

tladesignz:
This is mostly implemented, besides using QR codes from photos.
Not sure, if this should actually be open.
@n8fr8, what was your intention when reassigning this?

n8fr8:
[...]

PR #330, opened 2020-05-04
Fixed smaller issues with MoatActivity, added CustomBridgesActivity

tladesignz:
See comments on issues #309 and #324 for details.

When starting up cold into MoatActivity, I never seem to be able to get bridges. It works often, though, when I start Tor and switch to Meek manually, first.

It's hard to debug, unfortunately, because Moat is very slow over Meek and the service is kind of hostile to debugging, probably to slow down attackers.

Can you confirm that, or does it work for you?


n8fr8:
Works, and looks much better than before!

PR #341, opened 2020-05-18
Delete rs-rAR translation

nijel:
- I don't think there is Serbian used in Argentina. 
- The few existing strings seem to be actually Spanish, so it was probably meant to be es-rAR.


bitmold:
This is a nice catch! I also don't think it's needed but I'm not 100% certain. 

If this *isn't* needed though there's still a `values-rs-rAR/strings.xml` in the `app` project that needs to be deleted too.

nijel:
I've added the second file as well.

PR #345, opened 2020-05-22
Fixes #329 Tor Wont Fully Stop With VPN Enabled

### LINKED ISSUE #329: Orbot 16.2.0-BETA-3 tor-0.4.2.7 won't stop when VPN mode is on
As per title.

If I push the stop button Orbot gets stuck in "TorService is shutting down". The logs just reports INFO about channelpadding. 

However, this is not a huge problem. As a workaround, if I disable VPN mode and then push the stop button again, TorService shutdowns properly (the logs report "Orbot is deactivated").

I am running Orbot on Android 10, Pixel 3a, GrapheneOS.

bitmold:
I posted an explanation about the root of this bug on the issue page for #329. Essentially a `Service` can't be destroyed with an open `ParcelFileDescriptor` interface which is what `OrbotService` uses for Tun2Socks when VPN is enabled.

There's no other functionality here beyond the bug fix, but I modified code as I was working in related areas to improve readability & decrease code size:
- I consolidated duplicate UI code in `OrbotMainActivity` that is responsible for starting/stopping Tor.
- Removed unused imports 
- Changed inner classes like `Builder` to `VPNService.Builder`. This makes the link between the `ParcelFileDescriptor` and `OrbotService` a bit clearer.

Since turning on the VPN without tor on also turns on tor, I made it such that turning off tor with the VPN sets the VPN switch to the off position. **@n8fr8 i'm not sure if it would make more sense here to keep the VPN switch in the on position while turning off Tor...**

Before:
untitled

After:
untitled

### DIFF
--- a/app/src/main/java/org/torproject/android/OrbotMainActivity.java
+++ b/app/src/main/java/org/torproject/android/OrbotMainActivity.java
@@ -31,7 +31,6 @@ import android.view.Menu;
 import android.view.MenuInflater;
 import android.view.MenuItem;
 import android.view.View;
-import android.view.View.OnLongClickListener;
 import android.view.animation.AccelerateInterpolator;
 import android.widget.AdapterView;
 import android.widget.AdapterView.OnItemSelectedListener;
@@ -61,6 +60,7 @@ import org.torproject.android.service.TorServiceConstants;
 import org.torproject.android.service.util.Prefs;
 import org.torproject.android.service.vpn.VpnConstants;
 import org.torproject.android.service.vpn.VpnPrefs;
+import org.torproject.android.service.vpn.VpnUtils;
 import org.torproject.android.settings.Languages;
 import org.torproject.android.settings.LocaleHelper;
 import org.torproject.android.settings.SettingsPreferences;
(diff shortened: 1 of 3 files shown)


bitmold:
If we wanted to keep the switch activated we would just need to change line 216 in `OrbotMainActivity` from
`if (mBtnVPN.isChecked()) mBtnVPN.setChecked(false); // indirectly closes tun2socks interface`
to
`if (mBtnVPN.isChecked()) sendIntentToService(ACTION_STOP_VPN);`

I've tested it and have seen it working either way.

app/src/main/java/org/torproject/android/OrbotMainActivity.java
```diff
@@ -885,7 +889,6 @@ else if (request == REQUEST_VPN && response == RESULT_OK) {
         }
         else if (request == REQUEST_VPN && response == RESULT_CANCELED) {
             mBtnVPN.setChecked(false);
-            Prefs.putUseVpn(false);
```
bitmold, line 888: Got rid of this line because the event listener on `mBtnVPN` makes this preference setting call

orbotservice/src/main/java/org/torproject/android/service/vpn/OrbotVpnManager.java
```diff
@@ -446,23 +380,20 @@ private void startDNS (String pdnsPath, String torDnsHost, int torDnsPort, Strin
...
 
-    private boolean stopDns ()
-	{
+    private boolean stopDns () {
 		if (filePdnsPid != null && filePdnsPid.exists()) {
 			List<String> lines = null;
 			try {
 				lines = IOUtils.readLines(new FileReader(filePdnsPid));
 				String dnsPid = lines.get(0);
-				killProcess(dnsPid, "");
```
bitmold, line 456: I think not using `import static` on utility methods like this can really improve code readability

n8fr8:
Thanks for all of this excellent work!

Yes, keeping the VPN matters to some people, who are concerned about traffic leaks, while they may just be restarting tor with a new configuration. Now, Orbot's VPN isn't mean to be a leak protection service, though if you force the system-level always on VPN, I suppose you get that protection through there.

I will try both ways and see how it feels.

Issue #348, opened 2020-06-06
Can not find "TorVpnService" Service Class


SArsalanK:
Hey, I Tried to clone the project. I can not find Can not find "TorVpnService" Service Class

![Screen Shot 2020-06-07 at 3 20 37 AM](https://user-images.githubusercontent.com/2125947/83955617-057a6e00-a86e-11ea-9fd6-422178edc109.png)

bitmold:
Hi, this is because by default Android studio will try to build the `app-mini` build variant instead of the regular Orbot app. Recent code changes to Orbot broke the unreleased miniature app, but it's probably not what you were trying to build anyway.

bitmold:
At the top of your screenshot it says `app-mini` and you can change it to `app`.

SArsalanK:
> At the top of your screenshot it says `app-mini` and you can change it to `app`.

Thanks, It resolved the issue.

SArsalanK:
can you help me with https://github.com/guardianproject/orbot/issues/349#issue-632777273

Issue #323, opened 2020-04-27
 Orbot connects directly to raw.githubusercontent.com on startup 


n8fr8:
https://trac.torproject.org/projects/tor/ticket/34025

Hashik-Donthineni:
Hey, @n8fr8 You said you were going to provide a way to turn it off since folks in China might need it to get updates. Did you ditch that idea? If so, what's the plan to inform the said people about the updates.

bitmold:
I think that's a good idea, having the raw.githubusercontnet.com update check happen behind a preference that is initially set to off.

bitmold:
With this check removed, this ticket can be closed. This update check was never really update to date anyway as the `update.json` file pointed to a very outdated release of Orbot. I deleted the json file that `AppUpdater` was checking against just now

PR #305, opened 2020-02-27
Fixed the welcome screen being blank after rotation 

Hashik-Donthineni:
Fixed the welcome screen, when you open the app for the first time, is showing a blank screen when you rotate the device (change orientation). I test this on API 18, 28.  There are no issues nor crashes. I just saved the instances and restored them.

This fixes the issue #304 
Before: 
Before
After:
After

Issue #304, opened 2020-02-27
App's welcome screen showing blank screen (text is missing after orientation change)


Hashik-Donthineni:
The welcome screen, when you open the app for the first time, is showing a blank screen when you rotate the device (change orientation). 
Reference Image (All the slides are same):
![Screenshot_20200227-141135](https://user-images.githubusercontent.com/22850179/75426831-71922f80-596b-11ea-98ef-59c916d9ca57.png)

bitmold:
I think this can be closed with #305 being merged

bitmold:
I've tested this and text is no longer lost, but upon rotation you are always taken back to the first slide. I'm working on a fix for this right now

bitmold:
Both of these issues seem to still exist or were reintroduced :(

Hashik-Donthineni:
Hey @bitmold The text is going away again?

bitmold:
Yes, right now the onboarding flow will just close:
![close](https://user-images.githubusercontent.com/22125581/88683040-5e71c000-d0c1-11ea-8d25-2865204f94ce.gif)

I've been tinkering with this flow, and we don't actually need to do any complicated work to preserve activity state on rotation. You get the expected results by setting the `android:configChanges` property in the manifest for the `OnboardingActivity`:

```
        <activity
            android:name=".ui.onboarding.OnboardingActivity"
            android:configChanges="orientation|screenSize|screenLayout|keyboardHidden" />
```

This tells the system to not destroy/create a new activity when the device is rotated and it gives us the desired results we want. In fact, it lets us safely delete a bunch of code we have written!

bitmold:
@Hashik-Donthineni if you'd like please take a look at #373

Hashik-Donthineni:
Okay! thank you for working on this. But it's weird it reoccurring, It's supposed to work, are there recent complaints about this @bitmold ?

bitmold:
Not that I am aware of. Today/yesterday I've been going through older issues that are still open and trying to close out as many as possible. When revisiting this one I tested the code off of the `master` branch and found that this issue (or at least an issue similar to the one you initially reported with this ticket) still existed.

Hashik-Donthineni:
Can you please confirm that it's not working by testing? It's just that, I am doing everything that can be done to save the state, it's not possible to lose the state. What do you think?

bitmold:
[...]

PR #359, opened 2020-07-11
app-mini project builds again

bitmold:
The app-mini project wasn't building due to refactoring, when you load orbot in Android Studio the `app-mini` configuration is usually selected by default instead of `app`. It not building caused confusion for people cloning the repository for the first time (issue #348, among others).

Anyway, this applies the changes to `app-mini` project that were applied to `app` in commit ec47151a0c5e9bf789c38a9f36820031241b54f4 pertaining to the elimination of `VPNEneableActivity` and `TorVPNService` and `OrbotService` integrating into one `Service`.


n8fr8:
Thanks for getting this going again. We have some plans to get this app actually released soon, so maybe we can talk about that later this week.

bitmold:
Hey  ya id be absolutely delighted to help get that out pls pls pls keep me posted :+)

bitmold:
I think in order to be serious about releasing the mini app a very concerted effort would have to be made to make Orbot and Orbot Mini reuse a lot of the same code. When the `app-mini` module was reproduced into the repository, a lot of the code to get the UI and other features were just copied directly from the `app` module, but since then, the `app` module has changed significantly enough to make this very complicated. Modularizing the code between the two apps is certainly a big undertaking, but I think ensuring that there is parity between the two with bug fixes is less straightforward and an even bigger undertaking.

For example, since the introduction of the `app-mini` module in #310 a modification to `app`'s `org.torproject.android.settings.Languages` was made to fix the locale getting set incorrectly. This change was very small and since the file in `app` has diverged from the one in `app-mini` the bug still exists in `org.torproject.android.mini.settings.Languages` which causes the locale to be set incorrectly 
image 

A good start would be maybe making a third module, `orbot-core` or something that both `app` and `app-mini` build which contains files like the one above.

bitmold:
I just pushed a branch which is a proof of concept of something along these lines: 
https://github.com/bitmold/orbot/tree/app-mini-modularization

This introduces a new module `orbotcore` which `app` and `app-mini` both reference:

- I moved`Languages` and `LocaleHelper` [there](https://github.com/bitmold/orbot/tree/app-mini-modularization/orbotcore/src/main/java/org/torproject/orbotcore). 
- The code for `SettingsPreferences` was essentially identical between `app` and `app-mini` so I moved that into `orbotcore` too while remaining it to [`BaseSettingsActivity`](https://github.com/bitmold/orbot/blob/app-mini-modularization/orbotcore/src/main/java/org/torproject/orbotcore/BaseSettingsActivity.java) and making the class `abstract`. Subclasses need to implement a method `int getPreferencesXmlResource()` which specifies the Android XML preferences resource identifier that gets populated.
- Then in `app` and `app-mini` I inherited from that class with [`OrbotSettingsActivity`](https://github.com/bitmold/orbot/blob/app-mini-modularization/app/src/main/java/org/torproject/android/OrbotSettingsActivity.java) and [`OrbotMiniSettingsActivity`](https://github.com/bitmold/orbot/blob/cc3e1239bf08fd560699a1fa88377885b4e28f87/app-mini/src/main/java/org/torproject/android/mini/OrbotMiniSettingsActivity.java) which just implement that one method. 

Both apps build and run without error and there is now overall less code in the repository that needs to be understood/maintained. The Locale bug I mentioned in my last comment is now gone from `app-mini` since there's now one place to modify settings code for both Orbot apps.

bitmold:
Not sure where to document this experiment so I'll just post more on this thread:

- Moved some common ui elements to `orbotcore`: `NoPersonalizedLearningEditText` and `Rotate3dAnimation`
- Moved `OnBootReceiver` and `MainConstants` to `orbotcore`
- Ran IDE linting on `orbotcore` ... removed unused parameters that were probably leftover from a longtime ago and also made classes in the module use Java 8 lambdas over anonymous inner classes.

Both apps use the fragment `CustomSlideBigText` in their onboarding flows. This is another instance like the language bug above where the code has differed since `app-mini` was prototyped. In this fragment `app-mini` still has bug #304 and is more or less the same file as `app` without the fixes from #305

bitmold:
Also it should be noted orbot mini still connects to raw.githubusercontent which pertains to #323

Hashik-Donthineni:
@bitmold about #323 I guess we would need @n8fr8 input to proceed. I'll design an on/off switch for it, both for app and app-mini if he would like but seems like he's planning on ditching raw.github* idea as a whole.

bitmold:
On the branch I linked above took `AppUpdater` out of Orbot Mini for the reasons above and because it wouldn't even work, it pointed to the update information for the full Orbot app anyway

PR #365, opened 2020-07-18
Fixes #358, #364, #366 BridgeWizardActivity UI, Rotation + Network Bugs

### LINKED ISSUE #358: Bridges configuration fail
Galaxy 8S
Android 9
Whenever you try to change your current Bridge configuration to a different one once you select a different option and go back to the main screen Bridges are deactivated, if you go into Bridges settings again the option you last selected is not the one active, instead the app goes back to the "Custom Bridges" option.

### LINKED ISSUE #364: BridgeWizardActivity UI Disappears on Rotation
In The BridgeWizardActivity the top 3 options (connecting directly to tor, community servers and cloud servers) all execute a test to see if the corresponding host is reachable. During this process a message is displayed to the user indicating that the test is underway and then if it is successful or not. 

This screen:
image

During and after this process if the user rotates the device the all of these messages disappear. It needs to be investigated if the tests stop or if the TextView responsible for displaying this information just vanishes on rotation. 

 The TextView should remain visible and consistent no matter how many times the user rotates the device. This bug can be confusing if not potentially bad if a user thinks the absence of the status message indicates that the host in question was reachable. 

This bug I think can compound the confusion/strange behavior surrounding the BridgeWizardActivity that was reported in #358

bitmold:
The async `HostTester` will continue to update the UI in the BridgeWizard even after the device is rotated. On device rotation, the visibility and text values of the TextView `mTvStatus` are written to a `Bundle`. Furthermore, only one `HostTest` is allowed to run at a time, previously multiple were fired off each time a RadioButton was checked. This could confuse the user. Finally, network requests will not continue to be made to the hosts in question once the `BridgeWizardActivity` is closed or when it opens the `CustomBridgesActivity`.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -104,7 +104,10 @@
             android:name=".ui.onboarding.OnboardingActivity"
             android:configChanges="orientation|screenSize|screenLayout" />
 
-        <activity android:name=".ui.onboarding.BridgeWizardActivity" />
+        <activity
+            android:name=".ui.onboarding.BridgeWizardActivity"
+            android:configChanges="orientation|screenSize" />
+
         <activity
             android:name=".ui.onboarding.MoatActivity"
             android:windowSoftInputMode="stateHidden" />
--- a/app/src/main/java/org/torproject/android/ui/onboarding/BridgeWizardActivity.java
+++ b/app/src/main/java/org/torproject/android/ui/onboarding/BridgeWizardActivity.java
@@ -28,7 +28,8 @@ import java.net.SocketAddress;
 
 public class BridgeWizardActivity extends AppCompatActivity {
 
-    private static int MOAT_REQUEST_CODE = 666;
+    private static final int MOAT_REQUEST_CODE = 666;
+    private static final int CUSTOM_BRIDGES_REQUEST_CODE = 1312;
 
     private static TextView mTvStatus;
     private static HostTester runningHostTest;
@@ -37,6 +38,8 @@ public class BridgeWizardActivity extends AppCompatActivity {
     private RadioButton mBtObfs4;
     private RadioButton mBtMeek;
     private RadioButton mBtCustom;
+
+    private View mBtnConfgiureCustomBridges;
 
     private static final String BUNDLE_KEY_TV_STATUS_VISIBILITY = "visibility";
     private static final String BUNDLE_KEY_TV_STATUS_TEXT = "text";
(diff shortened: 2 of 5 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -101,7 +101,10 @@
         </activity>
 
         <activity android:name=".ui.onboarding.OnboardingActivity" />
-        <activity android:name=".ui.onboarding.BridgeWizardActivity" />
+        <activity
+            android:name=".ui.onboarding.BridgeWizardActivity"
+            android:configChanges="orientation|screenSize" />
```
bitmold, line 106: Setting `android:configChanges` here prevents the `BridgeWizardActivity` from getting destroyed and recreated upon device rotation. This means that the cancellation of any outstanding network requests in its `onDestroy()` will be bypassed when the device rotates since the user is still in the `BridgeWizardActivity`.

app/src/main/java/org/torproject/android/ui/onboarding/BridgeWizardActivity.java
```diff
@@ -72,31 +83,33 @@ public void onClick(View v) {
         });
 
         mBtObfs4 = findViewById(R.id.btnBridgesObfs4);
-        mBtObfs4.setOnClickListener(new View.OnClickListener() {
+        mBtObfs4.setOnCheckedChangeListener(new CompoundButton.OnCheckedChangeListener() {
```
bitmold, line 86: Changing this listener prevents requests from firing off when the user reselects an already checked RadioButton

bitmold:
Initially this pull request was just for #364 but it made sense to build off of it to tackle #366 and finally #358 as well. Since it's a little scattered here is an overview of all the changes made to `BridgeWizardActivity` here:

- rotating the device when testing a tor/bridge connection will continue to update the user on whether or not the status was set (#364) 
- changing from one bridge option to another or choosing to directly connect to Tor will stop the previous network request and fire off a new one, this way only one network request is ever updating the UI at the same time and it's the one that the user clicked on. (#366) 
- closing `BridgeWizardActivity` or opening up another Activity from within it will interurpt any ongoing network requests since they are no longer relevant (#366)
- the `BridgeWizardActivity will be set to the correct radio button when it is opened (#358) 
- clicking "Connect directly to Tor (Best)" won't display status information that pertains to bridges (#358) 
- `CustomBridgesActivity` has slight string updates (#358)

bitmold:
I am not ready for this to be merged, I want to test it a bit more...

PR #371, opened 2020-07-24
Users can't select Orbot itself through the VPN app manager

bitmold:
Removes Orbot from the list of apps to "torify" - clicking it won't do anything but confuse users. When it's clicked it shows up on the main screen's "Tor Enabled" apps view which I think especially confusing


bitmold:
_sorry for all the force pushing, had some git trouble 🤷‍♀️_

PR #373, opened 2020-07-28
Fixes #304 Onboarding text and slide position are preserved on rotation

### LINKED ISSUE #304: App's welcome screen showing blank screen (text is missing after orientation change)

bitmold:
Makes the system preserve the `OnboardingActivity` on device rotation so that we don't need to do any work to preserve state. This fixes a recurring bug where that Activity would close or be in the wrong state on device rotation:

fix

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -100,7 +100,10 @@
                 android:value=".OrbotMainActivity" />
         </activity>
 
-        <activity android:name=".ui.onboarding.OnboardingActivity" />
+        <activity
+            android:name=".ui.onboarding.OnboardingActivity"
+            android:configChanges="orientation|screenSize|screenLayout" />
+
         <activity android:name=".ui.onboarding.BridgeWizardActivity" />
         <activity
             android:name=".ui.onboarding.MoatActivity"
--- a/app/src/main/java/org/torproject/android/ui/onboarding/CustomSlideBigText.java
+++ b/app/src/main/java/org/torproject/android/ui/onboarding/CustomSlideBigText.java
@@ -15,36 +15,28 @@ import org.torproject.android.R;
 
 public class CustomSlideBigText extends Fragment {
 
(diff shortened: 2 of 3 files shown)


Hashik-Donthineni:
I agree, since onboarding activity isn't used often I guess there is no harm here using this flag.

bitmold:
Is there generally harm with using that flag?

Hashik-Donthineni:
Yeah, in general, we shouldn't do it unless we have a specific need, and we are handling the app being killed in the background by OS to free up space by ourselves. Please refer to this very informative [Stackoverflow answer](https://stackoverflow.com/questions/7818717/why-not-use-always-androidconfigchanges-keyboardhiddenorientation)

app/src/main/AndroidManifest.xml
```diff
@@ -100,7 +100,10 @@
                 android:value=".OrbotMainActivity" />
         </activity>
 
-        <activity android:name=".ui.onboarding.OnboardingActivity" />
+        <activity
+            android:name=".ui.onboarding.OnboardingActivity"
+            android:configChanges="orientation|screenSize|screenLayout|keyboardHidden" />
```
Hashik-Donthineni, line 105: I guess we don't need the keyboard, layout, screensize flags? we just want to save the state on orientation change.

bitmold:
Thanks for that Stack Overflow post ~ I agree I think here it's fine because the UI flow is rarely used and it simplifies a lot of our code/bugs.

app/src/main/AndroidManifest.xml
bitmold: I think screenSize and screenLayout are needed, but not keyboardHidden.

https://developer.android.com/guide/topics/resources/runtime-changes

PR #375, opened 2020-08-04
Removed "Tor-Enabled Apps" UI from devices < Lollipop 

bitmold:
The VPN is either on ("full device") or not so there's no need to show users running these devices stuff about "tor-enabled" apps ~ this is extra information that could confuse users running older versions of Android

Before:
image

After:
image

### DIFF
--- a/app/src/main/res/layout/layout_main.xml
+++ b/app/src/main/res/layout/layout_main.xml
@@ -1,342 +1,341 @@
 <?xml version="1.0" encoding="utf-8"?>
-<LinearLayout android:gravity="center_vertical|center_horizontal"
-	xmlns:android="http://schemas.android.com/apk/res/android"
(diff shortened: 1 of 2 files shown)


app/src/main/res/layout/layout_main.xml
```diff
@@ -1,342 +1,341 @@
 <?xml version="1.0" encoding="utf-8"?>
```
bitmold, line 1: The only functional change to this file is that I added an `android:id` to a `LinearLayout`. There are so many deltas because I took the opportunity to run this through Android Studio's XML formatter

bitmold:
I noticed other crashes on API 16 (and probably other < lollipop SDKs as well) that I am not used to seeing on more recent versions of Android while twiddling with this... will report later

PR #389, opened 2020-09-09
Fixes #73 only log debug information one time if preference is enabled

### LINKED ISSUE #73: Debug log logs info entries twice
Is it intentional that there are log info and log debug entries in the config when debug logging is enabled?

https://github.com/n8fr8/orbot/blob/5b255df92d8ddffb2b6941e2f1be1efc0714e8a4/orbotservice/src/main/java/org/torproject/android/service/TorService.java#L707

This ends up logging info messages twice in logcat.

bitmold:
-removes uneeded `Log info syslog` immediatelly followed by the overriding `Log debug syslog` from torrc 
-seems like we were listening too eagerly for logging statements and logging them preemptively

### DIFF
--- a/orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
+++ b/orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
@@ -39,7 +39,6 @@ import android.util.Log;
 import com.jaredrummler.android.shell.CommandResult;
 
 import net.freehaven.tor.control.ConfigEntry;
-import net.freehaven.tor.control.RawEventListener;
 import net.freehaven.tor.control.TorControlConnection;
 
 import org.apache.commons.io.FileUtils;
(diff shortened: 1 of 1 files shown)


orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
```diff
@@ -1152,18 +1142,6 @@ private int initControlConnection (int maxTries, boolean isReconnect) throws Exc
...
                         logNotice( "adding control port event handler");
 
-                        if (Prefs.useDebugLogging()) {
-                            conn.setDebugging(System.out);
-                            conn.addRawEventListener(new RawEventListener() {
-                                @Override
-                                public void onEvent(String keyword, String data) {
-
-
-                                    debug(keyword + ": " + data);
```
bitmold, line 1162: @n8fr8  This line here seemed to log each log statement, I deleted it because logging information seemed to come from syslog later on without us needing this code This ultimately fixes the bug

n8fr8: Yeah, I think I left that in accidentally, trying to fix the hang issue.

PR #390, opened 2020-09-10
First stab at modularizing shared code between app projects

bitmold:
- created an "appcore" module where shared code between the two app and proposed app mini projects can exist


n8fr8:
Looks, builds and works for me. I like breaking out these small bits to do migration to Kotlin as well.

Should I merge this, or do you want to do more work here?

n8fr8 (approved):

bitmold:
Ya i think it can be merged. As the requirements for orbot mini get a it more established there might be more/less that needs to be modularized than what is here right now, time will tell.

Issue #392, opened 2020-09-20
Hosted Services UI gets messed up on orientation change


bitmold:
The spinner for user/app services always gets reset to user services on orientation change.

I'm gonna get of rid of the spinner outright because it makes parts of the UI hard to read as well as fix this as part of #180

PR #395, opened 2020-09-24
Fixes #180, #392 Migration of Backup Features to Android Storage Access Framework (SAF) 

### LINKED ISSUE #180: Orbot crashes
There is a problem with Orbot on Android 5.1.1. When I try to backup hidden service application stops and it is impossible to save private keys. Can you fix this? The same version works well on newer Android (7.0). I tried older Orbot version and it does not work too, archive with keys is not saved (Android 5.1.1). My current version of Orbot is 16.0.5-RC-1. It would be good to have an opportunity to add private keys in Orbot settings to avoid such problem in the future.

bitmold:
There's a lot happening in this pull request...

- The backup/restore for cookies and onion services uses the Android Storage Access Framework on APIs 19+. This means:
   - that we only include the scary `WRITE_EXTERNAL_STORAGE` on APIs 16, 17, and 18. 
   - the app will be future proofed with regards to policy changes for how Android handles storage in Android 11+ (API 30+) https://developer.android.com/training/data-storage/
- On APIs 16, 17, 18 backup is still supported using the `WRITE_EXTERNAL_STORAGE` permission. Dialogs pertaining to selecting a backup were deleted from the project since a system dialog gets the job done.
- Many string resources pertaining to backups and dialogs were removed from the project
- All backup/restore dialogs follow dialog aesthetic guidelines. This also means we have a lot less code in the project. 
- The Floating Action Buttons in the app (all are found in the backup/restore flow) use an actual plus sign. 

GIFs and a more thorough breakdown of things to come...


bitmold:
I just tested backup and restore for client cookies and hosted services on API 16 and 30 and the functionality was expected!!

I shouldn't be pushing anymore code here unless you find something messed up in your review or testing @n8fr8 

Here is the new cookie backup/restore flow:
cookie_restore_new

Here is the new hosted service backup/restore flow:
backup_restore

The new cookie dialog and new hosted service dialogs:
cookie_create_new
hosted_create_new

n8fr8:
building and working well here. code looks good. will review a bit more, and merge soon.

PR #399, opened 2020-10-06
Fixes #398 don't show notification with empty text

### LINKED ISSUE #398: Empty Notification Text
Orbot shows an empty notification to the user that doesn't convey anything to the user and looks sort of suspicious. 

This happens because when the app is launched, `OrbotService` is automatically started which displays an empty notification with this line in `onStartCommand()`. 

```java
showToolbarNotification("", NOTIFY_ID, R.drawable.ic_stat_tor);
```

Getting rid of this line will remove the confusing notification, `OrbotService` will instead `showToolBarNotification` only when meaningful events occur concerning the tor connection itself, and not merely because an Android Service was initiated. 

The app can also be put into an inconsistent state where the user sees this empty notification when they first launch the app. Eventually they start and stop tor, and remain within the Orbot app. Since `OrbotService` will be stopped, the notification will be gone even though the user just saw the empty notification when tor wasn't running.

bitmold:
Addresses the issue described in #398 and cleans up code surrounding notifications in `OrbotService`

### DIFF
--- a/orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
+++ b/orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
@@ -22,7 +22,6 @@ import android.content.ContextWrapper;
 import android.content.Intent;
 import android.content.IntentFilter;
 import android.content.SharedPreferences;
-import android.content.pm.PackageManager;
 import android.database.Cursor;
 import android.net.Uri;
 import android.net.VpnService;
@@ -153,7 +152,6 @@ public class OrbotService extends VpnService implements TorServiceConstants, Orb
 
         if (Prefs.useDebugLogging()) {
             sendCallbackLogMessage(msg);
-
         }
     }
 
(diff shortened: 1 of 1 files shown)


n8fr8:
A major part of the reason we show a notification for Orbot is to have a Foreground Service, to ensure the background tor process is not killed. How does this account for that? Do we just remove the "empty" notification onStart(), but then when tor is actually started, there are still the other notifications with useful content?

bitmold:
>  Do we just remove the "empty" notification onStart(), but then when tor is actually started, there are still the other notifications with useful content? 

Yes that's exactly what happens here. The distinction is that instead of showing the notification when `OrbotService` is started, the notification is shown when `OrbotService` starts `tor` (which is when we start caring that our Foreground Service remains alive). Almost always this difference isn't noticeable since starting `OrbotService` usually corresponds with starting `tor`. However this is actually not the case when the app is first opened, in this scenario:

-  as `OrbotMainActivity` is created it starts an instance of `ObotService`. Since `OrbotService` is running without `tor` running the user would see the empty notification as they moved through the app/switched to other apps. 
-  at this point `OrbotService` is hanging around in an idle state, when the user clicks the onion button to establish a connection  `OrbotService` will run `tor`. At this point users would immediately see a notification that says "Obrot is starting" and eventually information about the connection bandwidth. 

What's interesting is that if the user deactivates Orbot, the empty notification will go away. Because stopping the connection in `OrbotMainActivity`'s `stopTor()` method issues a call to stop `OrbotService` as well (which brings down the notification). This leads to the app being in an inconsistent state where initially users would see the empty notification when tor is disconnected, but upon connecting and then disconnecting any number of times after that they would no longer see that notification. From this point if the user were to click the button and establish a new connection `OrbotService` would start back up and **very briefly** show that empty notification, but since it starts up **with the instruction to connect** it'll start `tor` and (from the user's perspective) immediately start showing the aforementioned notifications that are concerned with the tor connection.

bitmold:
What remains unclear to me is if there is any value in making sure `ObrotService` stays alive when `tor` is not actually in use. I'm pretttttty sure the answer to this question is no. But if there is actually value in keeping the idle `OrbotService` alive then:

- we *do* wanna keep that notification, but should make it not empty 
- we should address the point i raised in my last comment and instead of stopping `OrbotService` when the onion button is clicked, and instead just tell `OrbotService` to stop `tor`.

bitmold:
I need more time to think this through, actually... Will reopen in a bit..........

PR #402, opened 2020-10-16
Fixes #401 Only NEWNYM Through UI When Sensible

### LINKED ISSUE #401: Possible to Request A New Identity when there is no Tor Connection
This can be confusing/misleading users and serves to give them a false understanding about what's going on with tor under the hood of Orbot. 

In the Main Activity you can request a new identity by clicking the refresh button which spins the onion UI thing and displays "You have a new identity" or something to that effect (this happens even when the UI is greyed out because Orbot hasn't started)

Similarly, there's a "NEW IDENTITY" button in the notification that is showing so long as `OrbotService` is running, however clicking it often doesn't do anything.

bitmold:
- The "NEW IDENTITY" button in the notification is only visible when the user can click on it
- Clicking the refresh button in the main app gives you a new identity if you have one, does nothing if Tor is starting, and starts tor if it's off/stopping

### DIFF
--- a/orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
+++ b/orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
@@ -86,7 +86,7 @@ public class OrbotService extends VpnService implements TorServiceConstants, Orb
 
     public final static String BINARY_TOR_VERSION = org.torproject.android.binary.TorServiceConstants.BINARY_TOR_VERSION;
     private final static int CONTROL_SOCKET_TIMEOUT = 60000;
-    private static final int NOTIFY_ID = 1;
+    static final int NOTIFY_ID = 1;
     private static final int ERROR_NOTIFY_ID = 3;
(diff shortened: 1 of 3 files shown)


orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
```diff
@@ -223,64 +225,45 @@ private void createNotificationChannel() {
         mNotificationManager.createNotificationChannel(mChannel);
     }
 
-    @SuppressLint("NewApi")
+    @SuppressLint({"NewApi", "RestrictedApi"})
```
bitmold, line 228: This `RestrictedApi` thing is the call to `mNotifyBuilder.mActions.clear()` which IMO is fine to do. Normally you wouldn't need to clear actions because you wouldn't hang on to a notify builder object like we do here, but anyway ya I think it's fine tbh.

n8fr8 (approved):

PR #411, opened 2020-11-11
Switch to using IPtProxy for obfs4 and meek

n8fr8:
Instead of using the executable binary version of obfs4client, we are now switching to using IPtProxy (https://github.com/tladesignz/IPtProxy) which combines obfsclient and snowflake. This is also runs the PT in-process via goptlib / gomobile bind.

### DIFF
--- a/orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
+++ b/orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
@@ -79,8 +79,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeoutException;
 
-import info.pluggabletransports.dispatch.util.TransportListener;
-import info.pluggabletransports.dispatch.util.TransportManager;
+import IPtProxy.IPtProxy;
 
 public class OrbotService extends VpnService implements TorServiceConstants, OrbotConstants {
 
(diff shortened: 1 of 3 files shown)


orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
```diff
@@ -316,6 +314,11 @@ private void stopTorAsync() {
             sendCallbackStatus(STATUS_STOPPING);
             sendCallbackLogMessage(getString(R.string.status_shutting_down));
 
+            String bridgeList = Prefs.getBridgesList();
+            boolean useIPtProxy = bridgeList.contains("obfs3")|| bridgeList.contains("obfs4")||bridgeList.contains("meek");
```
bitmold, line 318: This variable declaration is written out twice in OrbotService. It should be a static function

orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
```diff
@@ -1381,62 +1374,55 @@ private StringBuffer processSettingsImpl(StringBuffer extraLines) throws IOExcep
...
+            if (bridgeList != null && bridgeList.length() > 5) //longer then 1 = some real values here
+            {
+                String[] bridgeListLines = parseBridgesFromSettings(bridgeList);
+                int bridgeIdx = (int) Math.floor(Math.random() * ((double) bridgeListLines.length));
+                String bridgeLine = bridgeListLines[bridgeIdx];
+                extraLines.append("Bridge ");
+                extraLines.append(bridgeLine);
+                extraLines.append("\n");
+                /**
+                 for (String bridgeConfigLine : bridgeListLines) {
```
bitmold, line 1407: Is this block comment still needed?

n8fr8: nope!

PR #419, opened 2020-12-10
Adding snowflake features from IPtProxy

n8fr8:
(also some previous commits from updating the CHANGELOG !)


orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
```diff
@@ -1414,7 +1429,10 @@ private StringBuffer processSettingsImpl(StringBuffer extraLines) throws IOExcep
             if (meekBridges)
                 extraLines.append("ClientTransportPlugin meek_lite socks5 127.0.0.1:" + IPtProxy.MeekSocksPort).append('\n');
 
-            if (bridgeList != null && bridgeList.length() > 5) //longer then 1 = some real values here
+            if (snowflakeBridges)
+                extraLines.append("ClientTransportPlugin snowflake socks5 127.0.0.1:" + IPtProxy.SnowflakeSocksPort).append('\n');
+
+            if (bridgeList != null && bridgeList.length() > 10) //longer then 1 = some real values here
```
bitmold, line 1435: This comment should be updating. I suppose this works but i wish we had a better way of checking for this

orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
```diff
@@ -326,12 +329,23 @@ private void stopTorAsync() {
 
     }
 
-    private static boolean useIPtProxy ()
+    private static boolean useIPtObfs4Proxy ()
```
bitmold, line 332: This should be renamed since it includes meek as well

bitmold (commented):
This looks good! Unfortunately i won't have access to my computer until later this weekend so I haven't tested it yet...

orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
n8fr8: Yeah, it is really bad. I will take a shot now at improving the whole approach.

orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
n8fr8: Agreed

Issue #370, opened 2020-07-21
Improvements to MOAT server access


n8fr8:
The MOAT server has its own direct Meek endpoint. We can use it to talk to it directly, without the need for Tor.

To do so, we should use the APT library, which is already built into Orbot. We can start up a Meek instance directly to a custom endpoint, like this:

https://github.com/guardianproject/AndroidPluggableTransports/blob/master/sample/src/main/java/info/pluggeabletransports/sample/SampleClientActivity.java#L94

Then we can read/write bytes to it to speak HTTP.

Let's see if we can get it working!

tladesignz:
@n8fr8, see the mentioned commit. This starts obfsproxy's meek_lite in a new thread (would need to move to a service, of course for production) and should prepare `Volley` resp. `HttpURLConnection` to use it as a SOCKS proxy.

No error messages can be seen, no response arrives. Can you have a look? Thanks!

PR #420, opened 2020-12-15
Issue #370

tladesignz:
Use Meek directly for Moat instead of going through Tor. Uses the new IPtProxy.


tladesignz:
Finally! Everything works, I can confirm. Meek bridge works again. Moat finally works via it's dedicated domain front.

n8fr8:
Tested and confirmed Moat and Meek both working. Moat lookup seems much faster now, as expected.

PR #438, opened 2021-02-03
Removed Barebones Bridge Config in Orbot -> Settings

bitmold:
It's confusing to have two ways of editing bridges, with the "Custom bridge" wizard in the app it makes more sense to use that (also less resources to localize)


bitmold:
similarly moved the scanning and sharing of QR code bridges off the main menu since that functionality exists more clearly in the bridge wizard part of the app

PR #454, opened 2021-04-27
switch to native TorService

eighthave:
Switch Orbot to use to as a native Android `Service`.
This relies on changes to:

* jtorctl
* tor-android
* netcipher

The tor-android bits are not included in this merge request, you can build them yourself at https://gitlab.com/eighthave/tor-android or download the most recent debug build from the job artifacts of the GitLab CI builds: https://gitlab.com/eighthave/tor-android/-/jobs  For example: https://gitlab.com/eighthave/tor-android/-/jobs/1215266105/artifacts/download

@n8fr8 @bitmold


app-mini/build.gradle
```diff
@@ -100,8 +100,8 @@ android {
         exclude 'lib/arm64-v8a/obfs4proxy.so'
         exclude 'lib/armeabi-v7a/obfs4proxy.so'
         exclude 'lib/armeabi/obfs4proxy.so'
+        exclude 'META-INF/androidx.localbroadcastmanager_localbroadcastmanager.version'
```
bitmold, line 103: is this necessary?

app/build.gradle
```diff
@@ -109,8 +109,11 @@ android {
             universalApk true
         }
     }
-}
 
+    packagingOptions {
+        exclude 'META-INF/androidx.localbroadcastmanager_localbroadcastmanager.version'
```
bitmold, line 114: same with this?

orbotservice/build.gradle
```diff
@@ -41,11 +41,13 @@ android {
         lintConfig file("../lint.xml")
     }
 
+    packagingOptions {
+        exclude 'META-INF/androidx.localbroadcastmanager_localbroadcastmanager.version'
```
bitmold, line 45: and this

orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
```diff
@@ -916,142 +834,111 @@ private void updateLegacyV2OnionNames() throws SecurityException {
...
+        };
+        // run the BroadcastReceiver in its own thread
+        HandlerThread handlerThread = new HandlerThread(receiver.getClass().getSimpleName());
+        handlerThread.start();
+        Looper looper = handlerThread.getLooper();
+        Handler handler = new Handler(looper);
+        registerReceiver(receiver, new IntentFilter(TorService.ACTION_STATUS), null, handler);
+
+        Intent serviceIntent = new Intent(this, TorService.class);
+        if (Build.VERSION.SDK_INT < 29) {
```
bitmold, line 927: Curious as to why this is handled differently on < 29

orbotservice/src/main/java/org/torproject/android/service/TorServiceConstants.java
```diff
@@ -9,14 +9,6 @@
...
-    String TOR_CONTROL_PORT_FILE = "control.txt";
-    String TOR_PID_FILE = "torpid";
-
-    //torrc (tor config file)
-    String TORRC_ASSET_KEY = "torrc";
-
-    String TOR_CONTROL_COOKIE = "control_auth_cookie";
-
     //geoip data file asset key
     String GEOIP_ASSET_KEY = "geoip";
```
bitmold, line 13: are these still used?

orbotservice/build.gradle
eighthave: last I checked it was required because there was a conflict

app-mini/build.gradle
eighthave: last I checked it was required because there was a conflict

orbotservice/src/main/java/org/torproject/android/service/TorServiceConstants.java
eighthave: Looks like one of them is a duplicate

eighthave: Both of those constants were there already.  I updated this pull request to remove one

orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
eighthave: because that version of `bindService()` was added in _android-29_:
https://developer.android.com/reference/android/content/Context#bindService(android.content.Intent,%20int,%20java.util.concurrent.Executor,%20android.content.ServiceConnection)

The new version of `bindService()` lets you specify the `Executor`.  Since there is already an `Executor` instance there, it can be reused on 29+

n8fr8:
can we move this to use the mavencentral release?
    implementation 'info.guardianproject:tor-android:0.4.5.7'

eighthave:
This now includes _tor-android_ and _jtorctl_ v 0.4.5.7 so its ready to merge!

eighthave:
#455 and #456 should be merged first

PR #457, opened 2021-05-04
start migration away from jcenter and gpmaven repos

eighthave:
This enables Maven Central and allows jcenter and gpmaven only for libs that are only on those repos. Ultimately, those libs will have to be moved to Maven Central or replaced.

* jcenter is shutting down in February
* Custom repos like the gpmaven one are not allowed by F-Droid.org since they are a conduit for any kind of code to be included in the app. 

@n8fr8 @bitmold


build.gradle
```diff
@@ -15,9 +29,21 @@ buildscript {
...
     repositories {
+        mavenCentral()
         google()
-        jcenter()
-        maven { url "https://raw.githubusercontent.com/guardianproject/gpmaven/master" }
         maven { url 'https://jitpack.io' }
+        maven {
+            url "https://raw.githubusercontent.com/guardianproject/gpmaven/master"
+            content {
+                includeGroup('info.pluggabletransports.aptds')
```
n8fr8, line 38: we no longer need this since we are using iptproxy from jitpack

eighthave: There is still this from that repo:

```gradle
implementation 'info.pluggabletransports.aptds:jsocksAndroid:1.0.4'
```

eighthave:
@tladesignz turns out only v1.2.0 of Volley is on Maven Central, would that work for your Moat code?

tladesignz:
> @tladesignz turns out only v1.2.0 of Volley is on Maven Central, would that work for your Moat code?

Shouldn't be a problem.

eighthave:
as in, just make the change now in this merge requests, and breakage should be 
readily apparent during the next test/release cycle?

tladesignz:
As in: The change log doesn't look like any changes are tangent to our usage at all:
https://github.com/google/volley/wiki/Release-Notes

I had it switched locally, already. Should compile without a problem.

eighthave:
Ok, switched to v1.2.0 from Maven Central.  This needs to be merged after #456 and #454

PR #462, opened 2021-05-25
Orbot changes for new verison of IPtProxy that dynamically picks ports

bitmold:
Pertains to this issue https://github.com/tladesignz/IPtProxy/issues/7


n8fr8:
Is this built against a released version of the IPtProxy aar. or is this still against an experimental branch?

bitmold:
It's built against the experimental branch which actually slightly changes the API of IPtProxy so this will need to be merged after that is released.

PR #483, opened 2021-06-30
Actually enable AvoidDiskWrites (again)

SkewedZeppelin:
Added in 18682e4b974b0dff0120c9ae48cadb597387925f
Enabled in 410ae032564e0a57b3df20391b7023085087d628
Disabled in e78c27bfb0db5cf7e4ea18ff23645e432bd00224
Enabled in 74fce069df64b5cd35003c8bad75cc0f0a2622d2
Disabled in d6ebcd7ef7771fa0ef42e3b7271e131eec246c69

Why has this not just been left on? If it is useful for debugging, put it behind a variable maybe?

### DIFF
--- a/orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
+++ b/orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
@@ -536,7 +536,7 @@ public class OrbotService extends VpnService implements TorServiceConstants, Orb
         extraLines.append("\n");
 
         extraLines.append("RunAsDaemon 0").append('\n');
-        extraLines.append("AvoidDiskWrites 0").append('\n');
+        extraLines.append("AvoidDiskWrites 1").append('\n');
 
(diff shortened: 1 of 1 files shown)


n8fr8:
There were different opinions on how to might be impacting performance and stability. We will make a decision and document it, as well as add a setting option for it.

n8fr8:
AvoidDiskWrites 0|1
If non-zero, try to write to disk less frequently than we would otherwise. This is useful when running on flash memory or other media that support only a limited number of writes. (Default: 0)

I think we want it on 0 because that is the default, and flash media in this context doesn't have a "limited number of writes".

SkewedZeppelin:
https://source.android.com/devices/tech/perf/flash-wear
https://www.snia.org/sites/default/files/SDCEMEA/2020/6%20-%20Aviad%20Zuck%20Apps%20can%20quickly%20destroy.pdf

4-FLOSS-Free-Libre-Open-Source-Software:
> I think we want it on 0 because that is the default, and flash media in this context doesn't have a "limited number of writes".

All devices this app is running on have limited number of writes, both on eMMC or sdcards or SSDs, because this we want it on 1 as default.

PR #487, opened 2021-07-05
Remove old default bridge

cohosh:
Someone pointed out to us that this bridge line (which is an old Tor Browser default bridge) is still in orbot.

This was removed from Tor Browser with commit https://gitlab.torproject.org/tpo/applications/tor-browser-build/-/commit/e8b008d6eb90c8ea06a2562ecda9341ea7a380f5


n8fr8:
I suppose with Moat support, we may not even need to have built in bridges anymore?

PR #492, opened 2021-07-10
Disable pdnsd debug log

SkewedZeppelin:
This output is never read by Orbot and is only accessible to those who are rooted.

Alternative and preferred option for https://github.com/guardianproject/orbot/pull/491

https://github.com/guardianproject/orbot/commit/7cd5a2884d3d2980c92b323ad3191afe48736c38#diff-72e79c1e7b70841e61823714bd77d0abd6ed1166663fea3f6969b20bad5dc5e1R420


syphyr:
I can confirm that pdnsd.debug is no longer created with this commit.

PR #507, opened 2021-08-01
Fixes #505: Custom bridges Send to Email feature 

### LINKED ISSUE #505: Custom bridges "Send to Email" feature does not include Subject or Body message
Custom bridges "Send to Email" feature does not include Subject or Body message on **Pixel** Android 11. I believe this also affects **Samsung** devices as well. The subject and body text should be encoded as url params on the email address for this to work.

sisbell:


n8fr8:
This does fix the sending an email with the proper subject issue. Unfortunately, the response we get back from bridges@torproject.org isn't very useful... copy and pasting the bridge lines in directly doesn't seem to work, and the QR code isn't encoded with a proper bridgeqr URL format.... something to work on in another ticket.

sisbell:
I was able to import bridges successfully by opening the gmail client on my laptop, double clicking the QR code to expand all edges of the image, and then scanning the QR code.

However, I couldn't highlight the bridge fields in the gmail client for copying to Orbot. So this use case fails for me as well on Android.

n8fr8:
I will test scanning the QR code again. It didn't work for me.

I do think we should also add the bridgeqr:// format links into the bridges response email, so that people receiving emails on mobile phones can just tap the link to set the bridges.

Some history here:
https://gitlab.torproject.org/tpo/anti-censorship/bridgedb/-/issues/12639
https://gitlab.torproject.org/tpo/anti-censorship/bridgedb/-/issues/11345

Issue #501, opened 2021-07-31
Duplicate Config Info in Build Files


sisbell:
There is a lot of duplication of code in the build files. This makes it harder to manage dependencies and configurations.

PR #510, opened 2021-08-03
Use Gradle Version Catalog

sisbell:
The toml version file is in the gradle directory.  I realize this may require some discussion since it is a new way of doing things. I based this branch off of #501. Also cleaned out unused dependencies and added the compile time dependencies that were brought in transitively from the orbotservice module to the app modules.


sisbell:
I rebased the branch with the latest master and pushed the changes to this branch.

n8fr8:
Thanks for all this work in helping us clean up and organize the dependencies. The only issue I have with it is that the libs.versions.toml file doesn't show up in Android Studio for some reason. If we called it "dependencies.gradle" perhaps it would?

PR #511, opened 2021-08-08
Fix Issue #498 Missing VPN App Icons 

### LINKED ISSUE #498: Firefox does not show on the VPN list
Android 11
Galaxy 20

When I select Firefox from the VPN mode list it doesn't show on the Tor-Enabled Apps list.[
Screenshot_20210702-133102_Orbot
](url)

bitmold:
- Some apps will were not showing on the main Orbot Screen 
- This allows a variable number of apps to be shown (previously the UI would break and the apps would spill off of the screen) 
- moved a lot of UI logic from Java to XML too

The first commit of this PR fixes #498 but is not very performant. The bug had to do with occasionally inaccurate information being written to disk when users selected apps for "torification". The first commit completely fixes the bug, however has to perform somewhat expensive logic to get around how the incorrect data was being written to disk. 

The second commit makes the UI logic fast and introduces a change to Orbot where the correct data is written to disk. This will mean that users will have to reopen the app selection screen to see their apps on the main screen. It doesn't actually stop an app, say Firefox, that never appeared on the main screen from going through Tor, so I think it's worth committing.

### DIFF
--- a/app/src/main/java/org/torproject/android/OrbotMainActivity.java
+++ b/app/src/main/java/org/torproject/android/OrbotMainActivity.java
@@ -87,9 +87,12 @@ import java.net.URLDecoder;
 import java.text.NumberFormat;
 import java.util.ArrayList;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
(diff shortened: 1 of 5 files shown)


n8fr8:
Working well, thanks!

PR #514, opened 2021-08-10
Fixes #474 Remove Briar from Orbot VPN Routing + UI

### LINKED ISSUE #474: Exclude Briar from full device VPN
The Briar messaging app includes its own copy of Tor, which is unable to connect to the Tor network when Orbot is providing a full device VPN. Please exclude Briar from the list of affected apps when providing a full device VPN.

Ideally, please also exclude Briar from the list of Tor-enabled apps when VPN mode is enabled for selected apps, as selecting Briar in this list breaks Briar's Tor connectivity.

bitmold:
This introduces a compile time list of packages to ignore from VPN routing and the app selection UI

Packages can be added by modifying the list in `OrbotService`s `VpnPrefs`. 

I also removed Tor Browser and Tor Browser Alpha. This goes beyond the Briar in #474, but is likely something we want. I could imagine confused users getting both Orbot and Tor Browser and stumbling into a slow (among other issues...) Tor Over Tor situation.

### DIFF
--- a/app/src/main/java/org/torproject/android/ui/AppManagerActivity.java
+++ b/app/src/main/java/org/torproject/android/ui/AppManagerActivity.java
@@ -34,6 +34,7 @@ import org.torproject.android.R;
 import org.torproject.android.service.OrbotConstants;
 import org.torproject.android.service.util.Prefs;
 import org.torproject.android.service.vpn.TorifiedApp;
+import org.torproject.android.service.vpn.VpnPrefs;
 
 import java.lang.ref.WeakReference;
 import java.util.ArrayList;
@@ -54,9 +55,13 @@ public class AppManagerActivity extends AppCompatActivity implements OnClickList
     private ListAdapter adapterApps;
     private ProgressBar progressBar;
 
-    // returns true if the given app is enabled and not orbot
+    /**
+     * @return true if the app is "enabled", not Orbot, and not in
+     * {@link org.torproject.android.service.vpn.VpnPrefs#BYPASS_VPN_PACKAGES}
+     */
     public static boolean includeAppInUi(ApplicationInfo applicationInfo) {
         if (!applicationInfo.enabled) return false;
+        if (Arrays.binarySearch(VpnPrefs.BYPASS_VPN_PACKAGES, applicationInfo.packageName) >= 0) return false;
         return !BuildConfig.APPLICATION_ID.equals(applicationInfo.packageName);
     }
 
@@ -117,7 +122,7 @@ public class AppManagerActivity extends AppCompatActivity implements OnClickList
 
         final LayoutInflater inflater = getLayoutInflater();
 
-        adapterApps = new ArrayAdapter<TorifiedApp>(this, R.layout.layout_apps_item, R.id.itemtext, mApps) {
+        adapterApps = new ArrayAdapter<>(this, R.layout.layout_apps_item, R.id.itemtext, mApps) {
 
             @Override
             public View getView(int position, View convertView, ViewGroup parent) {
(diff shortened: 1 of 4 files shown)


n8fr8:
This is the error I get when I used this where I already have an allowed application set:

    java.lang.UnsupportedOperationException: addAllowedApplication already called
        at android.net.VpnService$Builder.addDisallowedApplication(VpnService.java:770)
        at org.torproject.android.service.vpn.OrbotVpnManager.doLollipopAppRouting(OrbotVpnManager.java:356)
        at org.torproject.android.service.vpn.OrbotVpnManager.-$$Nest$mdoLollipopAppRouting(Unknown Source:0)
        at org.torproject.android.service.vpn.OrbotVpnManager$2.run(OrbotVpnManager.java:294)

Should this code not run if I have chosen specific apps to proxy?

n8fr8:
-        for (String packageName : VpnPrefs.BYPASS_VPN_PACKAGES)
-            builder.addDisallowedApplication(packageName);
-
-        if (!perAppEnabled)
+        if (!perAppEnabled) {
             builder.addDisallowedApplication(mService.getPackageName());
+            for (String packageName : VpnPrefs.BYPASS_VPN_PACKAGES)
+                builder.addDisallowedApplication(packageName);
+        }

n8fr8:
Seems to work. Having issues committing it here though.

n8fr8:
fixed here: https://github.com/guardianproject/orbot/commit/482696b9b6442e9b636fb6c299bd2b67c8810df7

SkewedZeppelin:
Have those apps been tested to work when Android is set to "Block connections without VPN"?

bitmold:
@SkewedZeppelin thanks for pointing this out. 

Tor Browser, Tor Browser Alpha, and Briar are all normally added to the "disallowed list" when Orbot is in "full device VPN mode". 

Unfortunately this is a limitation of Android's "Blocked Connections" feature:
https://developer.android.com/guide/topics/connectivity/vpn#blocked_connections

> Caution: When non-VPN traffic is blocked, apps that aren’t in an allowed list or in a disallowed list lose their network connection. Consider warning people when making allowed or disallowed lists. To learn more, see the following Per-app VPN section.

Tor Browser, Briar, etc are programmatically added to the disallowed list. They are hardcoded into the Orbot, so the user never selects them, instead they are silently added to that list. What's also troublesome is that Android provides no way for an app to know if the "Blocked Connections" mode is enabled. If there was a way, Orbot could at least indicate to the user that Tor Browser, Briar, etc wouldn't work. 

This means that the only way to let those apps connect when the "Blocked Connection" feature is enabled is to route them through Orbot. This is generally unideal (routing Tor Browser through Orbot) and incidentally breaks Briar as per #474 

@n8fr8 this all needs to be reconsidered from a UX perspective. Here are some considerations:

Firstly, Orbot can fully opt out of Always On VPN mode, which would also opt it out of Blocked Connections. 
```xml
<service
            android:name=".service.OrbotService"
            android:enabled="true"
            android:permission="android.permission.BIND_VPN_SERVICE"
            android:stopWithTask="false">
            <intent-filter>
                <action android:name="android.net.VpnService" />
            </intent-filter>
            <meta-data android:name="android.net.VpnService.SUPPORTS_ALWAYS_ON"
                android:value="false"/>
        </service>
```

This seems undesirable since people use this feature. 


At the least, something explaining the limitation of Blocked VPN Connections can be included in Orbot's Settings and on https://orbot.app - Orbot would be unable to enable/disable blocked connections (the user has to manually do that), but clicking on it could at least open the system VPN settings for Orbot for the user:
image

Finally, there could be a setting in Orbot to force Tor Over Tor. This would allow Orbot users to have the "Blocked Connections" feature enabled *and* to route other tor apps through it. 

Unfortunately, all of these options are pretty bad. All of them also are bound to complicate users as well.

Issue #509, opened 2021-08-03
Use Gradle Version Catalog


sisbell:
Use Gradle version catalog for dependency management. 

https://docs.gradle.org/current/userguide/platforms.html

bitmold:
Closing because #529 was merged

PR #529, opened 2021-09-28
Add geoip library to libs.versions.toml

syphyr:
Fixes "Implements #509: Use Gradle Version Catalog"


syphyr:
I would also like to mention this issue: https://github.com/guardianproject/orbot/issues/479

I believe we should be using that same geoip databases as the torproject.

PR #535, opened 2021-10-27
Fix typo in dependencies.gradle

syphyr:
Fixes: move libs.versions.toml to dependencies.gradle makes it
       easier to work with in Android Studio


syphyr:
Accidentally left out libs.androidx_coordinator from app-tv dependencies

syphyr:
Fixed alphabetical ordering of app-tv dependencies. I don't think I missed anything else.

syphyr:
Fixed a few more spacing issues in dependency list on last cleanup commit.

PR #543, opened 2021-11-16
Fixed some random compiler warnings, while reading through the code.

tladesignz:


orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
```diff
@@ -425,6 +428,7 @@ protected void logNotice(String msg) {
         }
     }
 
+    @SuppressWarnings("ResultOfMethodCallIgnored")
```
bitmold, line 431: This seems like a mistake, `onCreate()` is void and I don't see this warning on my machine

tladesignz: This is for code *inside* #onCreate. Esp. #mkdirs.

Issue #540, opened 2021-11-02
Improve support for app services


mstniy:
Orbot does support app services, which is when an app requests Orbot to expose a given local port as an Onion service. However, that seems to be where it ends. Orbot does not provide a mechanism for applications to monitor or manipulate their Onion services, including querying the Onion domains they are assigned, making this feature useless.

Please correct me if I am mistaken.

mstniy:
The manifest contains a content provider, `org.torproject.android.ui.hiddenservices.storage` , but there doesn't seem to be any mention of it in the code.

Perhaps it is yet to be implemented?

n8fr8:
This feature wasn't necessarily intended to be an automated app API. The idea is that you run some kind of local server app, which is listening on localhost. Then, you as a user setup the app service within Orbot, which then gives you back the onion address in the UI for you to share, copy and paste, etc, as needed.

That said, there is an Intent UI for an app to request an onion service.

You can see an example of it here:
https://github.com/guardianproject/haven/blob/master/src/main/java/org/havenapp/main/SettingsFragment.java#L720
https://github.com/guardianproject/haven/blob/master/src/main/java/org/havenapp/main/SettingsFragment.java#L341

mstniy:
My understanding is the following: Orbot has two types of Onion services: _user services_ are created by the user, and work the way you mentioned.
_App services_ are created by other applications via intents, like [https://github.com/guardianproject/haven/blob/master/src/main/java/org/havenapp/main/SettingsFragment.java#L725](https://github.com/guardianproject/haven/blob/master/src/main/java/org/havenapp/main/SettingsFragment.java#L725).
The activity response used by [https://github.com/guardianproject/haven/blob/master/src/main/java/org/havenapp/main/SettingsFragment.java#L341](https://github.com/guardianproject/haven/blob/master/src/main/java/org/havenapp/main/SettingsFragment.java#L341) seems to have disappeared after commit guardianproject/orbot@0b1e7be203f30c5f3d99abfb018f34b99873d80a, which removed support for V2 services. The new intent that enables V3 services doesn't seem to return a similar response, thus the impossibility for the requester to obtain the domain name assigned to the service.

n8fr8:
Ah that is a regression. I missed that it was removed in the v2 removal work. @bitmold let's talk about that!

bitmold:
@n8fr8 @mstniy 
[...]

PR #556, opened 2021-12-14
Interface for #540

bitmold:
- External apps can request orbot to spin up a v3 service
- When cancelled, apps are informed
- When service is created, apps get the .onion url

Also introdued a change where onion service domains are updated when a
fresh connection tor tor is achieved, so users don't have to restart app

### DIFF
--- a/app/src/main/java/org/torproject/android/OrbotMainActivity.java
+++ b/app/src/main/java/org/torproject/android/OrbotMainActivity.java
@@ -8,6 +8,7 @@ import android.app.AlertDialog;
 import android.app.Application;
 import android.content.BroadcastReceiver;
 import android.content.ContentResolver;
+import android.content.ContentUris;
 import android.content.ContentValues;
 import android.content.Context;
 import android.content.Intent;
@@ -16,6 +17,7 @@ import android.content.SharedPreferences;
 import android.content.SharedPreferences.Editor;
 import android.content.pm.ApplicationInfo;
 import android.content.pm.PackageManager;
+import android.database.Cursor;
 import android.net.Uri;
 import android.net.VpnService;
 import android.os.Build;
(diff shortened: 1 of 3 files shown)


bitmold:
I published  a sample app that can be tested against this branch of Orbot

https://github.com/bitmold/OrbotRequestV3ServiceExample/blob/master/app/src/main/java/com/example/orbotrequestv3serviceexample/MainActivity.kt

Down the line we should do a new release of netcipher ensuring that all the v3 intent APIs are correct and remove the unsupported v2 ones...

syphyr:
With this change, the Orbot UI goes blank when I clear the Orbot activity from the Recents menu.


Screenshot_20211216-025614

syphyr:
This is a workaround to prevent orbot activity from disconnecting the tor service from the UI.  Even without the changes to the Interface, removing Orbot's activity from Recents causes  the tor service from not showing the download speeds in Orbot's UI.

https://github.com/syphyr/orbot/commit/b79238eefa0a57eaaac0091b3eedc80b88648e2b

bitmold:
Thank you for finding this, I'm just going to fix what about these changes introduced this bug. 

[I've seen Orbot get into this state in other ways too](https://github.com/guardianproject/orbot/issues/552), such as when power (dis)connection occurs at certain points when the program is running. We're trying to get this release out now, but my desires for the next one is to tackle all of these lingering/longtime-y bugs where the app can be forced into states in which it is broken 

But anyway, for the time being, I'm going to investigate this for the upcoming release.....

bitmold:
> Even without the changes to the Interface, removing Orbot's activity from Recents causes the tor service from not showing the download speeds in Orbot's UI.

So @syphyr , to clarify, the bug that you found isn't related to the code that's introduced here?

I've seen this bug sporadically, but have been unable to reliably reproduce it following what you wrote here: 

> With this change, the Orbot UI goes blank when I clear the Orbot activity from the Recents menu.

syphyr:
With this Interface change, Orbot thinks that the tor service is stopped when it's not.  I know what you mean about similar behavior when Orbot is swiped away in Recents, but with this patch it is different because when I start Orbot again, it says tor is stopped when it is not.

bitmold:
Ah. Well I'm very glad that you caught this, I'll be patching this today - thank you for your continued support, it's invaluable

syphyr:
I have to select the Onion twice to make the status change to green and show its connected because the first selection tries to start tor but it is actually stopping it at this point.

bitmold:
Thanks, this fixes it for the time being! 

@n8fr8 @syphyr I think this PR is ready, please update if you feel otherwise ...

bitmold:
the force push was a slightly more elegant implementation of this hotfix, but nothing serious

syphyr:
The status issue is fixed now. Thanks.

PR #557, opened 2021-12-14
IPtProxy 1.3.0

bitmold:
- Bumps IPtProxy to 1.3.0
- Optional preference to display snowflake toast when a client connects to proxy 
- Uses `"https://cdn.ampproject.org/"` as hardcoded amp cache 

This still needs to be tested...


bitmold:
This is ready for testing, to summarzie all of the changes here, since there are now a lot:

- IPtProxy 1.20 -> IPtProxy 1.3.0 
- Preference for displaying a Toast when a client uses your snowflake proxy to bypass censorship 
- Removes the broken "host tester" from the bridge wizard 
- Fixes a bug where snowflake proxy snowfall animation would appear if you connect to tor via a bridge (despite snowflake proxy never running in this scenario) 
- Reworked UI for using a snowflake bridge via domain fronting 
- Added UI for users to connect to a snowflake bridge via AMP rendezvous

PR #562, opened 2021-12-21
Fixes #398 empty notification

### LINKED ISSUE #398: Empty Notification Text

bitmold:
Orbot notification text is constantly wiped out, this only sets the notification text if:

- a notification isn't showing 
- and if it is, sets it to a "disconnected" message and not an empty string


n8fr8:
Looks good. Nice find!

PR #563, opened 2021-12-22
Battery bugs

bitmold:
storing this here for later ...


n8fr8:
so the crash was due to being unsupported on Android L?

n8fr8:
tested on Pixel Android 11 and Samsung S6 Android 7

PR #577, opened 2022-01-19
OrbotService Refactor 

bitmold:
nothing functional here, just refactored to make there be less code and make things more readable:

-I was frustrated that OrbotService defined constants that were already defined in `TorService`. We'd use one in some place, and some in the other. I removed all duplicate constants from `OrbotService` and now all usage of these throughout all Orbot apps point to the ones in tor-android's `TorService` 
- After this, `TorServiceConstants` contained stuff that only pertained to `OrbotService` and not `TorService` so I renamed it to `OrbotServiceConstants`
- After this, we were left with `OrbotServiceConstants` and `OrbotConstants` so I deleted the latter and moved all of the definitions that were defined there into the former.
- this targets Orbot, but also the mini and TV variants too


syphyr:
This patchset has a conflict with origin/raw_event_listener branch:
deleted by them: orbotservice/src/main/java/org/torproject/android/service/TorEventHandler.java

bitmold:
Thanks, I didn't expect this to be merged anytime soon. When other stuff gets merged I'll rebase this one to ensure it's conflict free

PR #582, opened 2022-01-21
Fix crash on startup

syphyr:
Changes to the expanded notifications have caused Orbot to crash on startup when testing with my tablet running cm-14.1.  Also, fix broken expanded notifications settings option.

### DIFF
--- a/orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
+++ b/orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
@@ -972,7 +972,7 @@ public class OrbotService extends VpnService implements TorServiceConstants, Orb
         if (logMessage.indexOf(LOG_NOTICE_HEADER) != -1) {
             notificationMessage = notificationMessage.substring(LOG_NOTICE_HEADER.length());
         }
-        if (mCurrentStatus.equals(STATUS_ON))
+        if (mCurrentStatus.equals(STATUS_ON) && Prefs.showExpandedNotifications())
             showToolbarNotification(notificationMessage, NOTIFY_ID, R.drawable.ic_stat_tor);
         mHandler.post(() -> {
             Intent intent = new Intent(LOCAL_ACTION_LOG); // You can also include some extra data.
--- a/orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
+++ b/orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
@@ -972,7 +972,8 @@ public class OrbotService extends VpnService implements TorServiceConstants, Orb
         if (logMessage.indexOf(LOG_NOTICE_HEADER) != -1) {
             notificationMessage = notificationMessage.substring(LOG_NOTICE_HEADER.length());
         }
-        showToolbarNotification(notificationMessage, NOTIFY_ID, R.drawable.ic_stat_tor);
+        if (mCurrentStatus.equals(STATUS_ON))
+            showToolbarNotification(notificationMessage, NOTIFY_ID, R.drawable.ic_stat_tor);
         mHandler.post(() -> {
             Intent intent = new Intent(LOCAL_ACTION_LOG); // You can also include some extra data.
             intent.putExtra(LOCAL_EXTRA_LOG, logMessage);


syphyr:
Fixes "719e387b (2 days ago) (2 days ago) - Notification title doesn't say Orbot, instead it shows the status - Notification text shows state as orbot is starting - Bandwidth transfer only shows once there's an active tor connection (b)"

bitmold:
Thanks for testing things so fast, @syphyr 

Could you explain a bit more what this PR does?

Firstly, none of the code in this PR seems to actually pertain to the new stuff that shows the Exit Node IP Address + Country (when available)

bitmold:
```java
  if (mCurrentStatus.equals(STATUS_ON) && Prefs.showExpandedNotifications())
            showToolbarNotification(notificationMessage, NOTIFY_ID, R.drawable.ic_stat_tor);
```

This code is specifically to show the bootstrap sequence when Orbot is starting up in the notification. Nothing here pertains to the expanded notification setting.

bitmold:
In fact, checking to see if the current status is `STATUS_ON` will not show the info during the bootstrapping process, which we want. The idea was to create a more responsive UI state. I'm looking into what you said about Orbot starting up on boot though...

syphyr:
It wasn't just Orbot starting up on boot.  The problem occurs whenever Orbot is started.

bitmold:
Thanks again, I'm looking into this now ...

syphyr:
Maybe it helps to debug knowing that without this change, Orbot's notification message is stuck on "set background service to FOREGROUND"

bitmold:
So I noticed that, with the latest code in `master`, if I start Orbot with the expanded notifications preference unset, it won't show the exit IP and country in my notification. I noticed however that if I start Orbot with the preference checked, then uncheck it while it's running, the last set Exit IP and country (if there was one) will linger and not be updated. 

Adding a small check to `setNotificationSubtext` in Orbot service fixes this, at least

bitmold:
> Maybe it helps to debug knowing that without this change, Orbot's notification message is stuck on "set background service to FOREGROUND"

This does help a lot, that message only gets logged if you are running on a device lower than Oreo

```java
   if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
            startForeground(NOTIFY_ID, notification);
        } else if (Prefs.persistNotifications() && !mNotificationShowing) {
            startForeground(NOTIFY_ID, notification);
            logNotice("Set background service to FOREGROUND");
        } else {
            mNotificationManager.notify(NOTIFY_ID, notification);
        }
```
 
@n8fr8 before this new beta goes out I think I need to look into this on an older device...

syphyr:
The Exit IP also disappears and does not appear to return when selecting "New Identity" from notifications.

bitmold:
> The Exit IP also disappears and does not appear to return when selecting "New Identity" from notifications.

This, however, is to be expected (at least for now...): when you select New Identity you do not immediately obtain a new identity. The expanded notification feature works opportunistically, when a circuit is built and a stream is opened with that circuit that isn't something tor uses internally, and has a SOCKS5 and isn't one attempting to reach an onion service, Orbot will then query Onionoo with the exit node's fingerprint (if it hasn't already) to try and obtain the country and exit IP. 

I guess what I'm saying is that with this feature, that information isn't immediately available.

bitmold:
Geez, yes thank you so much for finding this.

@n8fr8 a fix for this is coming very soon.

The issue is that on devices lower than Oreo, you get an infinite loop since it creates a notification to log that message from that method that creates notifications. 

I just fixed this in 0a0861e3566576e6adc841f8003d023a97728d0d so I'm closing this.

syphyr:
Just tested "0c6b07a1 (31 minutes ago) (31 minutes ago) added small UI check pertaining to #582 (b)" and it seems to have no effect on Nougat at least. I started Orbot with expanded notifications enabled, the exit node is displayed as expected.  Then I disabled expanded notifications while Orbot was still running and the exit node/IP was still displayed.

bitmold:
You need to pull again, the commit that properly fixes this came after that one

syphyr:
Thank you.  The crash on startup is fixed.  Although, toggling expanded notifications is still not working while Orbot is running.

syphyr:
Although, toggling expanded notifications works as expected if I stop and start Orbot between each change in the settings.  Thats fine with me. Maybe its just another issue with legacy devices.

bitmold:
> Although, toggling expanded notifications works as expected if I stop and start Orbot between each change in the settings. Thats fine with me. Maybe its just another issue with legacy devices.

This is the case right now. In the future the service can listen to preference changes and then tell the tor controller to listen / stop listening to the events it needs for this minor UI feature. It didn't seem like doing the work to implement this kind of responsiveness was really warranted for such a small feature, but yeah I concede that the overall UX does slightly suffer from this ambiguity.

PR #583, opened 2022-01-21
Bootstrap Progress in Notification; Remove Premature Connected Notification 

bitmold:
- Bootstrap progress is visually represented in the notification to provide a more responsive UX when tor is starting up 
- Connected to tor notification was launched prematurely. Usually this notification disappears after a minute, but on devices with a lot of network/processing latency it sticks around and is buggy.


bitmold:
Also did a quick UTF-8 thing with our response to Onionoo to display a country's flag emoji next to its display name

PR #604, opened 2022-02-11
Dev gotun2socks

bitmold:
Putting your work here so I can review it on my phone


bitmold:
One concern i have with this approach is the IPtProxy handling. I understand there's the gomobile limitation so we are tied with tun2socks needing to be in that project, but i don't like there being a prebuilt AAR in the app since it's tricky where it came from 

Maybe the fork of IPtProxy with the tun project could exist as it's own git repo and then Orbot could retrieve the AAR from there via jitpack? That way it's clear what's being downloaded and it would also be trivial to merge in new releases changes of IPtProxy

orbotservice/src/main/AndroidManifest.xml
```diff
@@ -1,6 +1,6 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="org.torproject.android.service">
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
-
+    <uses-permission android:name="android.permission.INTERNET" />
```
bitmold, line 4: I don't think this is needed

bitmold:
I think the most ideal approach to deal with the go dependencies is this:

1. Have IPtProxy as a submodule dependency in Orbot. Ignore the .aar and .jar files that are in the root of that repository 
2. Have your new tun go code live in a directory parallel to it as a second go module. 
3. Run Gobind when building Orbot with the two go module as input. This will produce an AAR and source jar. You can't have two AARs generated by Gobind, but can have one AAR that connects multiple go modules into Java. This AAR can then be used locally like the one on this branch here 

I like this approach the best. All the code is in the Orbot repository, there are no AARs, the Java symbols generated can be cleaner (no more having to have tun code under the IPtProxy package), and it becomes trivial to upgrade to new releases of IPtProxy

bitmold:
In the about dialog there's a String resource that says Orbot is dependent on badvpn. If you scroll down there is also the license for badvpn.

bitmold:
The build instructions in the readme can be updated to not include building the now deleted C++ code. Of course we will probably also have to add new instructions there too once an agreed upon approach to getting this go code into the project

n8fr8:
Thanks for the feedback. I agree having a prebuilt AAR in the project isn't great. 

I do have concern about how easy/hard it is for a new developer to get involved and get the code running. Requiring people to figure out the gomobile toolchain is a big ask.

Another approach is just to pushing our own orbot-iptproxy repo, with our modifications, and release that AAR in our gpmaven repo or maven central even. I think this is what other projects have done with their own flavors of go-tun2socks.

This means, someone who wants to hack on Orbot doesn't also need to know the gomobile side of things.

orbotservice/src/main/AndroidManifest.xml
n8fr8: just being superstitious !

bitmold:
I think it would be trivial/ not much different than having someone install and run NDK build. You install the go tool chain, set the environment variables and run a script to generate the AAR. 

I'm going to introduce this functionality after my bike trip i strongly am opposed to not having the AAR in the repo

bitmold:
Also, at the least, there doesn't need to be 3 AARs i think each app can point to the same file. This would just not bloat up git

bitmold:
> Another approach is just to pushing our own orbot-iptproxy repo, with our modifications, and release that AAR in our gpmaven repo or maven central even. I think this is what other projects have done with their own flavors of go-tun2socks.

I like this approach too

PR #623, opened 2022-03-01
Dev fix onboot

n8fr8:
another attempt to fix corrupt onboot VPN TUN issue


appcore/src/main/java/org/torproject/android/core/OnBootReceiver.kt
```diff
@@ -28,4 +33,31 @@ class OnBootReceiver : BroadcastReceiver() {
     companion object {
         private var sReceivedBoot = false
     }
+
+    private fun isNetworkAvailable(context: Context?): Boolean {
```
bitmold, line 37: in Kotlin you can just make the argument to this method of type `Context` instead of `Context?` and then remove the null check below

bitmold:
Is `ACTION_START_ON_BOOT` used anywhere anymore? if not, we can delete it too

appcore/src/main/java/org/torproject/android/core/OnBootReceiver.kt
```diff
@@ -28,4 +33,31 @@ class OnBootReceiver : BroadcastReceiver() {
...
+                    capabilities.hasTransport(NetworkCapabilities.TRANSPORT_WIFI) -> {
+                        return true
+                    }
+                    capabilities.hasTransport(NetworkCapabilities.TRANSPORT_ETHERNET) -> {
+                        return true
+                    }
+                }
+            }
+        } else {
+            val activeNetworkInfo = connectivityManager.activeNetworkInfo
```
bitmold, line 56: this could all be written as 
```kotlin
else if (connectivityManager.activeNetworkInfo?.isConnected`)  {
     return true
}
```

orbotservice/src/main/java/org/torproject/android/service/vpn/OrbotVpnManager.java
```diff
@@ -262,36 +259,55 @@ private synchronized void setupTun2Socks(final VpnService.Builder builder) {
...
+                }
+            }, 5000);
+
+        } catch (Exception e) {
+            Log.d(TAG, "VPN tun setup has stopped", e);
+        }
+    }
+
+
+    DataInputStream fis;
```
bitmold, line 301: Are there any reasons why these can't be delcared in `startListeningToFD`

Issue #638, opened 2022-03-18
Orbot integration through NetCipher OrbotHelper.java fails 


hussainmohd-a:
[RethinkDNS](https://github.com/celzero/rethink-app/issues/131) uses [NetCipher's OrbotHelper.java](https://github.com/guardianproject/NetCipher/blob/fee8571b1fb75c4164c5f4209565a6697152f8d8/libnetcipher/src/info/guardianproject/netcipher/proxy/OrbotHelper.java) to start and connect with Orbot's SOCKS5 and HTTP proxies. The [latest Orbot release v16.6.0-RC-4](https://github.com/guardianproject/orbot/releases/tag/16.6.0-RC-4-tor.0.4.6.9) has broken this integration.

More specifically, [`org.torproject.android.intent.action.STATUS`](https://github.com/guardianproject/NetCipher/blob/fee8571b1fb75c4164c5f4209565a6697152f8d8/libnetcipher/src/info/guardianproject/netcipher/proxy/OrbotHelper.java#L642) does not return expected answers.

In logcat I see:

```
 W/Orbot: unhandled OrbotService Intent: org.torproject.android.intent.action.START
 W/Orbot: unhandled OrbotService Intent: org.torproject.android.intent.action.STATUS
```

Let me know if I should file this with the NetCipher project instead?

n8fr8:
Looking into it. Thanks for letting us know, and frustrated we missed this in testing!

Issue #637, opened 2022-03-17
Orbot no longer broadcasts a status message if it is already running when it receives an ACTION_START intent


sorenstoutner:
In version 16.4.1-RC-2-tor.0.4.4.6 and previous, Orbot would broadcast a status message with the string "ON" if it was already connected when it received a "org.torproject.android.intent.action.START" intent.  Beginning with version 16.6.0-RC-2-tor.0.4.6.9 this is no longer the case.

It does continue to broadcast the "ON" status update after it connects if it isn't connected when it receives the ACTION_START intent.

This is a problem for [Privacy Browser](https://f-droid.org/en/packages/com.stoutner.privacybrowser.standard/), which uses the status broadcast to decide whether to display a "Waiting for Orbot" dialog.  It can be tested by enabling the Tor proxy in the settings.

sorenstoutner:
Privacy Browser has a bug report regarding the issue at https://redmine.stoutner.com/issues/814.

hussainmohd-a:
Related? https://github.com/guardianproject/orbot/issues/638

n8fr8:
Working on it

sorenstoutner:
Orbot 16.6.1-rC-3-tor.0.4.6.10 exhibits the opposite behavior.  If Orbot is not running and is started by another app with the "org.torproject.android.intent.action.START" intent, it starts and connects but never broadcasts the "ON" status update when it connects.

If Orbot is already running but not connected, it connects and correctly broadcasts the "ON" status updated.  If it is already connected it correctly broadcasts the "ON" status update.

sorenstoutner:
Does anyone have any ideas what it would take to fix this problem?

Issue #621, opened 2022-02-22
Progress bar when orbot is starting gets stuck sometimes


syphyr:
The progress bar when Orbot starts up often gets stuck with a message saying "starting orbot".  This happens very often if Orbot is chosen start up at boot.  But, this also occurs often when just starting and stopping Orbot.  

In OrbotService.java, it looks like the status bar is waiting for a LOG_NOTICE_BOOTSTRAPPED, but the issue is that often Orbot's log does not contain this message after Orbot starts up.  For example, when starting and stopping Orbot, the log message often gets truncated for some reason.

syphyr:
The exact message when it gets stuck is:
Orbot is starting... Connected to tor control port

Notice how when this occurs, bootsrapping is stuck on 5% in the log but tor appears to be connected.


![Screenshot_20220222-235722](https://user-images.githubusercontent.com/5009268/155233847-ed9ced24-5cd2-478b-838d-00fdf737d3d7.png)

syphyr:
Also, I see this in the logcat if I let Orbot sit a few mins when the notification is stuck:

```
02-22 23:57:34.279 28216 30865 F libc    : Fatal signal 6 (SIGABRT), code -6 in tid 30865 (tor)
02-22 23:57:34.397 30866 30866 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
02-22 23:57:34.397 30866 30866 F DEBUG   : LineageOS Version: '14.1-20220221-NIGHTLY-gts210vewifi'
02-22 23:57:34.397 30866 30866 F DEBUG   : Build fingerprint: 'samsung/gts210vewifixx/gts210vewifi:7.0/NRD90M/T813XXU2BQD3:user/release-keys'
02-22 23:57:34.397 30866 30866 F DEBUG   : Revision: '4'
02-22 23:57:34.397 30866 30866 F DEBUG   : ABI: 'arm64'
02-22 23:57:34.397 30866 30866 F DEBUG   : pid: 28216, tid: 30865, name: tor  >>> org.torproject.android <<<
02-22 23:57:34.397 30866 30866 F DEBUG   : signal 6 (SIGABRT), code -6 (SI_TKILL), fault addr --------
02-22 23:57:34.397 30866 30866 F DEBUG   :     x0   0000000000000000  x1   0000000000007891  x2   0000000000000006  x3   0000000000000008
[...]

PR #647, opened 2022-03-21
Tor Status Reworking 

bitmold:
This makes `OrbotService` follow the status updates from `TorService`, basically removing a lot of vestigial code from the time when `TorService` didn't exist and was a part of `OrbotService`.

In doing so, this fixes the issues reported in #637 and #638 in which Orbot, through the `Intent` logic in netcipher, won't report an `ON` status if it is already running. 

This also removes vestigial code from handling an old + unused start intent in `OrbotMainActivity`. 

I created a simple netcipher app to test this functionality: https://github.com/bitmold/OrbotIntentTest 
You can use that app to start Tor with netcipher. When you start Orbot with it, each status update is logged. When Orbot already has a tor connection, if you press the button in that app a 2nd, 3rd, etc time `ON` is logged.  

I tested this code with that app pretty extensively, as well as trying to weed out any corner case scenarios when Orbot starts on boot. Issues with the notification lock still exist, however with this new code the app UI will at least update to say that Tor is active (though the notification is still broken per #621 )

This code also removes a pretty annoying bug where when you kill Orbot via the main menu, the icon will disappear, reappear and then disappear for a final time. 

PS: @syphyr if you're looking to help with more Orbot testing here I'd definitely appreciate that :)


bitmold:
Been using this throughout the day and there are a few more tweaks that need to be made here

The newnym notification command opens the app and also there are occasional crashes in the UI with starting and stopping that i believe are a regression. Will post here as these get addressed .....

bitmold:
@n8fr8 this is looking good to me now, curious how your testing will go...

syphyr:
When I stop the tor service by selecting the onion, the screen says "Orbot is shutting down".  But it seems like it should eventually says that it "is shut down".

bitmold:
I don't experience this issue anymore @syphyr - @n8fr8 would you also be able to have a look at this with the latest changeset ..

bitmold:
# No, it's still occurring... I'll fix this when I can :'(

bitmold:
Actually this was a **very** simple fix, I was able to get it in today. Should be good to go as far as I can see!

PR #649, opened 2022-03-23
Remove Persistent Notification Preference + Logic 

bitmold:
Orbot, on Android 8+ is always a persistent service. Users running 7 and lower probably keep this preference on by default, and don't want the system shutting down Orbot without their approval. If users want to kill Orbot, they can. There was also a vestigial `DummyActivity` that we waste cycles starting and closing and has long since outlived its purpose. 

But ya, by removing this preference which is increasingly not even a choice for the user there's less to configure and less logic in our codebase


syphyr:
Tested this patch set on Android 7 for a day and there were no issues with the service getting killed.

PR #705, opened 2022-06-26
Reply with dns port alongside http, socks addrs

ignoramous:
Apps like RethinkDNS can forward DNS requests over TCP and UDP to any `port` on `localhost` (doesn't matter if Orbot doesn't support DNS over TCP, though, as all DNS requests and most DNS responses fit in one UDP packet more often than not).

Sending along DNS `port` with HTTP/SOCKS5 proxy info really helps power-user DNS apps that integrate with Orbot (like RethinkDNS) avoid adding special case just for Orbot (https://github.com/celzero/rethink-app/issues/284)


ignoramous:
cc: @n8fr8 / @4-FLOSS-Free-Libre-Open-Source-Software

Issue #708, opened 2022-07-07
Orbot ignores Isolate destination addresses


hd5pcvfx7frpqjab:
# Software & Hardware

- Device: Google Pixel 4a
- Operating System: GrapheneOS (SQ3A.220705.003.A1.2022070600)
- Orbot version: 16.6.2-RC-1-tor.0.4.7.8 (arm64-v8 release from GitHub)

# Steps To Reproduce

- Enable `Isolate destination addresses` in Orbot.
- Turn on VPN Mode (with `Always-on VPN` and `Block connections without VPN`) or use `127.0.0.1:9050` as a proxy. The bug is reproductible in both cases.
- Visit at least two websites that show IP address e.g. [browserleaks](https://browserleaks.com/ip) and [ifconfig](https://ifconfig.me).
- Verify that IP address in all cases is the same.

# Additional Informations

I checked that this issue is reproductible in Privacy Browser 3.10.1, Bromite 102.0.5005.96 and Vanadium 103.0.5060.71. Another user confirmed that issue is reproductible on DivestOS ROM (Android 11).

pgerber:
Reporting this was on my todo list as well. I looked at it a while back and it looked like the isolation was enabled on socks:

```
SocksPort 9050 IsolateDestAddr …
```

but it's missing for transparently proxied connections. It should be enabled there too:

```
TransPort 9040 IsolateDestAddr …
```

bitmold:
@pgerber You're right, adding `TransPort 9040 IsolateDestAddr` for `TransPort` will fix this issue. As others have said in #709 the SOCKSPort settings in torrc are ignored because `TorService` starts tor with the `--SOCKSPort` command line arg. I'm working on a fix for this now ... 

Thanks everyone for reporting here

bitmold:
Yes this issue wasn't fixed in the latest release. There needs to be a new release of Tor android and a new orbot release that's still in progress to fix this

Issue #707, opened 2022-07-07
Orbot ignores Torrc Custom Config for SOCKS + HTTP Ports


2ys7wz9roe3i74vk:
# Software & Hardware

- Device: Google Pixel 4a
- Operating System: GrapheneOS (SQ3A.220705.003.A1.2022070600)
- Orbot version: 16.6.2-RC-1-tor.0.4.7.8 (arm64-v8 release from GitHub)

# Steps To Reproduce

- Add `SocksPort 9051` to Torrc Custom Config in Orbot.
- Try to use `127.0.0.1:9051` as proxy in the browser and it will return `ERR_PROXY_CONNECTION_FAILED` error.

# Additional Informations

I checked that this issue is reproductible in Privacy Browser 3.10.1 and Bromite 102.0.5005.96. I also tried using different ports than 9051, but it didn't fix the issue.

Issue #709, opened 2022-07-08
Cannot disable SOCKS/HTTP proxy


7546bgenwnr58v5w:
# Software & Hardware

- Device: Google Pixel 4a
- Operating System: GrapheneOS (SQ3A.220705.003.A1.2022070600)
- Orbot version: 16.6.2-RC-1-tor.0.4.7.8 (arm64-v8 release from GitHub)

# Steps To Reproduce

- Set `Tor SOCKS` and/or `Tor HTTP` to `0`. According to Orbot settings, this should disable the proxy.
- Try to use `127.0.0.1:9050` and/or `127.0.0.1:8118` as proxy in the browser and it will still successfully connect even though the proxy should be disabled.

# Additional Informations

I checked that this issue is reproductible in Privacy Browser 3.10.1 and Bromite 102.0.5005.96.

SkewedZeppelin:
These are unconditionally set in TorService: https://github.com/guardianproject/tor-android/blob/master/tor-android-binary/src/main/java/org/torproject/jni/TorService.java#L341-L342

SkewedZeppelin:
> This rule is simple for options that take a single value, but it can become complicated for options that are
>        allowed to occur more than once: if you specify four SocksPorts in your configuration file, and one more
>        SocksPort on the command line, the option on the command line will replace all of the SocksPorts in the
>        configuration file. If this isn’t what you want, prefix the option name with a plus sign (+), and it will be
>        appended to the previous set of options instead. For example, setting SocksPort 9100 will use only port
>        9100, but setting +SocksPort 9100 will use ports 9100 and 9050 (because this is the default).

These three bugs are the same.

TorService always overrides the commandline, which is #708 and #709

And for #707 correct syntax would be prepended with a `+`

pgerber:
> And for https://github.com/guardianproject/orbot/issues/707 correct syntax would be prepended with a +

I'm not sure about that. Wouldn't a `--+SocksPort 9050` still override a `SocksPort 9050 IsolateDestAddr` in the config? It states, "prefix the option name with a plus sign (+), and it will be appended to the previous set of options". I'm not sure how "appending" would work if the same port was specified twice.

pgerber:
I decided to just test it, if you combine `--+SocksPort 9100` with a config file containing `SocksPort 9100` you get an error:

```
[...]

PR #715, opened 2022-07-20
Proxy Port Fixes 

bitmold:
This branch is dependent on https://github.com/guardianproject/tor-android/pull/84 being released in tor-android and should not be merged until that is released 
Fixes: #708 and #709

### DIFF
--- a/orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
+++ b/orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
@@ -514,12 +514,14 @@ public class OrbotService extends VpnService implements OrbotConstants {
             ipv6Pref += " IPv6Traffic NoIPv4Traffic ";
         }
 
-        extraLines.append("SOCKSPort ").append(socksPortPref).append(isolate).append(ipv6Pref).append('\n');
-        extraLines.append("SafeSocks 0").append('\n');
-        extraLines.append("TestSocks 0").append('\n');
-
-        if (Prefs.openProxyOnAllInterfaces())
-            extraLines.append("SocksListenAddress 0.0.0.0").append('\n');
+        if (!Prefs.openProxyOnAllInterfaces()) {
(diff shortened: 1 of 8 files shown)


bitmold:
Also fixes #707

orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
```diff
@@ -511,12 +514,14 @@ private File updateTorrcCustomFile() throws IOException {
...
-        extraLines.append("TestSocks 0").append('\n');
-
-        if (Prefs.openProxyOnAllInterfaces())
-            extraLines.append("SocksListenAddress 0.0.0.0").append('\n');
+        if (!Prefs.openProxyOnAllInterfaces()) {
+            extraLines.append("SOCKSPort ").append(socksPortPref).append(isolate).append(ipv6Pref).append('\n');
+            extraLines.append("SafeSocks 0").append('\n');
+            extraLines.append("TestSocks 0").append('\n');
+        } else {
+            extraLines.append("SOCKSPort 0.0.0.0:").append(socksPortPref).append("\n");
```
SkewedZeppelin, line 522: should isolate/ipv6 be added here too?

bitmold: yes, thank you. added

syphyr:
It looks like R.id.tvSnowflake is undefined

syphyr:
This fixes the problem: https://github.com/syphyr/orbot/commit/63c0060a38d980fbcab73eb11d28fd0976325803

bitmold:
Thanks @syphyr , I just forgot to commit the XML file. I'll throw it up next time I'm online

syphyr:
One other thing I noticed is that the version string is only showing up for obfs4 in "About" screen.

bitmold:
Odd, I was seeing both versions. Will revisit when I commit the missing XML file ...

SkewedZeppelin:
eh, https://github.com/guardianproject/tor-android/pull/84 was merged and released?

weren't these supposed to be together?

PR #728, opened 2022-09-04
onBootReceiver: Ensure user setup has been completed before starting FGS

theimpulson:
## Summary

This PR introduces an additional check for the `onBootReceiver` class before starting the `OrbotService` to ensure that the user's onboarding has been completed. This is required as if the user's onboarding hasn't been complete (which means the user never opened the app), the service will still be triggered.

## Test

Install artifact generated with this patch, reboot the device, and ensure FGS no longer gets triggered until onboarding is complete.


grote:
Also potentially relevant: https://developer.android.com/guide/components/foreground-services#loc-camera

PR #733, opened 2022-09-13
Allow app rotation

syphyr:
Remove default portrait orientation to also allow landscape mode


n8fr8:
Happy to merge, but this will definitely still require more work to make it useful/usable.

PR #747, opened 2022-10-18
V16 maintenance fixes

syphyr:
These are fixes for upgrading to gradle 7.3.3


syphyr:
These should also be cherry picked to master branch

syphyr:
The errors caused by upgrading to gradle 7.3.3 can be reproduced by using "./gradlew assembleRelease"

n8fr8:
Perhaps this approach doesn't cause the fail, but maybe the classes are not being packaged in the final APK?

syphyr:
This approach seems to only work with OrbotLib.aar and not with tor-android-binary-release.aar for some reason. I did not see this issue at first because I was testing out an updated tor-android.aar from my local maven repo instead. When I include the tor-android.aar without my local maven repo, I can confirm that the Orbot apk has all the libs included but fails at runtime exactly like you mentioned. Although, I am not sure why your builds are not failing like mine do with the previous method like this: 

``
> Task :orbotservice:bundleReleaseAar FAILED

> Task :orbotservice:verifyReleaseResources
/media/delta/T5/android/orbot-latest/orbot/orbotservice/build/intermediates/merged_res/release/values/values.xml:107: Multiple substitutions specified in non-positional format of string resource string/snowflake_proxy_client_connected_msg. Did you mean to add the formatted="false" attribute?

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':orbotservice:bundleReleaseAar'.
> Error while evaluating property 'hasLocalAarDeps' of task ':orbotservice:bundleReleaseAar'
   > Direct local .aar file dependencies are not supported when building an AAR. The resulting AAR would be broken because the classes and Android resources from any local .aar file dependencies would not be packaged in the resulting AAR. Previous versions of the Android Gradle Plugin produce broken AARs in this case too (despite not throwing this error). The following direct local .aar file dependencies of the :orbotservice project caused this error: /media/delta/T5/android/orbot-latest/orbot/libs/tor-android-binary-release.aar, /media/delta/T5/android/orbot-latest/orbot/libs/OrbotLib.aar
``

syphyr:
The problem was that each aar needs its own separate module. This should work now.

syphyr:
I also just added updated geoip database using tor 0.4.7.10 release

syphyr:
Force pushed to fixe merge conflicts

syphyr:
Updated geop to 0.4.7.11 and rebased

PR #758, opened 2022-11-16
Expose IsolateDestPort and IsolateClientProtocol

SkewedZeppelin:
Would it be a good idea to enable IsolateDestAddr regardless if VPN mode is in-use?

And should IsolateClientProtocol be default enabled?

Is the wording on pref_isolate_protocol_summary accurate?


syphyr:
I have added this to my test build and it seems good to me. These isolation options are nice to have, but enabling them by default might cause orbot to use more battery life.  The wording of the descriptions look also ok to me.

n8fr8:
Possible performance issues with IsolateDestAddr on by default, but glad to expose it as an option.

SkewedZeppelin:
bump? or close? :duck:

PR #793, opened 2023-02-01
Do not merge: Dummy tests

johnhess:
This adds some dummy unit tests, 1 of which succeeds and 2 of which fail. 

This PR is meant to trigger bitrise's workflow and check that failed tests will stop it from running and produce an informative error.


johnhess:
task failed successfully.

johnhess:
gradle doesn't print the full output of assertions (e.g. expected <3> but got <2>) but does name specific failing tests

PR #801, opened 2023-02-08
build-orbotlib: Check for directory before cloning

syphyr:
Don't try to clone directory if it already exists to prevent fatal error in log


bitmold:
Thanks, i wasn't too concerned with this error message since it's not a part of building Orbot regularly (most people who build Orbot would just use the prexisting aar binary)

Eventually we are going to migrate toward a better way to distribute this artifacts. But as usual, thanks for checking things so thoroughly

PR #828, opened 2023-03-10
Start Orbot via always-on VPN without onboarding

t-m-w:
When the system tries to start Orbot, if the onboarding process has yet to be completed, automatically connect to Tor without requiring user intervention. This allows Orbot to get started automatically when it is bundled with the OS in always-on mode.

Test: Manual: Install Orbot and connect to Tor. In the system settings, set Orbot as an always-on VPN. Clear storage for Orbot. Reboot the device. Orbot should automatically connect and be available for use.

### DIFF
--- a/app/src/main/java/org/torproject/android/OrbotMainActivity.java
+++ b/app/src/main/java/org/torproject/android/OrbotMainActivity.java
@@ -267,6 +267,7 @@ public class OrbotMainActivity extends AppCompatActivity implements OrbotConstan
     private void sendIntentToService(final String action) {
         Intent intent = new Intent(OrbotMainActivity.this, OrbotService.class);
         intent.setAction(action);
+        intent.putExtra(OrbotConstants.EXTRA_NOT_SYSTEM, true);
         sendIntentToService(intent);
     }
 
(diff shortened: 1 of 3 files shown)


n8fr8:
This is a case where the OS has bundled Orbot, and the user has never set it up themselves?

t-m-w:
Yes — that, or the not-so-likely event that someone has Orbot as their always-on VPN and then wipes Orbot's data for some reason, which does not affect its VPN preferences at a system level, like its always-on state. In that case, post-patch, when the user reboots the phone, Orbot will be started by the system, recognize it is being started by the system (as per user request), and connect itself and become ready automatically. Pre-patch, it will not; the VPN service remains unavailable until the user manually opens Orbot at least once.

Specifically, in CalyxOS, we are working to add optional functionality that can make Orbot active and available immediately after the setup wizard without user intervention or requiring to launch Orbot as the active activity even once, and this patch intends to make that possible.

If we were to remove the `Prefs.onboardPending` check from this patch, it would also serve as an extra trigger for users who *have* been onboarded for Orbot to become ready whenever a user has it established as the always-on VPN, potentially even without a boot receiver.

Jester1826:
Hum

bitmold:
This looks legit, I'll test this later this week ...

chirayudesai:
Friendly ping

PR #846, opened 2023-03-24
Clean up and refactor gradle files

meenbeese:
This PR simplifies the Gradle files by cleaning messy code and streamlining the project structure. The old code was needlessly complicated and made maintenance more difficult.


syphyr:
why remove `$kotlin_version` and then hard code it to two different versions?

meenbeese:
I messed that up last time, now it's fixed.

bitmold:
Hey thanks for looking into this - even streamlined gradle files are a big pain to keep track of so I appreciate your desire to help with this 

One thing I noticed is that in `settings.gradle` you added back tor-android when previously it wasn't there

Do you know how to use `git rebase` to squash commits together? I don't know when I'll have the time to review this yet, but once if it's mergeable i'd rather not bring a bunch of "whoops missed this" commits or revert commits into the `main` branch 

- annette

syphyr:
I don't think we should bump the Java version from 11 to 18.

meenbeese:
I applied all of your suggestions and also tried to do the `squash and  merge into current branch` but I don't think I succeeded. It looks like I will need some help on that front.

syphyr:
Adding `ext.ndkVersion = '21.3.6528147'` is incorrect. NDK is not needed here.  Also, 21.4.7075529 is the stable version.

meenbeese:
Why is it not needed? I believe it is more practical to declare it in the top-level rather than in specific Gradle files.

syphyr:
> Why is it not needed? I believe it is more practical to declare it in the top-level rather than in specific Gradle files.

Nothing is being built with NDK in this project directly.  All libraries used by orbot are built in the  tor-android project, which is separate.

meenbeese:
I realized that I accidentally broke something with one of my changes as the Android emulator refuses to work. I will try to find and fix that. Please don't merge anything until then.

meenbeese:
I messed up pretty badly and quickly opened a new PR to replace this one: #915. Please excuse my incompetence with Git.

PR #868, opened 2023-04-08
Fix lint error Instantiatable

syphyr:
Fixes:
```
> Task :app-tv:lintVitalTeeveeRelease FAILED
/orbot/app-tv/src/main/AndroidManifest.xml:54: Error: SettingsPreferencesFragment must extend android.app.Activity [Instantiatable]
            android:name="org.torproject.android.core.ui.SettingsPreferencesFragment"
                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Explanation for issues of type "Instantiatable":
Activities, services, broadcast receivers etc. registered in the manifest
file (or for custom views, in a layout file) must be "instantiatable" by
the system, which means that the class must be public, it must have an
empty public constructor, and if it's an inner class, it must be a static.

### DIFF
--- a/app-tv/build.gradle
+++ b/app-tv/build.gradle
@@ -8,6 +8,9 @@ android {
         targetSdkVersion 33
     }
 
+    lintOptions {
+        disable "Instantiatable"
+    }
 
     flavorDimensions "teevee"
 


n8fr8:
or I should just remove that from the Manifest... I think I didn't mean the fragement to be in there

bitmold:
hmm, I believe that this came out of a recent refactor of settings stuff done by @n8fr8 ~ I don't regularly build.test the TV build but has anyone determiend that the setting screen still works in there? If so, we should keep it in the manifest and go through with @syphyr's change ...

bitmold:
hmm, I believe that this came out of a recent refactor of settings stuff done by @n8fr8 ~ I don't regularly build.test the TV build but has anyone determiend that the setting screen still works in there? If so, we should keep it in the manifest and go through with @syphyr's change ...

syphyr:
This is still needed for `gradlew assembleRelease` to complete successfully.

n8fr8:
fixed here a3efaee43caef5f0317edb23c5c94f0d11fe91c8

PR #879, opened 2023-04-12
Do not repeat Orbi animation infinitely

syphyr:
Note that repeat count must be an odd number to prevent Orbi from skipping up in the last frame


n8fr8:
Orbi is INFINITE!

syphyr:
When the animation is set to infinite and user selects another tab within Orbot application and then goes back, the position of Orbi becomes frozen in mid bounce.

bitmold:
It's true there are a lot of annoying bugs pertaining to infinite animations / keeping track of state

PR #900, opened 2023-05-03
refactor: :art: improve Brazilian Portuguese translation

adnanbezerra:
Plenty of the Brazilian Portuguese entries were actually written in Portuguese from Portugal, which is really different from Brazilian Portuguese. This can cause Brazilian users to, sometimes, not understand what is going on. Also, some translations were just plain wrong or just misspelled. I believe I've solved such problems.


n8fr8:
Thank you for this. We definitely rely on community input for our translations.

It seems you just changed the strings for the "app-tv" which is our Google/Android TV app. Can you also check the strings under the "app" and "appcore" modules?

adnanbezerra:
I can't find the strings at /appcore folder. are you sure there is something to translate in there?

n8fr8:
I meant these strings: https://github.com/guardianproject/orbot/blob/master/app/src/main/res/values-pt-rBR/strings.xml

PR #903, opened 2023-05-10
Use Java 17

syphyr:
Use java 17


syphyr:
Java 17 is the current LTS and seems to work fine. Tested on arm64.

meenbeese:
We don't need this PR, as migration to Java 17 is already being done in #846.

Issue #181, opened 2018-11-24
Quck Settings Tile for Orbot


tristan-k:
With Afwall being able to route apps through Orbot I would like to see [support](https://medium.com/androiddevelopers/quick-settings-tiles-e3c22daf93a8) for Quick Settings Tile in order to quickly toggle Orbot.

T0M0F:
Is there any progress regarding this issue? Otherwise is this project open for contributions? I'd like to try to implement this feature.

n8fr8:
Yes, for sure, please take a shot at it. We will review and merge if it is written decently enough and works.

meenbeese:
Since my PR was merged, this issue can be closed @n8fr8 (sorry for the ping btw)

PR #917, opened 2023-05-21
Added quick settings tile support

meenbeese:
This PR adds quick settings tile support to Orbot, allowing users to easily access key app functionalities from the quick settings panel. This enhancement resolves issue #181, providing improved usability and a seamless user experience. The addition of the tile icon, XML configuration, and behavior implementation enables users to access frequently the app without needing to open it.


SkewedZeppelin:
it just opens the app?

n8fr8:
This looks good and is a good start to the tile support. Will consider the off/on option in a future step.

PR #933, opened 2023-06-04
Minor rephrasing

ivadham:
Some textual changes in order to improve the user experience of those new to Tor and Orbot:

• Choose apps -> Choose apps to include
• Change exit -> Change exit node
• Refresh -> Refresh identity
• Turn Tor off -> Turn off Tor

• internet -> Internet
• wi-fi -> Wi-Fi


ivadham:
I added the `ic_globe.xml` file, but I did not reference it. Also, I did not remove the `🌎` emoji next to "Change exit". This is because my hardware isn't powerful enough to let me build and debug it. So I leave it to you, dear Guardians, if you please. Thanks!

n8fr8:
Appreciate the thoughtfulness and contribution. We will talk about this with our design team. There is some issue with length of strings and the current layout to consider.

PR #956, opened 2023-08-09
Fix linting issues

meenbeese:
Fix several linting problems by:

- Adding assert/NonNull for NullPointerException
- Converting if statements to when or switch
- Deleting large blocks of commented code
- Reducing the amount of empty lines/spaces

Tested on Pixel 5 API 33 device.


app/src/main/java/org/torproject/android/ui/onboarding/CustomBridgesActivity.java
```diff
@@ -86,10 +87,8 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
         findViewById(R.id.btShareQr).setOnClickListener(v -> {
             String setBridges = Prefs.getBridgesList();
             if (!TextUtils.isEmpty(setBridges)) {
-                try {
-                    integrator.shareText("bridge://" + URLEncoder.encode(setBridges, "UTF-8"));
-                } catch (UnsupportedEncodingException e) {
-                    e.printStackTrace();
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
```
n8fr8, line 90: Again why the limit for Android 13 on this call?

n8fr8:
Appreciate the work on this, and will try to merge. In general, we don't like massive commits that have many changes across many files. In the future, if you could break up this kind of work per subpackage or into smaller chunks, that would be helpful. Also see my comments about the checks that limit to Tiramisu/Android13

app/src/main/java/org/torproject/android/OrbotActivity.kt
```diff
@@ -35,6 +37,7 @@ class OrbotActivity : BaseActivity() {
 
     var previousReceivedTorStatus: String? = null
 
+    @RequiresApi(Build.VERSION_CODES.TIRAMISU)
```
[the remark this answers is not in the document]
meenbeese: That's correct, thanks for the heads up. I now moved the SDK check to its intended place.

app/src/main/java/org/torproject/android/ui/onboarding/CustomBridgesActivity.java
meenbeese: Passing a different UTF parameter to the ```
URLEncoder.encode()
``` caused required SDK Level 33 so I now reverted to the old param.

meenbeese:
Thank you for reviewing my PR and appreciate the feedback provided. I fixed the issues mentioned and did some more linting that I previously missed. After this is merged, I want to take on a bigger task of converting more of the code to Kotlin. Would that be on the roadmap or should I focus on fixing bugs first?

meenbeese:
> In the future, if you could break up this kind of work per subpackage or into smaller chunks, that would be helpful.

Indeed, I will try to do this even with this PR, as the changes have become increasingly large. Until I find a way to break up the commits, I will mark it as a draft.

meenbeese:
This PR has become huge and impossible to reasonably split and refactor. I will close it in favor of smaller future ones.

Issue #958, opened 2023-08-13
The confusing new icon


hax0rbana-adam:
This ticket is a request to change the Orbot icon.

I'm really surprised there isn't already an open issue about this, but... the new icon looks so much like a penis and testicles, it was difficult to figure out what it was supposed to be. After I figured out that it's the top of the little mascot's head, the only thing I could think is: why does he have a dick on his head?

![Orbot-cock](https://github.com/guardianproject/orbot/assets/118935832/60a8d111-392e-49fd-9654-d2b342d5e3c5)

Just to sanity check me, I've asked several other people, of various genders, what they think of the icon and everyone has said it looks like male private parts.

Just changing the icon to have the whole mascot instead of just the top of his head would suffice. Also changing the mascot's antennae, or whatever they're supposed to be, so they don't look dirty would be even better.

Orbie-is-super-cute:
I don’t see that thing in logo.

I think mascot is really cute and good looking.

Sigma-One:
i didn't see anything dirty at first but can definitely see how one would when only this part is in focus

i do think the icon is just plain confusing though, and should be changed to something which identifies the app better while also being clearly *something*

i couldn't connect this with the mascot until someone mentioned it, and i do agree that a better icon would be fore example the whole mascot

meenbeese:
> i couldn't connect this with the mascot until someone mentioned it, and i do agree that a better icon would be fore example the whole mascot

I think this is a much more mature and reasonable take on your part compared to the original one. It makes sense to shift the focus from the mascot's hair to make it more understandable.

Does anyone have a concrete idea or proposal for this?

lbdroid:
It doesn't matter what the icon is *meant* to be, it matters what people actually think when they see it. EVERYONE I've spoken to about it has independently come up with the thought that it looks like a dick and balls.

The icon looks *ok* at this older revision;
https://github.com/guardianproject/orbot/blob/865c5c475c09354943b3486e026d7aeac1d2ac96/app/src/main/ic_launcher-playstore.png

But now that the stuff on the top of the little guy's head is associated with a dick, it really should be removed -- make the top curve smooth and uninterrupted. It doesn't make sense to have hair or anything there anyway, since the little guy is clearly wearing a helmeted space suit.

hax0rbana-adam:
I made an updated version of Orbie based on @lbdroid's suggestion

[...]

Issue #397, opened 2020-09-27
Adaptive icon support


staticssleever668:
Currently the icon looks blurry and doesn't change it's shape, adaptive icon would fix both of these issues

svdm-artix:
Is there any progress on this? Or perhaps any way I could help?

n8fr8:
If someone can produce the new icon asset, we wouldn't definitely integrate it right away.

beedell-roke:
@bitmold, which version is it remediated in?

PR #965, opened 2023-08-22
Vector image of the Orbot logo for the launcher

### LINKED ISSUE #581: icon is blurry
Hi, the icon is blurry on my Android 11. My app version is 16.4.1, which is the latest version on f-droid. Is it a bug?

hax0rbana-adam:
- Fixes issue #958 about the new logo being confusing / undesirable
- Make launcher logo match the icon in the F-Driod and Google Play stores
- Uses vector based art to scale to different screen sizes
- I believe this will also resolve #581 and #397

Because Android vector graphics do not render in GitHub, here's an image showing the updated icon:

orbot-no-shadow-deep-purple-background_small


meenbeese:
Does this address #581 as well?

hax0rbana-adam:
> Does this address #581 and #397 as well?

Yes, I believe so. I was able to reproduce the issue #581 on an Android 11 device and v16.4.1 of Orbot from F-Droid and the launcher icon is slightly blurry. Issue #397 doesn't specify what version the reporter was using, but they explicitly say that and adaptive icon would fix this. As far as I can tell this vector format is the only type of icon that will adapt to various screen sizes.

Neither issue specifies whether it is the launcher icon or some other icon in the application, so it is possible they are reporting an issue with some other (or some additional) icon. If they are referring to an icon other than the launcher, this change will not resolve the issue, but it will get us one step closer. Whatever icon needs updated could be replaced with a [VectorDrawable](https://developer.android.com/develop/ui/views/graphics/vector-drawable-resources) and it'll be sharp at any resolution.

meenbeese:
> As far as I can tell this vector format is the only type of icon that will adapt to various screen sizes.

Nice! Thanks for taking a step to fix this.

> Neither issue specifies whether it is the launcher icon or some other icon in the application

I think it is safe to assume that they are both talking about the launcher icon, as that would be what one assumes without explicit mentioning. Otherwise, they would have probably mentioned what icon it is.

Again, thanks for implementing this as I remember vector icons being the recommended ones for Android launchers for solving various reasons described in said issues.

brianSchanbacher:
The new launcher icon looks really clean. A lot of the negative feedback in #958 seems dubious.

meenbeese:
> The new launcher icon looks really clean. A lot of the negative feedback in #958 seems dubious.

As I agree with you, maybe the shadow can also be removed too while we are at it @hax0rbana-adam ? It looks ancient and outdated imo, what do you think?

brianSchanbacher:
> As I agree with you, maybe the shadow can also be removed too while we are at it @hax0rbana-adam ? It looks ancient and outdated imo, what do you think?
> 
> Modern app icons tend to rely less on shadows but more on contrasting colors and backgrounds to differentiate the main subject on the icon

Good feedback, I hadn't noticed the shadow, but you are right. There isn't a single app on my homescreen w/ a shadow.

hax0rbana-adam:
> As I agree with you, maybe the shadow can also be removed too while we are at it @hax0rbana-adam ? It looks ancient and outdated imo, what do you think?

Yeah, I think it'd look fine without the shadow. I was trying to make it match the logos that are on the Google Play and F-Droid app stores.

Conveniently, since it's a vector graphic and not a PNG/GIF/JPG, removing the shadow is just a matter of removing the "shadow" path (the section pasted below). :smiley: Feel free to remove it.

Here's what it looks like rendered without the shadow:

orbot-no-shadow

Also if you want to change the shades of purple to make it pop more, I'd be fine with that too. Again, I was trying to be faithful to the logo that everyone recognizes. Changing the fillColor on the Outline to 4f0063 give a deep purple version that looks better in my opinion. It increases the contrast and makes it easier to see. Here's the rendered version (also without the shadow):

orbot-no-shadow-deep-purple-background

meenbeese:
> I think we need a mascot. Not only logo. Why just not use https://github.com/guardianproject/orbot/blob/865c5c475c09354943b3486e026d7aeac1d2ac96/app/src/main/ic_launcher-playstore.png ?

I understand that you want to express your opinions about the app icon but please cut back on the unnecessary noise you are creating on this thread. You already explained it once and that is enough, writing a message 3 times won't help your cause. @TimoLready

meenbeese:
> Here's what it looks like rendered without the shadow:
Thanks for removing the shadow! It already looks so much better.
> 
> Also if you want to change the shades of purple to make it pop more, I'd be fine with that too. Again, I was trying to be faithful to the logo that everyone recognizes. Changing the fillColor on the Outline to 4f0063 give a deep purple version that looks better in my opinion. It increases the contrast and makes it easier to see. Here's the rendered version (also without the shadow):

Btw using a more popping color is a good idea too that didn't immediately occur to me. I think it makes a lot of sense to make minor tweaks as needed while keeping the base familiar and recognizable.

These are the improvements that are hard to make but also yield a great result. That said, the color is a bit too popping for me and hurts my eyes ever so slightly. We could consider toning it down by making it just a little darker?

What do you guys think?

brianSchanbacher:
@meenbeese I understand where you are coming from. On my screen the sample is rendering at almost 15x15cm and at that size I agree. When I scale the image down to something that Is more indicative of "app icon size" I think the contrast and pop is good.

That being said if we want to rework the colors completely, I'd suggest the following palettes for the "onion" and "purple" colors respectively.

https://colorhunt.co/palette/3a10784e31aa2f58cd3795bd
https://colorhunt.co/palette/c3ba85dad5abf0f0cbfeffe2

meenbeese:
> When I scale the image down to something that Is more indicative of "app icon size" I think the contrast and pop is good.

I will trust your judgement on that as I did not test it myself. It probably looks good and I am just nitpicking.

> That being said if we want to rework the colors completely, I'd suggest the following palettes for the "onion" and "purple" colors respectively.

These resources look nice indeed! They can be very well used in a redesign if need be in the future.

hax0rbana-adam:
> Btw using a more popping color is a good idea too that didn't immediately occur to me.

Heh, well I still credit you for the idea of using more "contrasting colors and backgrounds". :smiley: 

And my bad on the size, I didn't realize the rendered version was so large until after I posted it. Here's a scaled down version of the same thing.

orbot-no-shadow-deep-purple-background_small

I think it looks less eye popping when rendered at normal icon size.

n8fr8:
We are still keeping the pre-v17 icon for now, as we transition users to the new design. We had feedback that changing the icon without warning caused confusion and concern for users in high risk, high anxiety context. I appreciate the work making this new vector icon, and am glad to merge your effort here.

Our new "Orbi" mascot and design is staying, and we'll settle on a new icon for both the Android and iOS apps soon, and when we do, we will make sure to make it vector like this.

QW21WQ:
@n8fr8 i hope you will NOT delete new mascot. hax0rbana-adam’s idea got plenty of dislikes:
IMG_0631

We like mascot. So… I hope it won’t be deleted

beedell-roke:
@n8fr8, should this also replace the `README`'s icon?

https://github.com/guardianproject/orbot-android/blob/bc3f48eeb74f165ec9d59cfedfd536008f313fb1/README.md?plain=1#L3

Issue #870, opened 2023-04-08
[Feature request] Some security measures


Hcugufyfucy:
Hello!
I propose to add a function to protect the application itself with a pin code, password or a graphic key to protect against external tampering with the application's settings. This would protect user data from external malicious tampering with the app's settings. This should help protect the user from theft of authorization data or keys that are stored in the application.

2. I suggest to disable the clear data button in the "about app" section (I mean the one in the list of all Android apps). [I am about this button](https://user-images.githubusercontent.com/130249945/230741093-cd623c88-58eb-41f6-bc6e-fec536e77b43.jpeg). This will help from accidentally deleting important data of the application itself (e.g. onion services) trying to optimize the available space on the device.

Here is some useful information about this:
[Useful information 1](https://stackoverflow.com/questions/6531173/how-to-disable-the-clear-data-button-in-application-info-of-manage-application) 
[Useful information 2](https://m.youtube.com/watch?v=i82K9PSsmu0) 

3. I also suggest encrypting the application database (authorization cookies and data from hosted onion services) and keeping them encrypted until the application is activated by entering the appropriate password. This will protect the user of your app from having the app data copied by law enforcement in totalitarian countries or other intruders. This method already exists in Aegis application

[Useful video about it](https://m.youtube.com/watch?v=k0KcHTkm6GQ)

PR #972, opened 2023-09-01
Disallow clearance of Orbot app data

meenbeese:
Partly addresses #870 by solving the second issue (relating to clearing app data) by disallowing this option. The system is directed to a dummy activity which kills itself instantly the user clears the data.

This is done to prevent any sort of tampering with the app and serves as a good security measure which can later be complemented with other features like app locking using biometric or pin code.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -35,11 +35,12 @@
     <application
         android:name=".OrbotApp"
         android:allowBackup="false"
-        android:allowClearUserData="true"
+        android:allowClearUserData="false"
         android:configChanges="locale|orientation|screenSize"
         android:description="@string/app_description"
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
+        android:manageSpaceActivity=".ui.ManageSpaceActivity"
         android:theme="@style/DefaultTheme"
         tools:replace="android:allowBackup"
         android:hasFragileUserData="false"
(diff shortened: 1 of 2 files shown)


SkewedZeppelin:
this is useless arbitrary restriction, as you could just uninstall the app

and if an attacker has your device unlocked in a state where they could clear arbitrary app data, they can do far worse things

meenbeese:
> this is useless arbitrary restriction, as you could just uninstall the app

Deleting the app is an entirely different case and different measures can be set in place to prevent this. Even without this however, the user will certainly know that something is wrong if for example the attacker has remote access to the device.

> and if an attacker has your device unlocked in a state where they could clear arbitrary app data, they can do far worse things

Security is often about combining multiple layers of protection. While this is obviously not foolproof, the goal is to make an attacker's job as difficult as possible. By implementing restrictions, you add an additional layer of security that can deter or slow down them.

As I stated previously, this is to be complemented with a biometric/pin lock feature which will add a lot more protection in the future. I would appreciate if you considered contributing to help implement this faster.

n8fr8:
Will have to think about this some more... there are other legitimate reasons people want to clear data, and breaking that basic feature of Android could cause problems.

We do have ideas and implementations for adding pin lock feature (see our Circulo app) and also app icon "camouflage" changing that I would like to implement to improve physical security measures.

bitmold:
Hey, closing this since we aren't getting rid of the ability to clear user data outright. 

Perhaps if you're interested @meenbeese you can discuss with Nathan about prototyping a pin lock feature...

PR #985, opened 2023-09-28
Remove unused STUN server

syphyr:
stun.altar.com.pl is no longer used by torproject


syphyr:
@n8fr8 I'm also not sure about stun.sonetel.com and stun.sonetel.net.  We might need to remove those as well, because they are not pingable.

syphyr:
It does not make sense to me that sonetel.com is remove from one line and sonetel.net is removed from the other configuration setting. Is that a mistake?

https://gitlab.torproject.org/tpo/applications/tor-browser-build/-/merge_requests/617/diffs

n8fr8:
Not sure. Happy to remove both of them though, if they are no longer live and available.

syphyr:
stun.sonetel.com and stun.sonetel.net might just be blocking ping requests.  I've just rebased removing stun.altar.com.pl for now.

PR #1013, opened 2023-11-04
Remove obsolete SDK version checks

meenbeese:
Needed for and spun off from #956 

- Removes old checks in app, appcore, and OrbotService
- Removes methods only used for older SDK versions
- Actually adds new checks for app-tv where they seem necessary

Tested on Pixel 7 API 34.

### DIFF
--- a/orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
+++ b/orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
@@ -67,7 +67,6 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 
 import IPtProxy.IPtProxy;
-import androidx.annotation.ChecksSdkIntAtLeast;
 import androidx.annotation.NonNull;
 import androidx.annotation.RequiresApi;
 import androidx.core.app.NotificationCompat;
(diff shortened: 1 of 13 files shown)


orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
```diff
@@ -1249,32 +1246,25 @@ private StringBuffer processSettingsImplObfs4(StringBuffer extraLines) {
...
-
-                    if (proxyUser != null && proxyPass != null) {
-                        if (proxyType.equalsIgnoreCase("socks5")) {
-                            extraLines.append("Socks5ProxyUsername").append(' ').append(proxyUser).append('\n');
-                            extraLines.append("Socks5ProxyPassword").append(' ').append(proxyPass).append('\n');
-                        } else
-                            extraLines.append(proxyType).append("ProxyAuthenticator").append(' ').append(proxyUser).append(':').append(proxyPort).append('\n');
-
-                    } else if (proxyPass != null)
+        var proxyType = prefs.getString("pref_proxy_type", null);
```
bitmold, line 1249: This is actually a bit wrong, this is only code you don't want to run only if we aren't using a VPN.

bitmold:
This was really helpful, thank you! 

I ended up going through a lot of this MR by hand, I think in OrbotService the IDE might have made a mistake and actually modified Orbot. 

Otherwise there were a few other things that were missing here: 
- Removing the legacy backup/restore file IO (I was sad to see this code go after having rewrote a lot of it almost 3 years ago) 
- Removing jscoks dependency from the project 

I didn't touch the orbot TV module, if you want to open a new MR that applies the changes there or some of the other stuff that this MR introduced that weren't about removing legacy codfe please feel free to.

PR #1016, opened 2023-11-06
Add WhatsApp to VPN package blacklist

### LINKED ISSUE #590: User Feedback: WhatsApp calls
Feedback from users in Senegal:

If you have Orbot running and need to make a call on WhatsApp, you'll have trouble. People need to go turn it off before they can make calls.

meenbeese:
WhatsApp calls have not been working for a long time (see #590) and this can be annoying for many people. They might turn off Orbot completely and lose many of the privacy benefits and get angry and uninstall the app completely. For this reason, it makes sense to blacklist WhatsApp until a more comprehensive fix is found, as many people depend on it regularly.

Closes #590

### DIFF
--- a/orbotservice/src/main/java/org/torproject/android/service/OrbotConstants.java
+++ b/orbotservice/src/main/java/org/torproject/android/service/OrbotConstants.java
@@ -159,18 +159,23 @@ public interface OrbotConstants {
     String PREFS_KEY_TORIFIED = "PrefTord";
 
     /**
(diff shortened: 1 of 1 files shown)


n8fr8:
Thanks for the work, but there is still value in proxying all the messages, file sharing, voice messages etc. It is only real-time calling that doesn't work.

Perhaps we can add a warning of some kind.

bitmold:
Closing this for now since we're not blocking Whatsapp, if you end up adding a warning feel like n8fr8 suggested feel free to open a new issue and we can review it :+)

PR #915, opened 2023-05-21
Streamline and clean-up all Gradle files

meenbeese:
This PR comes as a result of an unforeseen issue during a recent `git rebase` process in the previous PR (#846).

I basically modernized/simplified whatever I could modify about the Gradle files to match the currently recommended way of doing stuff and removed unnecessary/duplicate files. I also migrated all Gradle files to Kotlin as it is expected to replace Groovy for Gradle on Android in the future.

PR #1030, opened 2023-12-02
Merge commons-limited-device into app-tv Gradle

meenbeese:
- Allows us to simply remove commons-limited-device
- Important step to merge #915
- Safe changes that are tested on Android TV SDK 33


bitmold:
Since the #915 is a prerequisite draft I'm going to make this PR a draft for the time being

meenbeese:
> Since the #915 is a prerequisite draft I'm going to make this PR a draft for the time being

I had intended to do it the other way around. This PR has a more limited scope and is a precursor for #915. Is there a reason for it not to be merged?

PR #1041, opened 2023-12-11
Start Orbot via always-on VPN

t-m-w:
This resolves issue https://github.com/guardianproject/orbot/issues/1007 and is an updated version of #828 that was merged into v16_maintenance.

---

When the system tries to start Orbot as a result of the always-on VPN setting, automatically connect to Tor as a VPN without requiring additional user intervention. In this case, the system will also start Orbot on boot regardless of the "Start Orbot on Boot" setting. This support is especially important for scenarios where Orbot is bundled with the OS in always-on mode.

Test: Manual: Install Orbot and start the VPN. In the system settings, set Orbot as an always-on VPN. Clear storage for Orbot, and optionally turn on Notifiations. Reboot the device, or turn the always-on VPN setting off and on again. In either case, Orbot automatically connects and is available for use.


bitmold:
Looks great, thank you

t-m-w:
Thanks so much for the fixes and the merge!

PR #1047, opened 2023-12-21
Geoip performance optimization

bitmold:
Rather than taking geoip and geoip6 which are zipped into the APK and inflating them, then copying all 10MB of them each time `OrbotService` starts, we now only extract these files and write them to internal storage if they don't exist or the first time the user opens the app after an update. This way every single time `OrbotService` is subsequently created it doesn't need to do this disk IO. 

My concern was that blocking disk IO could somehow be deprioritized when Orbot starts in the background on device boot or power saving mode is turned onto the device. For users who start with always on VPN maybe this will make them enter the tor network faster. 

This geoip business hooks into a lightweight preference based system into `OrbotApp` that could allow for future on "orbot was just updated to a new version" events.

### DIFF
--- a/app/src/main/java/org/torproject/android/OrbotApp.kt
+++ b/app/src/main/java/org/torproject/android/OrbotApp.kt
@@ -9,7 +9,6 @@ import org.torproject.android.service.OrbotConstants
 import org.torproject.android.service.util.Prefs
 import java.util.Locale
 
-
 class OrbotApp : Application(), OrbotConstants {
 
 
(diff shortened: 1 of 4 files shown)


syphyr:
This PR is working as expected. I have tested upgrading and also initial installations.
Just to be sure, I added some additional logging to make sure things were happening correctly like this:

```
--- a/orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
+++ b/orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
@@ -466,6 +466,7 @@ public class OrbotService extends VpnService implements OrbotConstants {
                 // only write out geoip files if there's an app update or they don't exist
                 if (!hasGeoip || !hasGeoip6 || Prefs.isGeoIpReinstallNeeded()) {
                     try {
+                        Log.i("OrbotService", "Installing geoip files");
                         new CustomTorResourceInstaller(this, appBinHome).installGeoIP();
                         Prefs.setIsGeoIpReinstallNeeded(false);
                     } catch (IOException io) { // user has < 10MB free space on disk...
```

bitmold:
Excellent, I'll push your logging commit

PR #1061, opened 2024-01-24
Reduce OrbotLib binary size by ~53% with go compiler ldflags -w and -s

bitmold:
@tladesignz I can't remember if we ever discussed these flags before? It reduces the gomobile aar (containing code for each architecture) from about 50MB->25MB.

`go tool link`:
```
-s	disable symbol table
-w	disable DWARF generation
```

Making this a PR because I'm not sure if there is any reason to not set these flags... As far as my minimal testing is concerned IPtProxy/gotun2socks seems to be behaving fine.

My `gomobile bind` command from this build script: https://github.com/bitmold/OrbotIPtProxy/blob/master/build-orbot.sh
```sh
gomobile bind -o "OrbotLib.aar" -ldflags="-w -s" -target=android -androidapi="$MIN_ANDROID_SDK" -v -trimpath
```

Would appreciate any and all feedback/discussion/testing... :)


bitmold:
CC'ing @syphyr

syphyr:
I have been using those two flags without any problems with my other go applications for a long time.  Good idea.

tladesignz (approved):
Well, you loose nice stack traces and such, but compared to how often I had used these (like, never?), that's probably an ok trade-off, I guess?

Thanks for letting me know! I'll add this to IPtProxy. (Where I can confirm the ~50% savings on all target platforms.)

PR #1063, opened 2024-01-25
Changes to make minifyEnabled work...

bitmold:
Was testing this and it seems to work, shaves off some megabytes. 

It ought to be tested more thoroughly, but I'm confident with what I've seen so far. Open to feedback, of coruse.

You either need to make release builds, or do what I was doing and just change the `release` part of `buildTypes` in each module's build.gradle to `debug`...

Again, would kindly appreciate your thoroughness here @syphyr

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -32,8 +32,8 @@ android {
     }
     buildTypes {
         release {
-            shrinkResources false
-            minifyEnabled false
+            shrinkResources true
+            minifyEnabled true
         }
     }
 
@@ -87,6 +87,7 @@ android {
 
         implementation "com.squareup.retrofit2:retrofit:2.9.0"
         implementation 'com.squareup.retrofit2:converter-gson:2.9.0'
+        implementation "org.slf4j:slf4j-simple:1.6.1"
 
         androidTestImplementation libs.fastlane_screengrab
 //        coreLibraryDesugaring 'com.android.tools:desugar_j dk_libs:1.1.5'
--- a/appcore/build.gradle
+++ b/appcore/build.gradle
@@ -17,7 +17,7 @@ android {
 
     buildTypes {
         release {
-            minifyEnabled false
+            minifyEnabled true
             proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
         }
     }
--- a/intentintegrator/build.gradle
+++ b/intentintegrator/build.gradle
@@ -10,7 +10,7 @@ android {
 
     buildTypes {
         release {
-            minifyEnabled false
+            minifyEnabled true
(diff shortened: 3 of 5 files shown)


syphyr:
Is it necessary to add the `org.slf4j:slf4j-simple:1.6.1` implementation and additional proguard rules in order to minify the build?

syphyr:
Everything seems fine after applying this commit.  I also tried removing the proguard and slf4j changes since they did not seem to be necessary or related to minify.

syphyr:
Oh, I see now that slf4j is required for minify. Sorry about that..
WARNING:R8: Missing class org.slf4j.impl.StaticLoggerBinder
Now i see why you added it. ;-)

syphyr:
is `crunchPngs true` also something we want?

syphyr:
Also, since the resources become obfuscated by R8, maybe ` jniDebuggable false` is also appropriate for this.

syphyr:
@bitmold When I started testing this on the master branch, I ran into an error because of AGP 8.  This is what I did to solve the issue: https://github.com/syphyr/orbot/commit/36635a7afd00b7de8becf67ff8dfa5253a53e2c1

syphyr:
Minify is causing a bug with retrofit2 by renaming the classes.  This causes a fatal exception when selecting the "Ask Tor" feature.  I have tried to fix this issue by using the examples here, but so far I have been unsuccessful:
https://stackoverflow.com/questions/35506111/retrofit-2-not-sending-data-when-proguard-is-enabled

@bitmold if you have any further suggestions on this issue, please let me know.

syphyr:
One suggestion on fixing the retrofit2 issue was to add Keep annotations, but that seems to also not be working.

```
--- a/app/src/main/java/org/torproject/android/circumvention/CircumventionApi.kt
+++ b/app/src/main/java/org/torproject/android/circumvention/CircumventionApi.kt
@@ -12,10 +12,11 @@ import retrofit2.converter.gson.GsonConverterFactory
 import retrofit2.http.Body
 import retrofit2.http.GET
 import retrofit2.http.POST
+import androidx.annotation.Keep
 import java.net.InetSocketAddress
 import java.net.Proxy
 
-
+@Keep
 interface CircumventionEndpoints {
 
     @GET("countries")
```

bitmold:
I don't have any suggestions at the time being. Appreciate you posting what you tried here though. My strategy is just to experiment with building orbot with minify until everything works. It's just gonna b a lot of trial and error until it's done but as of right now it's hard to suggest anything without compiling+deploying & recompiling+redeploying, etc

syphyr:
Using the @Keep annotations seems like the easiest approach.  Here is some further info that seems useful:
https://stackoverflow.com/questions/6548213/how-to-keep-classes-containing-specific-members

bitmold:
Closing for now, this work has been lost to time,

PR #1075, opened 2024-01-27
Add preference to disable warning if root is enabled

syphyr:
Root detection warning can be disabled with this setting


bitmold:
Sick!

PR #1078, opened 2024-01-29
Make Orbot direct boot aware

syphyr:
Orbot will not startup quickly on boot if it is not direct boot aware. If Orbot is direct boot aware, it must check for internet connectivity first or it will try to start too soon and hang.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -110,7 +110,7 @@
             android:enabled="true"
             android:exported="true"
             android:permission="android.permission.RECEIVE_BOOT_COMPLETED"
-            android:directBootAware="false">
+            android:directBootAware="true">
             <intent-filter>
                 <action android:name="android.intent.action.BOOT_COMPLETED" />
                 <action android:name="android.intent.action.QUICKBOOT_POWERON" />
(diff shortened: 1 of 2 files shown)


bitmold:
@n8fr8 is there a reason why this was set to false and this code was commented out?

n8fr8:
To support DirectBoot, you need to do more than just enable a flag. You have to support the special temporary storage that is available BEFORE the user has unlocked their device after a boot:

https://developer.android.com/privacy-and-security/direct-boot
Android 7.0 runs in a secure, Direct Boot mode when the device has been powered on but the user hasn't unlocked the device. To support this, the system provides two storage locations for data:

Credential encrypted storage, which is the default storage location and only available after the user has unlocked the device.
Device encrypted storage, which is a storage location available both during Direct Boot mode and after the user has unlocked the device.

syphyr:
I realize now that making orbot direct boot aware is not really making orbot startup faster on boot.  Actually, what is making it startup faster is adding back these intent filters:

```
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -112,8 +112,11 @@
             android:permission="android.permission.RECEIVE_BOOT_COMPLETED"
             android:directBootAware="false">
             <intent-filter>
+                <action android:name="android.net.wifi.STATE_CHANGE" />
+                <action android:name="android.net.conn.CONNECTIVITY_CHANGE" />
                 <action android:name="android.intent.action.BOOT_COMPLETED" />
                 <action android:name="android.intent.action.QUICKBOOT_POWERON" />
+                <action android:name="android.intent.action.REBOOT" />
             </intent-filter>
         </receiver>
 
```
The REBOOT intent filter is probably not necessary for a faster startup time on boot.

syphyr:
I'm closing this because direct boot aware has never worked properly and just enabling it is not the correct thing. Also, my issue about slow startup on boot was fixed by adding those missing intent filters that I previously mentioned above.

PR #1103, opened 2024-03-08
Remove obsolete LibEvent URL in about

meenbeese:
The old URL now redirects to the new one: https://www.libevent.org

Edit: Orbot does not directly rely on LibEvent, thus all links have been removed.

### DIFF
--- a/app-tv/src/main/res/layout/layout_about.xml
+++ b/app-tv/src/main/res/layout/layout_about.xml
@@ -68,13 +68,6 @@
 				android:textColorLink="#ffffff"
 				android:paddingLeft="15px"
 				android:textColor="#ffffff" />				
-			<TextView android:text="@string/libevent_version"
-				android:layout_width="fill_parent" 
-				android:layout_height="wrap_content"
-				android:autoLink="web"
-				android:textColorLink="#ffffff"
-				android:paddingLeft="15px"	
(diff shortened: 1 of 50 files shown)


syphyr:
libevent (2.1.12) has moved into the tor-android project and is indirectly linked into Orbot from libtor.so.  This must have been left over from a very old version of orbot.  Descriptions of libevent in orbot should be removed, not updated.

meenbeese:
> libevent (2.1.12) has moved into the tor-android project and is indirectly linked into Orbot from libtor.so. This must have been left over from a very old version of orbot. Descriptions of libevent in orbot should be removed, not updated.

Done! Thanks for the valuable insight.

bitmold:
I disagree, we still compile libevent and that's a functional aspect of Orbot's tor binary, I think it's worth updating and not removing. 

I think it's actually bworthwhile to include this kind of info for not just libevent, but zlib, zstd and any other dependencies that shows what the little t tor within Orbot is running with. 

It'd actually be nice to include these in TorService so not just Orbot but any tor-android app can access this info. Adds transparency and shows users whether or not we are on top of things.

bitmold:
In the future for strings like this we can set traslateable=false on these kinds of string resources, wastes space and is confusing/noise for translators

PR #1115, opened 2024-04-01
Iptproxy webtunnel

syphyr:
Update to IPtProxy 3.7.0 development build with webtunnel support
    
Update to snowflake v2.9.2-8-g095e972
Update to lyrebird lyrebird-0.1.0-35-ga571773
    
IPtProxy sources from https://github.com/syphyr/IPtProxy

### DIFF
--- a/orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
+++ b/orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
@@ -343,6 +343,8 @@ public class OrbotService extends VpnService implements OrbotConstants {
         // @param fronts Comma-separated list of front domains.
         // @param ampCache OPTIONAL. URL of AMP cache to use as a proxy for signaling.
         //	Only needed when you want to do the rendezvous over AMP instead of a domain fronted server.
+        // @param sqsQueueURL OPTIONAL. URL of SQS Queue to use as a proxy for signaling.
+        // @param sqsCredsStr OPTIONAL. Credentials to access SQS Queue.
         // @param logFile Name of log file. OPTIONAL. Defaults to no log.
         // @param logToStateDir Resolve the log file relative to Tor's PT state dir.
         // @param keepLocalAddresses Keep local LAN address ICE candidates.
@@ -350,7 +352,7 @@ public class OrbotService extends VpnService implements OrbotConstants {
         // @param maxPeers Capacity for number of multiplexed WebRTC peers. DEFAULTs to 1 if less than that.
         // @return Port number where Snowflake will listen on, if no error happens during start up.
          */
-        IPtProxy.startSnowflake(stunServer, target, front, null, null, true, false, false, 1);
+        IPtProxy.startSnowflake(stunServer, target, front, null, null, null, null, true, false, false, 1);
 
     }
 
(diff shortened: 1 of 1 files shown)


bitmold:
I'm going to update the repo for OrbotLib. @n8fr8 orbotlib really ought to be properly distro'd  as a dependency rather than a local AAR in the project. But in any case, it's a nice contribution but we we can't accept a merge request that has changes a large binary asset like this.

syphyr:
I understand, but you can just pull my IPtProxy patches and rebuild it. I already have it all working.

bitmold:
I don't handle the releases of that project, the way that we get code pluggable transports code into Orbot starts with @tladesignz adding it and us making any android changes if need be. IPtProxy is used for a handful of other projects that aren't Orbot on Android so the procedure here is important. 

Further, (while your code seems really well written) there may be considerations on other platforms that your code doesn't address which is why we haven't gone and made our changes directly into IPtProxy. 

tla also has used different conventions in how he commits changes than your 6 commits. anyway, i linked your commits on the iptproxy repository and it's really up to tla whether they want to directly merge your commits into the repo or to use them in a looser sort of way.

tladesignz:
I released a new version 3.8.0 of IPtProxy containing Lyrebird 0.2.0 with Webtunnel support. Thanks a ton @syphyr, for your support!
https://github.com/tladesignz/IPtProxy/commits/3.8.0

syphyr:
> I released a new version 3.8.0 of IPtProxy containing Lyrebird 0.2.0 with Webtunnel support. Thanks a ton @syphyr, for your support! https://github.com/tladesignz/IPtProxy/commits/3.8.0

no problem. I think there is one line change missing.
https://github.com/syphyr/IPtProxy/blob/9455873ea1392479bfe4c76aedc24d8c851d16a8/IPtProxy.go/IPtProxy.go#L453

tladesignz:
Yeah, stupid me...

syphyr:
Also, the logging now only prints "0.2.0" here without the `getVersion()` call being used again which normally printed out `lyrebird-0.2.0` in the logs.
https://github.com/syphyr/IPtProxy/blob/18a2f875ee0fc273a4db55f7c9eaa20f2db594a4/lyrebird.patch#L111

tladesignz:
> Also, the logging now only prints "0.2.0" here without the `getVersion()` call being used again which normally printed out `lyrebird-0.2.0` in the logs. https://github.com/syphyr/IPtProxy/blob/18a2f875ee0fc273a4db55f7c9eaa20f2db594a4/lyrebird.patch#L111

Fixed with [tladesignz/IPtProxy/720ab2f5](https://github.com/tladesignz/IPtProxy/commit/720ab2f574f28370001b394da928589a058fbf9b)

syphyr:
Well, the reason I wanted to fix the version string in that particular way is because orbot uses it to display the version in about screen. The solution you used now still requires me to change the code in orbot now so it is displayed correctly.

tladesignz:
**sigh** ok, ok, you'll get another release later today.

tladesignz:
Wait, what? You should use IPtProxyLytebirdVersion() to get the version. That should display correctly.

Your getVersion addition should have nothing to do with it!?

syphyr:
Without having to make any further changes, Orbot expects a number only string `0.2.0` and lyrebird's logging expects it as `lyrebird-0.2.0`

tladesignz:
Well, sorry. I'll follow the upstream change. You would have gotten the same result if you used Lyrebird directly.
As few patches as possible, as close to the original as possible.

PR #1119, opened 2024-04-06
Fix grammar in internet title string key

meenbeese:
Correct misspelled `no_interent_title` -> `no_internet_title`

No visual changes, only affects the string key in resources.


n8fr8:
interent!

PR #1123, opened 2024-04-18
Rewrite OnionServiceDeleteDialogFragment in Kotlin

meenbeese:
Continues the Kotlin rewrite with the Onion Service.

Tested on Pixel 8 API 34.


bitmold:
Looks good, but wondering if you have you used this code to delete an onion service and have seen the functionality to be the same?

bitmold:
Cool thanks! If you're up for it, it'd be cool to have this screen use an empty view for when there are no user/app services being listed..

meenbeese:
Thanks for the feedback, you mean a placeholder text/image like this, right? I was already planning to improve the UI by replacing the radio buttons with segmented buttons and this can be a nice addition to that.

bitmold:
Yup, that's what I meant !

PR #1140, opened 2024-05-25
Script to Update Snowflake Bridge Config from Tor Project; Use When Making Releases  

bitmold:
I realized that the default config for the two snowflake preset bridges was pretty out of date which might cause users to be censored in some areas where snowfalke is effective. 

I removed that info from the `fronts` asset file, and created a script that automatically grabs the latest bridge info from tor project. 

```bash
./update_snowflake_bridges.sh
```

This updates a new asset file in OrbotService called `assets/snowflake-brokers` which `OrbotService.java` pulls from at runtime. Just run the script and commit any changes to the asset file to be on top of these changing targets for snowflake.

Since there's another script I've been using for updating geoip files, which get updated every major tor release in `libs/build-geoip-jar.sh` (which I ran when updating Orbot from tor 0.4.8.7 -> 0.4.8.11) I created a script that calls both of these scripts called `do_release_housekeeping.sh`

### so @n8fr8 whenever you want to make a new Orbot release run `./do_release_housekeeping.sh` from the root of the project and geoip, geoip6, and snowflake bridge config info will all be the latest and greatest.


bitmold:
I know you just made an Orbot release, but if you merge this it would a good idea to make a hotfix releasre on top of that one with this latest snowflake config info

bitmold:
@tladesignz it looks like you do something similar, but perhaps this may be helpful for you with [IPtProxyUI-ios](https://github.com/tladesignz/IPtProxyUI-ios)... I see that you're hitting the MOAT api, but as far as I know in projects like Tor Browser the `pt_config.json` from the `tor-browser-build` repo is an authoritative source, see: https://gitlab.torproject.org/tpo/applications/tor-browser/-/merge_requests/877

tladesignz:
> @tladesignz it looks like you do something similar, but perhaps this may be helpful for you with [IPtProxyUI-ios](https://github.com/tladesignz/IPtProxyUI-ios)... I see that you're hitting the MOAT api, but as far as I know in projects like Tor Browser the `pt_config.json` from the `tor-browser-build` repo is an authoritative source, see: https://gitlab.torproject.org/tpo/applications/tor-browser/-/merge_requests/877

Thanks. I'm not convinced, however. This file is already 1 month old. And I would be surprised if the source of this isn't actually Bridge DB / RdSys / Moat.
I rather stay with reading the latest info from RdSys.

n8fr8:
ironically, recently, the fact we were out of date (still using github as a front) meant we helped user STAY connect to snowflake when other newer options suddenly stopped working.

that said, I appreciate what this is for and will review.

bitmold:
That makes sense. 

For the record the basis of this script was ripped from Tor's `tor-browser-build` repo, there's a module in there called `tor-android-service` which has this in its build script:

```bash
# Update bridges list
    PT_CONFIG="$tebdir/tor/pluggable_transports/pt_config.json"

    # Writes bridge-line prefs for a given bridge type
    BRIDGES_TXT="service/src/main/assets/common/bridges.txt"
    echo -n "" > "${BRIDGES_TXT}"
    function bridges_conf {
      local bridge_type="$1"
      jq -r ".bridges.\"$bridge_type\" | .[]" "${PT_CONFIG}" | while read -r line; do
        echo $line >> "${BRIDGES_TXT}"
      done
    }

    # Iterate over our bridge types and write default bridgelines for each
    for bridge_type in $(jq -r ".bridges | keys[]" "${PT_CONFIG}"); do
      bridges_conf $bridge_type
    done
```

meenbeese (commented):
Move scripts to the `scripts` directory?

PR #1165, opened 2024-08-01
Enable debuggable build type for debugging

meenbeese:
This PR introduces a debuggable build variant into the project. Now, with the flag `debuggable true` and `applicationIdSuffix` set to `.debug`, devs are able to generate and run a debuggable version of the app. This is especially helpful on real devices because it helps prevent package name conflicts between `debug` and `release` versions and allows them to be installed side-by-side. This change will thus provide more effective identification and fixing of issues during the development process.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -38,6 +38,10 @@ android {
             shrinkResources false
             minifyEnabled false
         }
+        debug {
+            debuggable true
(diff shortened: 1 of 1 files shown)


meenbeese:
@bitmold @n8fr8 can we merge this?

PR #1166, opened 2024-08-03
Leverage DiskUtils function for TV activity

meenbeese:
The same implementation is present in DiskUtils, thus the code should not be duplicated.


bitmold:
Thanks, this is nice

Issue #790, opened 2023-01-30
Update UI Navigation and Home View


cstiens:
c to add views for:
- first connect
- connected
- second connect

PR #1167, opened 2024-08-04
Refresh the navbar with new icons and animations

### LINKED ISSUE #911: Change 'connect' icon
We need to use something other than this:
Screenshot 2023-05-16 at 8 41 34 AM

meenbeese:
This PR enhances the navbar icons and transition animations, thus improving navigation, clarity, and content.

Part of #790 
Closes #911 

### Icons:
Now each navbar icon has two variations: `outline` and `filled`. The filled one is used when the item is selected, and the outline is used otherwise. This is done via a selector for each of the icons that manages this, so the navbar icon can be set to the selector. The old icons are not deleted and new ones are defined instead to not break other usages of the icons. This approach helps maintain consistency and ensures that the navigation experience is intuitive and visually appealing.

<table>
    <tr>
        <th>Before</th>
        <th>After</th>
    </tr>
    <tr>
        <td>
</td>
        <td></td>
    </tr>
</table>

### Animations:
There are four animations in total, which are the left and right variations of both `slide-in` and `slide-out`. These animations are used to create a smooth transition effect when navigating between different fragments in the app. The `navOptionsLeftToRight` is used when navigating to a fragment with a higher item ID, and `navOptionsRightToLeft` is used when navigating to a fragment with a lower item ID. This ensures that the transitions feel natural based on the direction of navigation.

#### Before:

[Animations-before.webm](https://github.com/user-attachments/assets/d5623767-d9f3-48c4-8208-56beb8a070ad)

#### After:

[Animations-after.webm](https://github.com/user-attachments/assets/88deaa4e-cefd-4303-a1e3-e8b900349e5a)

Tested on Pixel 8 API 34 Emulator.


n8fr8:
I like a lot of this, but why did you replace the Onion connect with a key?

meenbeese:
> I like a lot of this, but why did you replace the Onion connect with a key? 

The onion icon was found to be confusing in #911, and the best replacement I found from [Google Icons](https://fonts.google.com/icons) was the key. I thought it signified a "secure connection" as it relates to Tor, but I am open to changing it for sure.

syphyr:
This is a really nice improvement to the UI,, but I think guardianproject would rather use their tor icon instead of the key.

syphyr:
Do you think you could add an additional patch that replaces the key with the onion icon? This way, both icons can be reviewed.

syphyr:
This PR introduces a bug where the Ports to not get updated upon initial start on the MoreFragment

n8fr8:
@meenbeese happy to merge this if you address the bug from above and revert to the icon we have for Connect.

Thanks for all the work!

meenbeese:
> @meenbeese happy to merge this if you address the bug from above and revert to the icon we have for Connect.
> 
> Thanks for all the work!

Thanks for the feedback. I fixed the bug and reverted back the connect icon.

syphyr:
> > @meenbeese happy to merge this if you address the bug from above and revert to the icon we have for Connect.
> > Thanks for all the work!
> 
> Thanks for the feedback. I fixed the bug and reverted back the connect icon.

On the "More" Fragment, it still says "Proxy Ports: Ports Not Set" after connecting to tor with a direct connection. I am also in Power User Mode, if that makes a difference.

meenbeese:
> On the "More" Fragment, it still says "Proxy Ports: Ports Not Set" after connecting to tor with a direct connection. I am also in Power User Mode, if that makes a difference.

I actually don't know why this is still the case. I have reverted all of my changes related to the logic and fragments. Can you please test it on the master branch and see if it happens there as well? It might be a regression unrelated to this PR.

syphyr:
> > On the "More" Fragment, it still says "Proxy Ports: Ports Not Set" after connecting to tor with a direct connection. I am also in Power User Mode, if that makes a difference.
> 
> I actually don't know why this is still the case. I have reverted all of my changes related to the logic and fragments. Can you please test it on the master branch and see if it happens there as well? It might be a regression unrelated to this PR.

I just tested latest fix (with and without power user mode enabled) and the ports are still not being set after connecting to tor with a direct connection.

Screenshot_20250202

n8fr8:
Fixed port issue and merged. Thanks!

PR #1200, opened 2024-11-07
Fix path in moat url

syphyr:
There needs to be a "/" at the end of this url according to the source code


cohosh:
It doesn't matter if the trailing slash is there or not.

syphyr:
OK. It seems this is wrong.

PR #1203, opened 2024-11-19
Update fronts in snowflake-broker

syphyr:
The fronts in snowflake-broker have been updated to match the contents of ./orbotservice/src/main/assets/fronts


syphyr:
I've also removed stun.sonetel.com and stun.sonetel.net because they no  longer work.

syphyr:
Also, I've been just informed that the owner of stun.stunprotocol.org has requested this domain no longer be used by the tor project, so, I've added another change to remove it.

mScott224 (approved):

mScott224 (approved):

PR #1206, opened 2024-11-22
Fix app selection

pipcet:
I was quite shocked today when I enabled an app in the Orbot app selector, only to find out it hadn't been torified.

There are two problems with the current code:

1. When "back" or the left arrow are used in the app selection "activity", all apps you added to the torified app list are forgotten. This means users are running unsecured apps rather than torified ones. That's a severe bug.
2. When you figure out that you have to hit the "save" button, you might think that actually makes the settings take effect. No. You have to leave the app selector activity using the "back" button (after hitting "save") for the app to finally be added to the VPN. This means if you select an app, "save", and then switch to the app, it won't be torified. That's another severe bug.

I really don't think there's any excuse to display a checked checkbox next to an app logo unless that app is actually running torified at that very moment. I think we can agree that is a problem that needs to be fixed, even if we disagree about whether I've chose the right way to fix it.

My fix is to remove the "save" button entirely and simply restart the VPN synchronously whenever a checkbox changes. This doesn't seem to cause noticeable delays, but even if it were to cause them, they're necessary, because we really don't want the user to falsely believe they're running through Tor when they're not.

Again, I'm happy to find another way to solve this

### DIFF
--- a/app/src/main/java/org/torproject/android/ui/AppManagerActivity.kt
+++ b/app/src/main/java/org/torproject/android/ui/AppManagerActivity.kt
@@ -32,7 +32,9 @@ import kotlinx.coroutines.withContext
 
 import org.torproject.android.BuildConfig
 import org.torproject.android.R
+import org.torproject.android.core.putNotSystem
 import org.torproject.android.service.OrbotConstants
+import org.torproject.android.service.OrbotService
 import org.torproject.android.service.util.Prefs
 import org.torproject.android.service.vpn.TorifiedApp
 
@@ -82,10 +84,7 @@ class AppManagerActivity : AppCompatActivity(), View.OnClickListener, OrbotConst
     }
 
     override fun onOptionsItemSelected(item: MenuItem): Boolean {
-        if (item.itemId == R.id.menu_save_apps) {
-            saveAppSettings()
-            finish()
-        } else if (item.itemId == android.R.id.home) {
+        if (item.itemId == android.R.id.home) {
             finish()
             return true
         }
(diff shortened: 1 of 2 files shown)


pipcet:
Sorry, I realized my bug report was incorrect in that "SAVE" automatically returns from the activity. So point 2 isn't necessarily valid. My apologies.

n8fr8:
Thanks for the report and effort to improve this area. We will review and consider!

pipcet:
> The problem with this is that everytime you check or uncheck an app in the UI, it will fully reboot the VPN stack, potentially leaking traffic each time you do that.

If that is an actual bug (in Android or Orbot), it would happen with or without my patch, and we need to warn users about it! They cannot safely change the apps which are VPN'd while Orbot is in use, no matter how we modify the app set. It would happen slightly more often with the patch, admittedly, but I don't see how that's relevant: it shouldn't happen at all, and if it's unavoidable, the UI needs to reflect that. A "save" button certainly doesn't imply that traffic is being leaked while the save is in progress.

> Not sure this is an improvement.

I've asked a few people, and the "SAVE" button simply isn't being recognized. Either they swipe back, which silently (there's not even a warning!) throws away any modifications to the app set, or they switch apps, which means the modifications will be ignored until the user switches back and hits "SAVE", if they ever do. So the current state is that Orbot fails to protect traffic from apps that the user went to the trouble to select, and that's definitely something we should strive to improve.

meenbeese:
How about displaying a toast message to communicate this to the user? This seems like the most straightforward way until a fix can be found.

bitmold:
A Toast message has been added actually in a recent rework of the app selection screen. Additionally, the VPN is only reloaded when the choose apps screen is closed - if and only if - any new apps were checked/unchecked.

PR #1236, opened 2025-01-16
Fix build failure and use default config for app-tv settings

syphyr:
Fix build failure because dependencies require API 35


meenbeese:
Thanks for fixing this. I should have definitely tested my changes :)

PR #1242, opened 2025-01-24
Add support for PollInterval

syphyr:
The recomended value for mobile devices is 120 seconds. This should save some battery life on mobile devices.

Note: IPtProxy must be updated to 4.0.1 before this can be merged


syphyr:
More info:
https://github.com/tladesignz/IPtProxy/pull/58#issuecomment-2527607585

Issue #1186, opened 2024-10-09
Please, add minSdk21 (Android 5.0) !!! [FeatureRequest]


DI555:
Please, add minSdk21 (Android 5.0) , or, if not possible, than minSdk22 (Android 5.1) !!!
Cli version of tor works fine in sdk21..
So will be very gratefull to get this functionability back again.
Thank you in advance!

syphyr:
You would also need to build OrbotLib with sdk 21, but that also works fine.  I don't see why it could not be done.

syphyr:
Fixed by https://github.com/guardianproject/orbot-android/pull/1243

PR #1243, opened 2025-01-24
Set minSDK to 21 to support legacy devices

syphyr:
This commit also requires the following change before merging: https://github.com/guardianproject/OrbotIPtProxy/pull/2

### DIFF
--- a/app-tv/build.gradle
+++ b/app-tv/build.gradle
@@ -6,7 +6,7 @@ android {
 
     defaultConfig {
         applicationId "org.torproject.android.tv"
-        minSdk = 23
+        minSdk = 21
         targetSdk = 35
     }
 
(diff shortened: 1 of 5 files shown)


syphyr:
For some reason, changing the minSDK to 21 cuts the size of Orbot's apk from ~40MB to ~20MB, which is an added benefit.

syphyr:
I think the reason that the size of Orbot's apk is cut in half when Orbot's minSDK is changed from 23 to 21 is because tor-android is set to minSDK 21 and Orbot's minSDK should match what tor-android is set to.

meenbeese:
Can you add `Closes #1186` to the PR description to automatically close said issue?

syphyr:
> Can you add `Closes #1186` to the PR description to automatically close said issue?

Done. Thanks.

PR #1246, opened 2025-01-30
Add preference to enable device rotation

syphyr:
Currently, device rotation is locked into portrait for phones and tablets are locked into either portait or landscape, depending on the position upon startup.  This new preference allows device rotation for all devices when enabled.


app/src/main/java/org/torproject/android/OrbotActivity.kt
```diff
@@ -38,24 +38,25 @@ class OrbotActivity : BaseActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
-        /** TODO TODO TODO TODO TODO
```
meenbeese, line 41: Remove TODO?

meenbeese:
I don't really get why we should turn this into a pref. Locking the orientation is bad UX, we should focus on the underlying issue that caused the orientation to not work properly in the first place.

syphyr:
> I don't really get why we should turn this into a pref. Locking the orientation is bad UX, we should focus on the underlying issue that caused the orientation to not work properly in the first place.

I agree that we should also determine the cause of these rotation problems, but this still seems like it could be a preference. Maybe some users want the rotation locked. In the meantime, the rotation issues will never get fixed if it is always locked and can't be tested by users.

meenbeese:
> > I don't really get why we should turn this into a pref. Locking the orientation is bad UX, we should focus on the underlying issue that caused the orientation to not work properly in the first place.
> 
> 
> 
> I agree that we should also determine the cause of these rotation problems, but this still seems like it could be a preference. Maybe some users want the rotation locked. In the meantime, the rotation issues will never get fixed if it is always locked and can't be tested by users.

Ok seems like a good starting point to me. It also makes testing rotation on the emulator easier. And when the rotation issues are fixed, we can get rid of the pref.

PR #1248, opened 2025-02-01
Fix support for Snowflake Amp

syphyr:
This fixes support for Snowflake Amp.


syphyr:
Snowflake brokers should only set fingerprint and utls-imitate in order to support both snowflake and snowflake-amp. The rest of the IPtProxy settings for snowflake and snowflake-amp are done within the OrbotService with mIptProxy.set*.

syphyr:
This PR also fixes a bug where the snowflake client settings are being set twice. Once in OrbotService and once within the snowflake-broker file. The snowflake and snowflake-amp client specific settings should only be set by mIptProxy.set*.

syphyr:
@n8fr8 I'm wondering if this PR should be separated into two separate commits.  One that re-enables Snowflake Amp and the other that fixes the bug in snowflake-broker which causes IPtProxy settings to be set twice.

PR #1258, opened 2025-02-27
Implement syntax check for custom bridges

### LINKED ISSUE #1250: [BUG] Crash orbot when start
**To Reproduce**
1. Go to 'orbot'
2. Click on 'click start'
3. See error

Image

Image

- Device: [xiaomi mi 8]
- OS: [e.g. Android 10]
- Version: [e.g. v17.4.1 Beta 3]

meenbeese:
This PR implements a syntax check using Regex to verify custom webtunnel and obfs4 bridges from the user.

Needs testing to be done.

Closes #1250

### DIFF
--- a/app/src/main/java/org/torproject/android/CustomBridgeBottomSheet.kt
+++ b/app/src/main/java/org/torproject/android/CustomBridgeBottomSheet.kt
@@ -1,6 +1,8 @@
 package org.torproject.android
 
 import android.os.Bundle
+import android.text.Editable
+import android.text.TextWatcher
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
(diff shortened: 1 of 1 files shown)


syphyr:
> @syphyr can you please check if this works?
It looks good. The only thing is that the bridge line should be able to handle a new line at the end of the bridge configuration.  If I go to https://bridges.torproject.org/options and copy the two bridge lines and paste it into the custom bridges, there is a new line at the very end which causes the regex to say the config is invalid. If I remove the new line at the end, then it works, but it should probably not have to do that.

meenbeese:
> > @syphyr can you please check if this works?
> > It looks good. The only thing is that the bridge line should be able to handle a new line at the end of the bridge configuration.  If I go to [bridges.torproject.org/options](https://bridges.torproject.org/options) and copy the two bridge lines and paste it into the custom bridges, there is a new line at the very end which causes the regex to say the config is invalid. If I remove the new line at the end, then it works, but it should probably not have to do that.

How do newlines really work? If it is possible that the user input may have more than one, I think it makes more sense to first trim all the newlines and then pass it to the regex. This way, we can keep the regex focused on the actual parameters. Though in that case, we are doing shadow transformations on the input, so I don't know if that should be done or not. WDYT?

syphyr:
I think the best way to handle this is to only apply the regex to non-empty lines (lines with one or more visible characters).

if ( line != null ) { apply regex checks }

meenbeese:
I added a filter for empty and blank lines before passing it to the regex, see if this approach works.

n8fr8:
Okay, just tested, and it does, and all is well. Thank you!

PR #1259, opened 2025-02-27
Implement search bar to app chooser page

### LINKED ISSUE #1006: feat: Add search bar to Orbot "Choose apps" list
**Is your feature request related to a problem? Please describe.**
When using the Orbot app on Android, it can be difficult to find the app you want to use Orbot within the "Choose apps" list, especially if you have a lot of apps installed.

**Describe the Solution You'd Like**
I would like to request the addition of a search bar to the "Choose apps" list. This would make it much easier to find the app you want to use Orbot with, regardless of how many apps you have installed.

**Additional Context**
A search bar would be a simple and effective way to improve the usability of the "Choose apps" list. It would be easy for users to learn how to use it, and it would save them time and effort when using Orbot.

I hope you will consider adding this feature to Orbot. Thank you for your time and consideration.

meenbeese:
This PR adds a search bar to the top of the app manager activity so that users can easily search through their apps if they have a lot of them installed on their device.

This PR also:
- Implements caching of app hashes
- Implements async coroutines loading
- Implements serialization using Gson
- Breaks up and simplifies the logic

Screenshots:
<table>
    <tr>
        <td>
		</td>
        <td>
		</td>
        <td>
		</td>
    </tr>
</table>

Closes #1006 

Tested on Pixel 8 API 34 Emulator.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -118,6 +118,7 @@ dependencies {
     implementation(libs.androidx.navigation.fragment.ktx)
     implementation(libs.androidx.navigation.ui.ktx)
     implementation(libs.converter.gson)
+    implementation(libs.gson)
     implementation(libs.retrofit)
     implementation(libs.rootbeer.lib)
 
--- a/app/src/main/java/org/torproject/android/ui/AppManagerActivity.kt
+++ b/app/src/main/java/org/torproject/android/ui/AppManagerActivity.kt
@@ -5,24 +5,29 @@ import android.Manifest
 import android.content.Context
 import android.content.Intent
 import android.content.SharedPreferences
-import android.content.pm.ApplicationInfo
 import android.content.pm.PackageManager
 import android.os.Bundle
+import android.text.Editable
+import android.text.TextWatcher
+import android.view.LayoutInflater
 import android.view.Menu
 import android.view.MenuItem
 import android.view.View
-import android.view.View.OnFocusChangeListener
 import android.view.ViewGroup
-import android.widget.ArrayAdapter
+import android.view.inputmethod.InputMethodManager
 import android.widget.CheckBox
-import android.widget.GridView
 import android.widget.ImageView
-import android.widget.ListAdapter
 import android.widget.ProgressBar
 import android.widget.TextView
 
 import androidx.appcompat.app.AppCompatActivity
-import androidx.core.content.ContextCompat
+import androidx.recyclerview.widget.LinearLayoutManager
+import androidx.recyclerview.widget.RecyclerView
+
+import com.google.android.material.textfield.TextInputEditText
+import com.google.android.material.textfield.TextInputLayout
+import com.google.gson.Gson
+import com.google.gson.reflect.TypeToken
 
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
@@ -36,9 +41,6 @@ import org.torproject.android.service.OrbotConstants
 import org.torproject.android.service.util.Prefs
 import org.torproject.android.service.vpn.TorifiedApp
 
-import java.util.Arrays
-import java.util.StringTokenizer
-
 class AppManagerActivity : AppCompatActivity(), View.OnClickListener, OrbotConstants {
     inner class TorifiedAppWrapper {
         var header: String? = null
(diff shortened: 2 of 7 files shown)


n8fr8:
Is this backwards compatibility to existing apps people have selected?

meenbeese:
> Is this backwards compatibility to existing apps people have selected?

I checked, and yes, the existing apps list is persisted.

meenbeese:
Also replaced Gson with kotlinx-serialization-json since it is faster and designed for Kotlin.

meenbeese:
This PR has become too big, I will try to break it up.

bitmold:
Two issues I've found with this:

First, type an app in partially, say "Chrom" for Chrome, and you'll see that the list gets filtered down. Then rotate the device and you'll notice that "Chrom" is still in the search field, but the apps are no longer filtered 

Secondly, there used to  be filtering out of some internal system apps, with mater even after merging in #1290. Afterwards, this filtering no longer occurs. (On this random brand junk phone I use for debugging, some internal apps still show, but as you can see, *everything shows* with this code applied:

Before:
Screenshot_20250520-074434


After:
open
Screenshot_20250520-073618

meenbeese:
> Two issues I've found with this:
> 
> First, type an app in partially, say "Chrom" for Chrome, and you'll see that the list gets filtered down. Then rotate the device and you'll notice that "Chrom" is still in the search field, but the apps are no longer filtered
> 
> Secondly, there used to be filtering out of some internal system apps introduced here, which still functioned normally even merging in #1290. With this code, the filtering no longer occurs. (On this random brand junk phone I use for debugging, some internal apps still show, but as you can see, _everything shows_ with this code applied:
> 
> Before: Screenshot_20250520-074434
> 
> After: Screenshot_20250520-073618

Tried to fix both of those issues, thanks for testing! PTAL.

bitmold:
Can you post screenshots please

meenbeese:
> Can you post screenshots please

Sure, the PR looks like this:
<table>
    <tr>
        <td>
		</td>
        <td>
		</td>
    </tr>
</table>

Also the master branch seems to be broken now:
<table>
    <tr>
        <td>
		</td>
        <td>
		</td>
        <td>
		</td>
        <td>
		</td>
    </tr>
</table>

meenbeese:
Opened #1298 to fix the layout issues on master.

bitmold:
Were you able to reproduce and then fix the issues I had found? Been very busy so haven't ran your new code but since there's so much force pushing i can't really tell what's changed since last time by looking at diffs

meenbeese:
Yes i added code to refilter when the orientation is changed and restored the old `getApps` function to filter out the system apps. Previously on the emulator 116 apps were displayed and now only 47 are.

bitmold:
Unfortunately this still doesn't work because:

- it reintroduces the really egregious top margin back to the activity that was taken off the main branch (see screenshot above) 

I also found another bug that's not introduced by your code, but you're welcome to tackle it here if you'd like:

- when you check an app, and then rotate the device (whether there is a search query or not), the app becomes unchecked. Assume this goes both ways too, where if you uncheck a previously checked app it probably rechecks it when you rotate.

bitmold:
This was specifically with your rebsaed commit `38195b53942f950be8537454db2b724d4faadbf7`. I think on some devices, this doesn't really show up render whereas on others it does. The margins being put it in the first place, and how my _fix_ brought the issue you brought forward above. In any case I'm excited to merge this, but not will not until it looks proper

bitmold:
In the future when you get feedback on PRs can you please commit additional commits rather than force pushing the same commit over and over. It makes it so I can't tell what's you merely rebasing vs what's new code that you've introduced into the PR after receiving feedback on github

Merging in one commit can be nice in the end, once it's good to go if a PR has too many commits either you or I can squash them, but until we get to that point it makes reviewing harder + slower because I can't quickly see what's changed between points A and B.

meenbeese:
Will rebase and try to revive this PR.

meenbeese:
Should be good for an initial review. Also way fewer changes than last time. @bitmold

bitmold:
Working on integrating this work now, so will merge it shortly. But for one thing there's a bug, I imagine its in onResume() or some similar logic in which the searched apps are repopulated with duplicates 

image

Here's a blurry webcam pic of 3 Signals, since I didn't feel like changing settings to allow myself to take a proper screenshot in Orbot

bitmold:
In fact, everytime onResume is called, the entire `uiList` variable (all apps and additionally elements for textviews and stuff) will get increased in size.

bitmold:
Thanks for this work, I merged it while adding:

```
commit 078af14aeddbd44153b50f74f7d6a659dc7b7389 (HEAD -> master, origin/master, origin/HEAD, meenbeese_search)
Author: a <dsnake@protonmail.com>
Date:   Wed Nov 5 16:34:41 2025 -0500

    clear app search textfield when theres text

commit 1f967df9a588644aa6cb411088dbd0a2654575db
Author: a <dsnake@protonmail.com>
Date:   Wed Nov 5 16:11:05 2025 -0500

    added function to normalize search quieries when searching for apps with accent marks

commit 0170856a98f421a809d98423c528c9a69a09d292
Author: a <dsnake@protonmail.com>
Date:   Wed Nov 5 16:06:50 2025 -0500

    Fix duplicate bug with app manager activity reworking in #1259, renamed variable uiList to be more clear, added small top margin to new search bar


```

bitmold:
The second commit removes accent marks from things you're searching for, for instance if you want to search for an app called "Échecs" searching "echecs" (without the accent) should return that app

Issue #1263, opened 2025-03-12
[BUG] App is regularly killed by the OS due to incorrect foreground service type


t-m-w:
**Describe the Bug**
In Android 15, as of 17.4.1-BETA-4-tor-0.4.8.13, Orbot stops running on a regular basis, and it may fail to automatically start later as well. I believe it was fine in the previous version in the F-Droid repo.

**To Reproduce**
1. Install and use Orbot without any customizations - or, switch to Power User Mode.

**Expected Behavior**
Orbot stays connected until you disconnect.

**What Custom Configuration Do You Use?**
It happens in Power User Mode or in standard Full Device VPN mode.

**Screenshots**
N/A

**Smartphone (please complete the following information):**
- Device: Pixel 6 Pro
- OS: Android 15 QPR1
- Version: 17.4.1-BETA-4-tor-0.4.8.13

**Crash Logs (Advanced)**
```
AndroidRuntime: android.app.RemoteServiceException$ForegroundServiceDidNotStopInTimeException: A foreground service of type dataSync did not stop within its timeout: ComponentInfo{org.torproject.android/org.torproject.android.service.OrbotService}
```

```
AndroidRuntime: android.app.ForegroundServiceStartNotAllowedException: Time limit already exhausted for foreground service type dataSync
```

(Sorry that these are just short snippets but they should give the gist; let me know if you need more.)

**Additional Context**
A colleague believes this has started happening because of the increase to the target SDK level, which brings [timeouts for the `dataSync` foreground service type](https://developer.android.com/about/versions/15/behavior-changes-15#datasync-timeout).

The VPN service could use `systemExempted` instead of `dataSync` to avoid this. For the proxy, some other apps use `specialUse` and write in a justification in `android:value` so that Google Play will approve it.

cc: @theimpulson

rustybird:
Why not revert targetSdk to 34 in the meantime as an interim fix?

n8fr8:
Will consider both solutions and issue a fix this week.

n8fr8:
[FOREGROUND_SERVICE_TYPE_SYSTEM_EXEMPTED](https://developer.android.com/reference/android/content/pm/ServiceInfo#FOREGROUND_SERVICE_TYPE_SYSTEM_EXEMPTED) should work since Orbot is a VPN app. Have attempted to implement this.

n8fr8:
@meenbeese @syphyr thoughts?

meenbeese:
[...]

PR #1249, opened 2025-02-02
Center the Orbot activity vertically

meenbeese:
I consider 9440e723f29b9b282474b0a192f96811a588d12e to be a regression which makes the UI essentially unusable on smaller devices and tablets. Moreover, it adds too much padding and shifts the UI an excessive amount from the previous releases. This PR aims to center the UI and provide a more comfortable middle ground.

<table>
    <tr>
        <th>Before</th>
        <th>After</th>
    </tr>
    <tr>
        <td></td>
        <td></td>
    </tr>
    <tr>
        <td></td>
        <td></td>
    </tr>
    <tr>
        <td></td>
        <td></td>
    </tr>
</table>

Closes #1261 

Tested on Pixel 8 API 34.

PR #1275, opened 2025-04-16
implementing FOREGROUND_SERVICE_TYPE_SYSTEM_EXEMPTED to stop service …

n8fr8:
…kill for #1263

### DIFF
--- a/app/src/main/res/layout/activity_orbot.xml
+++ b/app/src/main/res/layout/activity_orbot.xml
@@ -2,6 +2,7 @@
 <androidx.constraintlayout.widget.ConstraintLayout
         android:layout_width="match_parent"
         android:layout_height="match_parent"
+        android:layout_marginTop="140dp"
         android:backgroundTint="@color/new_background"
         xmlns:android="http://schemas.android.com/apk/res/android"
         xmlns:app="http://schemas.android.com/apk/res-auto">
--- a/app/src/main/res/layout/activity_orbot.xml
+++ b/app/src/main/res/layout/activity_orbot.xml
@@ -1,23 +1,50 @@
 <?xml version="1.0" encoding="utf-8"?>
 <androidx.constraintlayout.widget.ConstraintLayout
-        android:layout_width="match_parent"
-        android:layout_height="match_parent"
-        android:layout_marginTop="140dp"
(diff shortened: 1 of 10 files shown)


n8fr8:
Working well for me after overnight testing

syphyr:
> Working well for me after overnight testing

Would this not break older devices?

orbotservice/src/main/java/org/torproject/android/service/OrbotService.java
```diff
@@ -229,6 +231,8 @@ else if (mCurrentStatus.equals(STATUS_ON)) {
         }
 
         startForeground(NOTIFY_ID, mNotifyBuilder.build());
```
syphyr, line 233: shouldn't this line be commented out?

n8fr8: yikes, good catch. not sure how that happened.

Using ServiceCompat should allow it to work on older devices.

app/src/main/res/layout/activity_orbot.xml
```diff
@@ -2,6 +2,7 @@
 <androidx.constraintlayout.widget.ConstraintLayout
         android:layout_width="match_parent"
         android:layout_height="match_parent"
+        android:layout_marginTop="140dp"
```
syphyr, line 5: A recent change like this 9440e723f29b9b282474b0a192f96811a588d12e caused a regression in the UI for lower resolution devices. This change fixed the 140dp margin by doing this: https://github.com/guardianproject/orbot-android/pull/1249/files

I think a similar fix like above is also required here.

app/src/main/res/layout/layout_apps.xml
```diff
@@ -5,7 +5,9 @@
     xmlns:tools="http://schemas.android.com/tools"
     android:background="@color/new_background"
     android:orientation="vertical"
-    android:duplicateParentState="true">
+    android:duplicateParentState="true"
+    android:paddingTop="140dp"
```
syphyr, line 9: Same here requires previous suggestion.

n8fr8: Well, the issue is on higher resolution the UI overlaps the actionbar

PR #1279, opened 2025-04-20
Fix Orbot showing it is disconnected on device rotation

### LINKED ISSUE #1278: [BUG] ConnectFragment displays Orbot is disconnected after rotating device
**Describe the Bug**
There is a bug in the UI that causes `ConnectFragment` to incorrectly display that Orbot is disconnected. Connect to the Tor network, then select either the Kindness or More fragment and then rotate the device. If the Kindness or More Fragment is not selected after connecting to the Tor network, then rotating the device does not make Orbot show that it is disconnected. Orbot is actually still connected to the Tor network when it says it is disconnected. After `ConnectFragment` incorrectly shows the status of the connection, switching back to Kindness or More Fragment and then back again to `ConnectFragment` then shows that the device is connected to Tor Network again.

**To Reproduce**
1. Enable rotation in settings
2. Connect to Tor
3. Select Kindness or More Fragment
4. Rotate device
5. Orbot now displays it is disconnected 

**Expected Behavior**
Orbot should not display it is disconnected when it is actually connected.

No custom configuration

**Screenshots**

- Device: T813
- OS: cm-14.1
- Version: 17.4.1-BETA-4-tor-0.4.8.13-64-g2eb68419

**Crash Logs (Advanced)**

**Additional Context**

syphyr:
Fixes Orbot showing it is disconnected when device is rotated.

closes #1278

### DIFF
--- a/app/src/main/java/org/torproject/android/OrbotActivity.kt
+++ b/app/src/main/java/org/torproject/android/OrbotActivity.kt
@@ -32,6 +32,7 @@ import org.torproject.android.ui.LogBottomSheet
 
 class OrbotActivity : BaseActivity() {
 
+    private lateinit var bottomNavigationView: BottomNavigationView
     private lateinit var logBottomSheet: LogBottomSheet
     lateinit var fragConnect: ConnectFragment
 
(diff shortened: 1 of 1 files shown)


syphyr:
I thought this fixed the issue, but there seems to be another case. If I start orbot and connect to tor, then rotate the device, Orbot still shows that it is connected.  But, If I start Orbot and connect to tor, then click on another tab, such as More, then rotate the device, Orbot shows it is disconnected.  There is something about switching fragments after connecting that causes the connection fragment to display it is disconnected after rotation.

PR #1280, opened 2025-04-21
Improve navigation handling with rotation

### LINKED ISSUE #1278: [BUG] ConnectFragment displays Orbot is disconnected after rotating device

meenbeese:
Fixes #1278 

Also, I discovered and fixed another issue during the process where immediately after rotation, it was not possible to go to a different tab, or at least go back to the previous selected tab.

Tested on Pixel 8 API 34.

### DIFF
--- a/app/src/main/java/org/torproject/android/ConnectFragment.kt
+++ b/app/src/main/java/org/torproject/android/ConnectFragment.kt
@@ -49,14 +49,13 @@ class ConnectFragment : Fragment(), ConnectionHelperCallbacks,
     lateinit var progressBar: ProgressBar
     private lateinit var lvConnectedActions: ListView
 
-    private var lastStatus: String? = ""
+    private val lastStatus: String
+        get() = (activity as? OrbotActivity)?.previousReceivedTorStatus ?: ""
 
     override fun onAttach(activity: Activity) {
         super.onAttach(activity)
 
         (activity as OrbotActivity).fragConnect = this
-        lastStatus = activity.previousReceivedTorStatus
-
     }
 
     override fun onCreateView(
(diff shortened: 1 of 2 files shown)


meenbeese:
@syphyr please review, thanks.

syphyr:
I just tested this. Unfortunately, I am still seeing the bug.

syphyr:
With or without Power User Mode did not make any difference.

meenbeese:
> I just tested this. Unfortunately, I am still seeing the bug. Maybe the problem is that I am using power user mode.
> 
> It seems like `lastStatus` is not getting updated correctly.
> 
> Interestingly, I tried removing this line and something still called `doLayoutOff()`:
> 
> ```
> --- a/app/src/main/java/org/torproject/android/ConnectFragment.kt
> +++ b/app/src/main/java/org/torproject/android/ConnectFragment.kt
> @@ -98,7 +98,6 @@ class ConnectFragment : Fragment(), ConnectionHelperCallbacks,
>                  OrbotConstants.STATUS_STARTING -> doLayoutStarting(requireContext())
>                  OrbotConstants.STATUS_ON -> doLayoutOn(requireContext())
>                  OrbotConstants.STATUS_STOPPING -> {}
> -                else -> doLayoutOff()
>              }
>          }
>      }
> ```

Ok, I investigated some more and looks like we need to save more state. See if it works now.

syphyr:
Great work! All the rotation bugs are fixed now. Also, the bug when switching between tabs is also fixed (sometimes they would get stuck after rotation). Thanks!

syphyr:
I think I found an additional bug in the rotation.  If I select Connect Frag and rotate from portrait to landscape to portrait.. then select Kindness Frag and rotate from portrait to landscape to portrait, then select More Frag and rotate portrait to landscape to portrait... Orbot will start to show the wrong Fragment name.

meenbeese:
Wow, how do you even discover a bug like that? I think my last fix resolves it though, because I haven't been able to reproduce it after that.

syphyr:
I was testing with your latest revision. It appears that Orbot starts displaying the incorrect tab name after the fourth tab switch + rotation.

bitmold:
very nice :star: :fire:

PR #1281, opened 2025-04-21
Convert ClientAuthActivity into a Fragment

meenbeese:
Avoids unnecessary configuration associated with activities and simplifies the app.

Testing is needed!!


meenbeese:
This PR is ready to go! 🚀

bitmold:
This is a big feature some users rely on. I know you said testing is needed, wondering what testing remains to be done. Can you demonstrate that you can connect to an onion service that needs client authentication? And after that, can you demonstrate that all the backup features and db manipulations still work as intended. 

I'm always very very reluctant to change large swaths of the application in order to get some of it into Kotlin if all that's happening is that the app is "simplified". Making changes like this does also remove all the git history from the old java app, which as a longtime maintainer actually unsimplifies the app for me when debugging and troubleshooting code someone (in this case me) wrote a few years ago.

Recently I tried to isolate functionality from OrbotService by extracting specific parts of the class that weren't critical to the interfacing with TorService + the VPN running into other classes (the new circumvention and db packages for example). I was then able to convert these smaller classes into Kotlin.

Also I just did a bit of work + refactoring on the UI side of the java app, since this PR is from a long time ago I'd reckon that  those changes aren't reflected here since there's technically not a merge conflict (the java file was deleted and replaced with a new kotlin file)

I for real appreciate this effort though and am not averse to merging it, but want to see it working myself before I take the time to test the nitty gritty of it.

bitmold:
For this, the easiest way to test is to run `tor` on a machine to host an onion service that points to say an HTTP server with an empty webpage and run `openssl pkey` to create authentication for it 

https://community.torproject.org/onion-services/advanced/client-auth/

then in Orbot use those credentials.

bitmold:
Closing for now, open to revisiting this when there's detailed testing of the features still working

meenbeese:
> Closing for now, open to revisiting this when there's detailed testing of the features still working 

Maybe you can help with the testing @syphyr ? I don't expect to have broken anything in this PR though...

PR #1282, opened 2025-04-22
Redesign MoreFragment with a responsive grid layout

meenbeese:
Improves the more fragment by:
1. Implementing a more material style
2. Increasing the touch targets
3. Refreshing the old colours
4. Making the icons bigger
5. Increasing the spacing

<table>
    <tr>
        <th>Before</th>
        <th>After</th>
    </tr>
    <tr>
        <td>
</td>
        <td></td>
    </tr>
</table>
<table>
    <tr>
        <th>Before</th>
        <th>After</th>
    </tr>
    <tr>
        <td></td>
        <td></td>
    </tr>
</table>

### DIFF
--- a/app/src/main/java/org/torproject/android/MoreFragment.kt
+++ b/app/src/main/java/org/torproject/android/MoreFragment.kt
@@ -1,5 +1,7 @@
 package org.torproject.android
 
+import android.animation.AnimatorSet
+import android.animation.ObjectAnimator
 import android.app.Activity
 import android.content.Intent
 import android.content.pm.PackageManager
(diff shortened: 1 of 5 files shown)


syphyr:
For consistency, maybe you can also do the same thing for the Connect Fragment.

meenbeese:
There's some more considerations for that, and it's a more critical part of the app. If this one works well and we don't get complaints, I can implement it for the connect fragment as well in a different PR.

meenbeese:
Pushed new changes to center the text, fix margins and sizes, and add an animation to Orbie.

meenbeese:
Pushed new changes to center the text, fix margins and sizes, and add a jiggle animation to Orbie.

syphyr:
The Orbie jiggle is a nice touch.

bitmold:
I like this a lot, very dope! Can't think of anything off of the top of my head right now, but it'd be so much better if there wasn't an odd number of items leaving that awkward bottom row. 

I thought about removing Choose Apps since you see it on the main screen when you're connected, but there's a use case where users may want to configure which apps get tunneled before/between tor connections. idrk right now :woman_shrugging:  :woman_shrugging:  :woman_shrugging: 

I think moving Settings and then Log to the top also makes sense, since most users aren't spinning up or authing into onion services. For an eighth item i was thinking about popping a link open to the Orbot website or to check.torproject.org but this would run the risk of deanonymizing a user who isn't connected.... 

Maybe an 8th item could be something to invoke the ExitNode Dialog (which normally appears on main screen when connected)? 

Maybe something that opens the device's VPN settings page for Orbot so users can more trivially set always/on and block connections without VPN android features ???

bitmold:
@meenbeese cherry-picked your commit because I wanted to make some modifications, but your work has been pushed to master as `269799ae1a6123952ed59b0e128f7dd5438915bf` - thanks a lot !

PR #1283, opened 2025-04-23
Finally remove last commons.gradle file

meenbeese:
This was making the Gradle config unnecessarily complicated and was a roadblock towards moving to Kotlin DSL.


n8fr8:
I don't understand why it is bad to share common config values in a shared file? Now we have the same settings copy and pasted multiple times.

bitmold:
Going to not merge this, I know Orbot TV doesn't get a lot of love right now, but duplicating this logic unnecessarily complicates things and is going to cause someone a headache down the line when they change one file and aren't immediately penalized for making the change in the other file. 

I appreciate the push to Kotlin DSL though, and thankfully this kind of thing isn't actually a road block to migrating over to it since it supports this exact same feature 

https://stackoverflow.com/questions/66688929/how-to-apply-plugin-from-root-directory-in-kotlin-dsl

meenbeese:
I have a few reasons to remove the shared config file:

1. The API to import files into Gradle and I couldn't get it to work with Kotlin DSL even after so many attempts. At this point, I am not sure if it is even possible because I haven't seen any good examples of projects that have done it or official documentation on it for that matter, while I have a working version right now with this file removed. So if you have managed to do it @bitmold, please contribute a PR.
2. The duplicated logic doesn't really matter because it is trivial. For example, we don't have a TV build and I am not sure if we will ever have one because the code there is very messy and kind of antiquated. Even if it worked at some point, it surely won't work today and I have no interest in fixing or working on a code base that maybe 5 people will use.

meenbeese:
I have a few reasons to remove the shared config file:

1. The API to import files into Gradle is terrible and I couldn't get it to work with Kotlin DSL even after so many attempts. At this point, I am not sure if it is even possible because I haven't seen any good examples of projects that have done it or official documentation on it for that matter, while I have a working version right now with this file removed. So if you have managed to do it @bitmold, please contribute a PR.
2. The duplicated logic doesn't really matter because it is trivial. For example, we don't have a TV build and I am not sure if we will ever have one because the code there is very messy and kind of antiquated. Even if it worked at some point, it surely won't work today and I have no interest in fixing or working on a code base that maybe 5 people will use.

bitmold:
1. Kotlin DSL is pretty low on my priority list right now, and instead want to continue bringing stability, security and usability performances out to users. I haven't managed to do it because I haven't tried to, so no PR coming from me about this anytime soon :woman_shrugging: 
2. I see your point, but for the time being not interested in duplicating *even more* logic between the TV app and the main app for a gradle feature that's not particularly important to me right now. If down the line Guardian Project wants to salvage the TV app for whatever reason this would just amount to there being yet another point of divergence. Love your enthusiasm and contributions, makes sense you don't feel interested in working on the TV app!

PR #1284, opened 2025-05-01
Add setting to keep isolated circuits alive

syphyr:
Add setting to enable KeepAliveIsolateSOCKSAuth to keep alive isolated circuits that have at least one stream with SOCKS authentication active. After such a circuit is idle for more than MaxCircuitDirtiness seconds, it can be closed.

This option is default for Tor Browser.

### DIFF
--- a/app-tv/src/main/res/values/strings.xml
+++ b/app-tv/src/main/res/values/strings.xml
@@ -181,6 +181,8 @@
     <string name="pref_isolate_port_summary">Use a different circuit for each destination port</string>
     <string name="pref_isolate_protocol">Isolate client protocols</string>
     <string name="pref_isolate_protocol_summary">Use a different circuit for each connecting protocol</string>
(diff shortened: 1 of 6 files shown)


syphyr:
It should be noted somewhere that it is very important to include a random username and password for each proxy configuration for Orbot.  Orbot itself does not need to require a username and password to connect to the socks proxy.  But, when the client submits a random username and password to Orbot, the circuit gets special isolation.  This is exactly how Tor Browser isolates its own circuits.

bitmold:
Going to merge this feature in since it's usable in how it modifies torrc and to begin the process of getting translations of the preference. Thanks! 

Curious how to begin to approach what you had said about the SOCKS proxy authentication and circuit isolation in your last comment, @n8fr8 @syphyr do you have any ideas?

syphyr:
I would use the definition of KeepAliveIsolateSOCKSAuth from `man tor`

PR #1290, opened 2025-05-14
Rewrite TorifiedApp in Kotlin and add serialization

meenbeese:
Broke off from #1259 so that said PR will not be huge.


orbotservice/src/main/java/org/torproject/android/service/vpn/TorifiedApp.kt
```diff
@@ -0,0 +1,120 @@
...
+                    app.usesInternet = true
+                }
+
+                if (!app.usesInternet) {
+                    continue
+                } else {
+                    apps.add(app)
+                }
+
+                app.isEnabled = aInfo.enabled
```
bitmold, line 94: In future when doing these Kotlin conversions please try to use more Kotlin-esque features like `apply` for instance when doing lots of assignments on objects like this

PR #1291, opened 2025-05-15
Update release script for snowflake and meek configs

syphyr:
The configuration settings for snowflake and meek are located in the fronts file. The snowflake-brokers file now only contains common configuration settings in order to support both snowflake and snowflake amp, so it does not require any modifications.


syphyr:
@bitmold I've updated your script. Please let me know what you think.

syphyr:
There is an additional change I have to add to this MR so that snowflake-brokers is also kept updated.

PR #1292, opened 2025-05-16
Make sure snowflake-brokers is current

syphyr:
It's possible that something could change in snowflake-brokers common configuration, so lets make sure that is checked as well.


meenbeese:
Can we move this script to under /scripts? It would keep things tidier imho.

PR #1305, opened 2025-05-23
Bump Retrofit and switch to Kotlin converter

meenbeese:
Eliminates Gson dependency because we already have kotlinx-serialization-json dependency in our app.

I tested the moat bridges connection and it seems to work fine.


meenbeese:
Rebased on latest master.

PR #1308, opened 2025-05-25
Dialog and permission logic needed for power user mode regarding #1263

bitmold:
For a detailed description of why this code is added, please see my comments in #1263 

This PR:

- Adds the `SCHEDULE_EXACT_ALARMS` permission to the manifest 
- Implements a dialog for power user mode to ensure that this permission is granted at runtime before `OrbotService` is started in power user mode.

a
b
c


tladesignz:
How do we go about Samsung testing? Should we merge and have @ingfabby grab a nightly build, or should we build a special version and let her use that?

n8fr8:
Regarding Samsung testing, I can share our BrowserStack account with anyone who needs access. Nightly builds get pushed there, and you can test on various Android models.

Issue #1264, opened 2025-03-14
Spoof notification content


d4f5409d:
**Is your feature request related to a problem? Please describe.**
Yes, you can't hide Orbot notification otherwise Android kills it from running it in the background.

**Describe the Solution You'd Like**
The pitch is to hide the fact that you are using Orbot and the Tor network and make a switch to enable some kind of duress mode where you can replace `connected to the tor network` string and the orbot logo to something else like the `sun is shining`

**Describe the Alternatives You've Considered**
Not yet

**Additional Context**
Signal made a feature to allow their app to disguise as something else: 
https://support.signal.org/hc/en-us/articles/6325632279834-Change-the-Signal-app-icon-on-your-phone
https://www.androidpolice.com/signal-update-hide-app-icon-name/

n8fr8:
I actually designed that feature in Signal, and provided the code to them. It is true we should provide this in Orbot. Thanks for the thought and will get it done, both for the icon and for the notifications.

d4f5409d:
Glad I could help the community!

n8fr8:
We have this code in Circulo: https://gitlab.com/circuloapp/circulo-keanu-android/-/tree/main/circuloapp/src/main/java/org/article19/circulo/next/settings/camouflage?ref_type=heads

n8fr8:
Also hiding the app from Recent Apps

There might be some activity settings we could play with: https://stackoverflow.com/questions/20516832/exclude-current-activity-from-recent-tasks

android:excludeFromRecents="true"
https://developer.android.com/guide/components/activities/recents

bitmold:
Notably on API 26+  you can programmatically alter the application label set in the `<application>` tag of the manifest.

bitmold:
Closing, this has been implemented

PR #1312, opened 2025-05-31
Camouflage Mode for #1264

bitmold:
Uses the icons and library found in Circulo 

- Adds new setting for opening up a Camouflage Mode 
- User can select from several disguise apps 
- When in camouflage mode, app icon, label and notification text are hidden 
- *note that the expanded notification text will still read "Orbot", however if the notification is minimized it just reads the name of the camouflaged app*

I recorded a GIF of this feature in action but it's too big to host on GitHub so I uploaded it here: https://imgur.com/a/RxmtwFB


bitmold:
This PR partially addresses #870

tladesignz:
Damn. Right after merging it, I found a problem:

When app is in Camouflage mode, a tap on the permanent notification doesn't open the app anymore. I guess, this is due to using another activity-alias when camouflaged.

We could also see this as intentional, so unauthorised people inspecting the phone don't have it so easy to look at the app. For a normal user, it's a UI nuisance, though.

bitmold:
Oh dang! hmmm. This was unintentionally implemented that way, but I could actually see it as a _feature_ in its own right. The fix for this could be behind an additional preference in the settings which I think should default to true: "Prevent the Camouflaged Notification from Opening Orbot"

PR #1314, opened 2025-05-31
Setting to Lock Access to app UI with Device PIN 

bitmold:
- This uses the device PIN to prevent users from entering OrbotActivity. Prevents someone who steals phone from entering Orbot and meddling with settings, looking at onion service credentials, etc. 
- When this setting is set, notifications still run, the user just has to enter their password in order to enter the app itself. 

This code is dependent on #1312 The Camouflage Mode PR and should only be merged once that branch is merged. I'm setting up this PR against the camo branch instead of master in order to diff of this feature.

**This feature integrates tightly with camo mode in #1312** If the user has camouflaged Orbot to be another app, this app locking setting uses the camouflaged display name to hide that the app behind the lock is Orbot. Camouflage mode also obscures notification content, so if  both of these settings are enabled a Tor VPN can be run pretty discretely.


bitmold:
untitled
Password failed and success

Note that if the user backs out of the password prompt `finish()` is called and the app closes. While the password prompt is open `OrbotActivity`s root View's visibility is set to `INVISIBLE` such that someone can't quickly see the UI of Orbot while `finish()` is being called.

bitmold:
untitled

Here Orbot is camouflaged as the "Assistant" app. If the user backs out the password lock prompt there is no visible UI. When they enter their PIN Orbot opens as normal

bitmold:
This PR partially addresses #870

bitmold:
Merging in for testing with new camo feature

tladesignz (commented):
Tested on Nothing Phone(1) running Android 15. Looking fine.

tladesignz:
I suggest extending this, so the user could choose, if they want to be forced to use the device code or biometry (that is fingerprint, on most devices.)

bitmold:
That's true, my thinking was that it's easier for a hostile actor to (physically) force someone to put your thumb on the device than it is to make them enter/give up their password. It's trivial to add that flag to this though, we could do it.

@n8fr8 curious on your thoughts with this

tladesignz:
> That's true, my thinking was that it's easier for a hostile actor to (physically) force someone to put your thumb on the device than it is to make them enter/give up their password. It's trivial to add that flag to this though, we could do it.

And in many jurisdictions, there's a big difference between the police being able to force you to hand over your finger to unlock the phone and them making you obliged to give up a password.

That should be made clear to the user, when we give them that possibility to decide.

However, if you e.g. want to defend against a nosy partner, the passphrase might just be a little too much and the fingerprint would instead be the right balance between security and usability.

Hence the suggestion.

bitmold:
Yeah you're totally right. will do it that way!

PR #1317, opened 2025-06-03
Fix potential crash in setPreferenceForSmartConnect()

cohosh:
Setting both circumventionApiBridges and circumventionApiIndex as class variables could result in a crash if the list of bridges shrinks unexpectedly on a subsequent call to setPreferenceForSmartConnect(). Since circumventionApiBridges is not accessed outside of this function, this makes it a function input rather than a class variable.

Consider the following example:
- a User selects "ASK TOR" and [`CircumventionApiManager.getSettings`](https://github.com/guardianproject/orbot-android/blob/cd946c986a4debb8cd824e72f4f677357d2c8e86/app/src/main/java/org/torproject/android/ConfigConnectionBottomSheet.kt#L237C9-L237C76) returns a `List<Bridges?>` of size 2, which is set in `circumventionApiBridges`. `setPreferenceForSmartConnect()` is called and `circumventionApiIndex` is set to 1.
- a User selects "ASK TOR" again and circumventionApiIndex is set to 2.
- a User selects "ASK TOR" again and this time, `CircumventionApiManager.getSettings` returns a `List<Bridges?>` of size 1. Since the check [`it.size == circumventionApiIndex`](https://github.com/guardianproject/orbot-android/blob/cd946c986a4debb8cd824e72f4f677357d2c8e86/app/src/main/java/org/torproject/android/ConfigConnectionBottomSheet.kt#L289) is false, there will be an [out of index access](https://github.com/guardianproject/orbot-android/blob/cd946c986a4debb8cd824e72f4f677357d2c8e86/app/src/main/java/org/torproject/android/ConfigConnectionBottomSheet.kt#L297) that will cause a crash

Using class variables also makes this unsafe code to be called concurrently, which could happen in theory. I'm not sure of the reason for updating the index on each call, so it would be good to check that always using the 0 index (which should correspond to the top priority bridge settings) is the desired behaviour here. Another potential way to solve this is to have a more robust check of `circumventionApiIndex` against the size of the current list.

### DIFF
--- a/app/src/main/java/org/torproject/android/ConfigConnectionBottomSheet.kt
+++ b/app/src/main/java/org/torproject/android/ConfigConnectionBottomSheet.kt
@@ -201,9 +201,6 @@ class ConfigConnectionBottomSheet :
         if (pref.equals(Prefs.PATHWAY_DIRECT)) rbDirect.isChecked = true
     }
 
-    private var circumventionApiBridges: List<Bridges?>? = null
-    private var circumventionApiIndex = 0
-
     private fun askTor() {
 
         val dLeft = AppCompatResources.getDrawable(requireContext(), R.drawable.ic_faq)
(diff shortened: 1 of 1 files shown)


bitmold:
This looks fine to me, thanks for noticing and fixing this.I'm not sure why those were class variables to begin with, around the release of Orbot v17 - which was a major redesign - there was a lot of refactoring and perhaps this was unfortunately something untidy that stuck around from that process.

PR #1324, opened 2025-06-10
Use current list of defined bridge types

syphyr:
There is no "snowflake-amp" bridge type defined anymore. These undefined bridge types are left over from the older version of IPtProxy.


meenbeese:
I think this was an old/unused file and it is now removed from master, so this PR can be closed.

bitmold:
@meenbeese is right, I'm closing this. 

However this kind of check is definitely important and something to keep an eye out for, it's unideal how there are currently two sets of checks for bridge types between the app and orbotservice projects

Issue #1323, opened 2025-06-10
[Feature Request] Add support for Meek-Azure


syphyr:
**Is your feature request related to a problem? Please describe.**

Orbot used to support MeekLite bridges, but support was lost when IPtProxy was recently upgraded. The function `bridgesEnabled()` checks if the locale is Farsi and automatically sets MeekLite as the default bridge, even though it is undefined now. Tor Browser has support for Meek-Azure, so lets add support for it as well. The bridge line used in Tor Browser is:

`meek_lite 192.0.2.20:80 url=https://1314488750.rsc.cdn77.org front=www.phpmyadmin.net utls=HelloRandomizedALPN`

**Describe the Solution You'd Like**
Add support for Meek-Azure bridge

**Describe the Alternatives You've Considered**

**Additional Context**

syphyr:
@bitmold I looked at the code and I'm not sure what the best way is to implement this.  IMHO, Meek-Azure should just be an option when setting up a custom bridge, same as snowflake bridges.  There is only one known working meek-azure bridge, the same one that Tor Browser uses. So, the Meek-Azure bridge is not a bridge that users would manually enter in as a custom bridge.

Since Tor Browser lets users manually select Meek-Azure as a bridge, I dont think this should be an issue if it was also added to Orbot.

syphyr:
In CustomBridgeBottomSheet.kt, regex should also support meek bridges. Currently, it is only checking for obfs4 and webtunnel.

https://github.com/guardianproject/orbot-android/blob/f26ee8e7c31eccf46abd2e43b485ad0e9e4c31a8/app/src/main/java/org/torproject/android/CustomBridgeBottomSheet.kt#L18

syphyr:
@meenbeese When you added the regex check in CustomBridgeBottomSheet.kt, we only tested obfs4 and webtunnel.  Could you also look at meek support using the bridge line from the first post here?

tladesignz:
I will take care of this next week and streamline with our Apple implementations.

bitmold:
FWIW `bridgesEnabled` is actually an unused function

tladesignz:
Related: #1196, #1102

meenbeese:
> [@meenbeese](https://github.com/meenbeese) When you added the regex check in CustomBridgeBottomSheet.kt, we only tested obfs4 and webtunnel. Could you also look at meek support using the bridge line from the first post here?

[...]

PR #1328, opened 2025-06-11
Add Meek Azure support

syphyr:
Add support for Meek Azure bridge used by Tor Browser.

Fixes: #1323


syphyr:
This commit also fixes ability to add meek azure bridge to Orbot as a custom bridge because when parsing the custom bridges,  "utls=" was undefined.

syphyr:
Added regex support for meek_lite bridge when validating format in custom.

syphyr:
Rebased against latest changes to master branch

tladesignz:
Sorry. Instead of patching our way forward on a pretty wild-grown code to configure bridges, I'm currently working on a complete rewrite, which mirrors the work we did on Orbot Apple fully, has separation-of-concerns and a straightforward way to updated bridge info, and which you can watch here:
https://github.com/tladesignz/orbot/tree/bridge_refurbish

Thank you for your effort and sorry, that it won't end up in Orbot.

Issue #1335, opened 2025-06-16
[BUG] Samsung Galaxy M13 NPE on AppManager


n8fr8:
**Describe the Bug**
Error log / stack trace from Google Play

**To Reproduce**
Error log from Google Play

**Expected Behavior**
No crash

**Smartphone (please complete the following information):**

Samsung Galaxy M13
Android 14 (SDK 34)
Version: 1750300101 (17.5.0-RC-1-tor-0.4.8.16)

**Crash Logs (Advanced)**
```
Sample attributes:
(https://play.google.com/console/u/0/developers/6147505171125588874/app/4973609425637044610/vitals/crashes/5c146882f8952177e3b230eae97fb8d4/details?days=28&isUserPerceived=true&versionCode=1750300101%2C1750300102#)
[Android 14 (SDK 34)](https://play.google.com/console/u/0/developers/6147505171125588874/app/4973609425637044610/vitals/crashes/5c146882f8952177e3b230eae97fb8d4/details?days=28&isUserPerceived=true&versionCode=1750300101%2C1750300102#)
Version: 1750300101 (17.5.0-RC-1-tor-0.4.8.16)Occurred: 3 hours ago
Type
java.lang.NullPointerException
Generate crash insights with Gemini
Exception java.lang.NullPointerException:
  at org.torproject.android.ui.AppManagerActivity.saveAppSettings (AppManagerActivity.kt:228)
  at org.torproject.android.ui.AppManagerActivity.onOptionsItemSelected (AppManagerActivity.kt:87)
  at android.app.Activity.onMenuItemSelected (Activity.java:4696)
  at androidx.activity.ComponentActivity.onMenuItemSelected (ComponentActivity.kt:461)
  at androidx.fragment.app.FragmentActivity.onMenuItemSelected (FragmentActivity.java:264)
  at androidx.appcompat.app.AppCompatActivity.onMenuItemSelected (AppCompatActivity.java:269)
  at androidx.appcompat.view.WindowCallbackWrapper.onMenuItemSelected (WindowCallbackWrapper.java:109)
  at androidx.appcompat.app.AppCompatDelegateImpl.onMenuItemSelected (AppCompatDelegateImpl.java:1235)
  at androidx.appcompat.view.menu.MenuBuilder.dispatchMenuItemSelected (MenuBuilder.java:833)
  at androidx.appcompat.view.menu.MenuItemImpl.invoke (MenuItemImpl.java:157)
[...]

PR #1337, opened 2025-06-16
Fixed issue #1335: Replaced all force-unpacks (!!) with proper null checks (?).

tladesignz:
These force-unpacks are probably in other places and need to go eventually.

### DIFF
--- a/app/src/main/java/org/torproject/android/ui/AppManagerActivity.kt
+++ b/app/src/main/java/org/torproject/android/ui/AppManagerActivity.kt
@@ -229,10 +229,13 @@ class AppManagerActivity : BaseActivity(), View.OnClickListener {
     }
 
     private fun saveAppSettings() {
+        val allApps = allApps ?: return
+        val suggestedApps = suggestedApps ?: return
+
         val tordApps = StringBuilder()
         val response = Intent()
 
-        for (tApp in allApps ?: emptyList()) {
+        for (tApp in allApps) {
             if (tApp.isTorified) {
                 tordApps.append(tApp.packageName)
                 tordApps.append("|")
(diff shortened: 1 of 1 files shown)


bitmold:
It looks good, I'm testing it now though. there are some force unpacks that make sense in the app but mostly it's not idiomatic and makes bugs like this sneak in. I think the most vulnerable part of the app to this kind of crash is in classes that model JSON API responses, where data may sometimes be present and othertimes not...

bitmold:
I'm curious exactly *how* this NPE happened, the offending line is the `allApps!!` in the for loop, it gets init'd to a real object in `loadApps()` but it would seem that this crash was caused by the user invoking save apps before the apps were loaded, which is odd.

```kt
private fun saveAppSettings() {
        val tordApps = StringBuilder()
        val response = Intent()
        for (tApp in allApps!!) {
            if (tApp.isTorified) {
                tordApps.append(tApp.packageName)
                tordApps.append("|")
                response.putExtra(tApp.packageName, true)
            }
        }
```

bitmold:
Trivial to reproduce if you do something like this 

```kt
private fun loadApps() {
        Timer().schedule(object : TimerTask() {
            override fun run() {
                loadApps2()
            }

        }, 5000)
    }

    private fun loadApps2() {
   // ... original function
}
```

I've still yet to run your fix with this `Timer` test put into place, but I'm worried that your implementation might right out empty data since it defaults to empty lists if they're not instantiated. In the event that the lists aren't instantiated I think we'd want to abort the save operation.

tladesignz:
Ah, yeah true. I was so fixated on the ugly force-unpacks that I forgot about the context.

bitmold:
Looks good!

PR #1339, opened 2025-06-16
Make geoip files gen a Gradle build task

meenbeese:
This PR replaces the manual script-based process for generating `geoip.jar` with a Gradle task, improving automation and consistency in the build process.

Key Changes:
- Added a Gradle task to clone the Tor repo, extract geoip files, and package them into geoip.jar under src/main/assets.
 - Hooked the task into the preBuild phase alongside copyLicenseToAssets.
- Removed the old shell script (libs/build-geoip-jar.sh) and committed binary (libs/geoip.jar).
- Updated .gitignore to exclude the generated JAR.

Benefits:
- geoip.jar is built automatically during the Gradle build.
- Ensures consistent builds across environments and contributors.
- Eliminates committed binaries and legacy scripts.


syphyr:
The issue I see with this is that the newly generated geoip db will not be used unless the `versionCode` is also updated.

meenbeese:
This process is tied to new releases, where the version code is always updated right?

syphyr:
The current process is to publish latest libraries (for Tor, IPtProxy, etc), bump the versionCode and update the tag, then build and release.  Perhaps the better approach here is to also publish the geoip database and add this change:
https://github.com/syphyr/orbot/commit/30f9c524994da67d68783063a8b2ed64885fabb6

meenbeese:
@n8fr8 is this approach feasible?

bitmold:
Does this git clone tor every single time the project is built?

syphyr:
It looks like it

meenbeese:
Changed to only run when building release app.

bitmold:
But then on a fresh clone of the repository there would be no geoip files, no?

meenbeese:
> But then on a fresh clone of the repository there would be no geoip files, no?

Right, I added a commit to mitigate this edge case. With that:
- In release mode -> Files are always rebuilt
- In debug mode and no `geoip.jar` -> Files are rebuilt
- In debug mode and `geoip.jar` exists -> Files are not rebuilt

bitmold:
This is looking better for me, but now I'm thinking I never really use release builds when I'm debugging and am a pretty forgetful person. Local copies of orbot that reoccuring developers use will have an out of date version of these jars and will now have to manually delete the jars in order to get them to update on their local machine.

Cloning it each time isn't the answer, but I'm not sure this is either

bitmold:
Also will this fail silently on release builds if any of the steps `git clone` or others fail, or will the build continue to be made. Orbot *can* ship without geoip jars, so if these aren't included in the app it wouldn't break per se, there'd just be a loss of functionality. 

I like the spirit of this project, but unless it can be *perfect* than i don't see much utility in this, it's not a serious development cost to manually run the script whenever there's a new release.

bitmold:
I feel very hesitant about putting anything nontrivial into the release APK as part of the build process. 

I think that this would also mean that two release APKs built at different times with no changes to the code, will potentially have different artifacts in them. Orbot is a ways away from being built reproducibly, but even without this goal in mind it seems pretty undesirable to build the same code from the same commit at two different points in time and potentially have two different APKs with fundamentally different data in them.

meenbeese:
Yeah I understand your concerns. Maybe we can tweak and revisit the build process in the future though...

bitmold:
My priorities are primarily focused on:

- implementing pressing new features 
- closing existing bugs
- ensuring the app doesn't break with future Android API releases and NDK releases (there is some work to do with upgrading the native code infrastructure)

If there are parts of the build process that can be improved seamlessly I'm open to it, but that's not where my energy is and there would need to be a pretty compelling reason to spend time testing and integrating the new changes. 

With this focus on stability in the medium term I'm less interested in process changes and changes to libraries that may modernize things or have cool bells and whistles but could (and sometimes do) break things in subtle ways. 

 Like i said earlier this is a cool idea and I appreciate you thinking outside of the box here.  I personally won't be spending time on this in the future, but you're welcome to. I love the contributions you've made that close existing UI bugs, and also some of the really cool UI stuff like with the more screen. Those are the kinds of things i have the most bandwidth for and enthusiasm to review, test and ultimately merge

PR #1352, opened 2025-06-23
Replace go-tun2socks with hev-socks5-tunnel C library

bitmold:
Work in progress for now, just putting this here to showcase that things are feasible and to ask for input if I get stuck debugging this further.

- Configures Orbot service to build the submodule with NDK (all handled within Android Studio) 
- adds the native methods from hev-socks5-tunnel into `OrbotVpnManager` class
- Removes OrbotLib and old go code referencing the tun2socks implementation, switches to directly using the maven-hosted IPtProxy artifact


bitmold:
So far, there's some kind of race condition or something preventing it from working. 

I can resolve DNS requests to check.torproject.org just fine to 116.202.120.181, however google chrome set to go through Orbot stays loading and eventually gives up when I try to load https://check.torproject.org

When I put a breakpoint on this [line to write processed IP data back out to the VPN file descriptor](https://github.com/guardianproject/orbot-android/pull/1352/files#diff-72e79c1e7b70841e61823714bd77d0abd6ed1166663fea3f6969b20bad5dc5e1R276) and step through it with the debugger at a slower-ish pace, the page successfully loads with "your browser is configured to use Tor" so I need to do more debugging around there.

Also SOCKS username/password renames to be implemented.

orbotservice/src/main/jni/Application.mk
```diff
@@ -0,0 +1,6 @@
+APP_OPTIM := debug
```
bitmold, line 1: This should be release

orbotservice/src/main/java/org/torproject/android/service/vpn/OrbotVpnManager.java
```diff
@@ -188,18 +198,8 @@ private synchronized void setupTun2Socks(final VpnService.Builder builder) {
             // https://developer.android.com/reference/android/net/VpnService.Builder#setMetered(boolean)
             if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
                 builder.setMetered(false);
-
```
bitmold, line 191: @n8fr8 was there a reason these allow family calls were only on a >= Q API check?

orbotservice/build.gradle
```diff
@@ -7,10 +7,25 @@ plugins {
...
 
     defaultConfig {
         minSdk = 24
         targetSdk = 35
+        ndk {
+            abiFilters "armeabi-v7a", "arm64-v8a", 'x86', 'x86_64'
+        }
+        externalNativeBuild {
+            cmake {
+                arguments "APP_CFLAGS+=-DPKGNAME=org/torproject/android/service/ -ffile-prefix-map=${rootDir}=."
```
bitmold, line 20: This may be superfluous, will check.

orbotservice/build.gradle
```diff
@@ -7,10 +7,25 @@ plugins {
 android {
     namespace = "org.torproject.android.service"
     compileSdk = 35
+    ndkVersion "26.3.11579264"
```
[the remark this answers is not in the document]
syphyr: I would go with latest LTS

syphyr:
> When I put a breakpoint on this [line to write processed IP data back out to the VPN file descriptor](https://github.com/guardianproject/orbot-android/pull/1352/files#diff-72e79c1e7b70841e61823714bd77d0abd6ed1166663fea3f6969b20bad5dc5e1R276) and step through it with the debugger at a slower-ish pace, the page successfully loads with "your browser is configured to use Tor" so I need to do more debugging around there.
> 

The race condition seems like the DNS resolver is not resolving fast enough.

tladesignz:
I stumbled over this one: https://hev.cc/posts/2025/tun2socks-mapped-dns/

We can actually make use of that. However, that still doesn't give us resolution of onion services.

Just checked this with the master branch. That *also* doesn't give me onion services. At least not with Firefox Android. Did it ever? What do I actually need to do to reach onion services?

tladesignz:
Nevermind. Firefox Android is just fucked up. Works on Chrome.
So, this seems to work.
I'm not exactly happy with the solution, having this DNS map in between.
And I'm wondering, if these DNS cache settings are actually sensible for our use-case. (They're taken from the hev-socks-tunnel docs.)

What do you think, @bitmold?

bitmold:
I'll take a look soon...

bitmold:
> I'm not exactly happy with the solution, having this DNS map in between.
And I'm wondering, if these DNS cache settings are actually sensible for our use-case. (They're taken from the hev-socks-tunnel docs.)

Haven't yet had a look to dig deeper into this. 


So far just dealt with merging around the recent refactoring of orobtservice so I can get your changes to run on top of the latest commits in master. At a glance, can see that this implementation works, that's excellent! But yeah, need to investigate the specifics of the DNS Cache settings and haven't otherwise rigorously tested these changes yet.

app/src/main/jni/Application.mk
```diff
@@ -0,0 +1,6 @@
+APP_OPTIM := debug
```
bitmold, line 1: @tladesignz this should be changed to `release` once we're happy with the implementation and ready to merge...

bitmold:
Well I wish there was more documentation around the mapdns feature. One thing that wasn't clear about the cache size feature is that it it doesn't refer to caching onto disk, but with heap allocated memory the lib mallocs.

```c
static int
mapped_dns_init (void)
{
    HevMappedDNS *dns;
    int cache_size;
    int network;
    int netmask;

    network = hev_config_get_mapdns_network ();
    netmask = hev_config_get_mapdns_netmask ();
    cache_size = hev_config_get_mapdns_cache_size ();

     if (!cache_size)
        return 0;

    dns = hev_mapped_dns_new (network, netmask, cache_size);
    if (!dns)
        return -1;

    hev_mapped_dns_put (dns);

    return 0;
}
```

tladesignz:
> Well I wish there was more documentation around the mapdns feature. One thing that wasn't clear about the cache size feature is that it it doesn't refer to caching onto disk, but with heap allocated memory the lib mallocs.

 ```c
struct _HevMappedDNSNode
{
    HevRBTreeNode tree;
    HevListNode list;
    char *name;
    int idx;
};

struct _HevRBTreeNode
{
    unsigned long __parent_color;
    HevRBTreeNode *right;
    HevRBTreeNode *left;
} __attribute__ ((aligned (sizeof (long))));

struct _HevListNode
{
    HevListNode *next;
    HevListNode *prev;
};

```

Well, it's 10,000 * `HevMappedDNSNode`. I'm not great at identifying C structure memory usage, esp. since it seems self-referencing. How much could that be? 100 - 200 bytes?

So, that's 10,000 * 200 = 2.000.000 / 1024 / 1024 = 1.9 MByte.

Is that a problem for our users?

File: orbotservice/.gitignore [hunk 6]
```diff
@@ -0,0 +1,2 @@
+/build
+jniLibs
```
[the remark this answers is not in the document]
bitmold: this has been resolved, but thanks

app/build.gradle.kts
```diff
@@ -83,6 +83,13 @@ android {
         }
     }
 
+    ndkVersion = "28.2.13676358"
+    externalNativeBuild {
+        ndkBuild {
+            path = file("src/main/jni/Android.mk")
+        }
+    }
+
```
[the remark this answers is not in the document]
bitmold: Yeah, it's one of the rare things that works actually works pretty damn well within the gradle/android developer ecosystem

bitmold:
> So, that's 10,000 * 200 = 2.000.000 / 1024 / 1024 = 1.9 MByte.  
>  Is that a problem for our users?

This shouldn't be, no. Hopefully in a future release hev-socks5-tunnel will just allow us to use tor's DNS port and we can move past this. But for now, I'm satisfied with it. And it's nice to finally be able to use Orbot (albeit without kindness mode and non-vanilla bridges) on a modern Android with memory tagging enabled. Hopefully gomobile finally fixes this soon, but not holding my breath on that.... Very exciting to finally move to purely using IPtProxy too. 

So I think we're good as far as this PR goes? I changed the compile flag for hev-socks5  from `debug` to `release` so we're compiling with the lighter/streamlined version of the library. If you don't think there are any other things left to implement please merge this once you've had a last look at it. 

Oh, and thanks again for helping me finish this too! I very much appreciated your fresh eyes on this after getting stuck on that DNS issue on the older hev-socks release.

tladesignz:
> This shouldn't be, no. Hopefully in a future release hev-socks5-tunnel will just allow us to use tor's DNS port and we can move past this. 

Tried to ask nicely: https://github.com/heiher/hev-socks5-tunnel/issues/261

> So I think we're good as far as this PR goes? I changed the compile flag for hev-socks5 from `debug` to `release` so we're compiling with the lighter/streamlined version of the library. If you don't think there are any other things left to implement please merge this once you've had a last look at it.

Will do in a minute!

> Oh, and thanks again for helping me finish this too! I very much appreciated your fresh eyes on this after getting stuck on that DNS issue on the older hev-socks release.

This was great team work! You figured out a lot of things which I would have struggled hard with. Thank you!

PR #1355, opened 2025-06-24
Encapsulates Logic for using Meek to Talk to Moat

bitmold:
When the ask tor button is used to talk to the MOAT API, a global HTTP auth was set based on the url and front.

This was **never** turned off, which is not desired and could in theory be leaky.
    
This logic was also all in the UI fragment where the AskTor button lived. Made an object that lets you spin up meek for using moat that handles configuring this authentication when starting meek for this purpose and also a function to clear this value when you stop meek.
    
Noticed this when I realized that Meek never gets shutdown on API error in b489e13ad7b5e80329e7709ae9412a5463df04a7


tladesignz:
As you know, I'm working on a complete rewrite on all bridge-related code. It mirrors what we did in Orbot Apple and contains said encapsulation.

Thank you for pushing this, and sorry, that this code won't end up in Orbot Android, since it would be removed right away anyway.

You can watch progress here: https://github.com/tladesignz/orbot/tree/bridge_refurbish

PR #1360, opened 2025-06-30
Bridge refurbish

tladesignz:
Let's start the discussion!


tladesignz:
Probably also @syphyr and @meenbeese might be interested in this.

Please note: This is a pretty exact match what we did in Orbot Apple. *Including* having a UX expert on board.

Orbot Android did copy a lot of the UI, but unfortunately not the rest of the implementation.

I'm not willing to discuss any questions like "can you add this or that?", "why doesn't it have x?", etc.
The discussion here should be about any bugs you might find and *eventually* about the code design. Although, I consider that code design way better than the mess I saw here until now, so you'll have a hard time convincing me to change a lot.

BUT: When we're done merging this, I'm happy to discuss further improvements, which I will then also mirror in Orbot Apple.

What this adds:
- Meek support
- Telegram and Email options
- built-in Obfs4 bridges
- automatic updates of built-in bridges and custom bridges, when we get them through the "Ask Tor" feature.

What's still missing: Smart connect. Will work on this right away.

app/src/main/java/org/torproject/android/util/RequirePasswordPrompt.kt
bitmold, line 1: pretty sure this was added back in by mistake. We switched to DeviceAuthenticationPrompt

app/src/main/java/org/torproject/android/Constants.kt
```diff
@@ -0,0 +1,14 @@
+package org.torproject.android
+
+import androidx.core.net.toUri
+
+object Constants {
```
bitmold, line 5: minor, since all of these constants pertain to bridges maybe rename this Object

app/src/main/java/org/torproject/android/Constants.kt
```diff
@@ -0,0 +1,14 @@
...
+
+    val bridgesUri = "https://bridges.torproject.org/".toUri()
+
+    const val emailRecipient = "bridges@torproject.org"
```
bitmold, line 9: getBridgesEmailRecipient

orbotservice/src/main/java/org/torproject/android/service/circumvention/AutoConf.kt
```diff
@@ -0,0 +1,157 @@
...
+        var response = api.settings(MoatApi.SettingsRequest(country))
+
+        val error = response.errors?.firstOrNull()
+        if (error != null) {
+            if (error.code == 404 /* Needs transport, but not the available ones */
+                || error.code == 406 /* no country from IP address */) {
+                cannotConnectWithoutPt = true
+            }
+            else {
+                throw error
```
bitmold, line 66: in the event that this error is caught somewhere down the line wouldn't meek still be running

do_release_housekeeping.sh
```diff
@@ -1,5 +1,9 @@
...
-./update_snowflake_bridges.sh
+#!/usr/bin/env sh
+
+BASE=$(dirname "$0")
+
+# Update geoip files from tor.
+sh "$BASE/libs/build-geoip-jar.sh"
+
+# Update built-in bridges.
+sh "$BASE/scripts/update-bridges.sh"
```
syphyr, line 9: This `update-bridges.sh` script does not support json

orbotservice/src/main/assets/fronts
```diff
@@ -1,13 +1,3 @@
-snowflake-target https://1098762253.rsc.cdn77.org/
-snowflake-front www.cdn77.com,www.phpmyadmin.net
-snowflake-stun stun:stun.antisip.com:3478,stun:stun.epygi.com:3478,stun:stun.uls.co.za:3478,stun:stun.voipgate.com:3478,stun:stun.mixvoip.com:3478,stun:stun.nextcloud.com:3478,stun:stun.bethesda.net:3478,stun:stun.nextcloud.com:443
 snowflake-target-direct https://snowflake-broker.torproject.net/
```
syphyr, line 1: Keeping these variables in one location was easier to maintain. I've noticed some of these are now moved into various locations.

orbotservice/src/main/java/org/torproject/android/service/circumvention/BuiltInBridges.kt
```diff
@@ -0,0 +1,88 @@
...
+            if (instance == null && context != null) {
+                try {
+                    instance = read(getUpdateFile(context).readText())
+                }
+                catch (_: Throwable) {}
+            }
+
+            if (instance == null && context != null) {
+                try {
+                    instance =
```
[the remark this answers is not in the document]
tladesignz: No. The logic flow is:
1. Try to read an update file from app-cache.
2. If that fails, read the compiled-in asset file.
3. If that fails, well... That shouldn't happen, but at least don't crash anyway.

do_release_housekeeping.sh
tladesignz: Not sure why that would be important. Can you clarify?

syphyr:
Will this PR require an updated version of IPtProxy to test meek-azure support?

orbotservice/src/main/assets/builtin-bridges.json
```diff
@@ -0,0 +1 @@
...2C08904527FEA90C4C4F8C1083EA59FBC6FAF cert=ItvYZzW5tn6v3G4UnQa6Qz04Npro6e81AP70YujmK/KXwDFPTs3aHXcHp4n8Vt6w/bv8cA iat-mode=0","obfs4 193.11.166.194:27015 2D82C2E354D531A68469ADF7F878FA6060C6BACA cert=4TLQPJrTSaDffMK7Nbao6LC7G9OW/NHkUwIdjLSS3KYf0Nv4/nQiiI8dY2TcsQx01NniOg iat-mode=0"],"snowflake":["snowflake 192.0.2.4:80 8838024498816A039FCBBAB14E6F40A0843051FA fingerprint=8838024498816A039FCBBAB14E6F40A0843051FA url=https://1098762253.rsc.cdn77.org/ fronts=www.cdn77.com,www.phpmyadmin.net ice=stun:stun.antisip.com:3478,stun:stun.epygi.com:3478,stun:stun.uls.co.za:3478,stun:stun.voipgate.com:3478,stun:stun.mixvoip.com:3478,stun:stun.nextcloud.com:3478,stun:stun.bethesda.net:3478,stun:stun.nextcloud.com:443 utls-imitate=hellorandomizedalpn","snowflake 192.0.2.3:80 2B280B23E1107BB62ABFC40DDCC8824814F80A72 fingerprint=2B280B23E1107BB62ABFC40DDCC8824814F80A72 url=https://1098762253.rsc.cdn77.org/ fronts=www.cdn77.com,www.phpmyadmin.net ice=stun:stun.antisip.com:3478,stun:stun.epygi.com:3478,stun:stun.uls.co.za:3478,stun:stun.voipgate.com:3478,stun:stun.mixvoip.com:3478,stun:stun.nextcloud.com:3478,stun:stun.bethesda.net:3478,stun:stun.nextcloud.com:443 utls-imitate=hellorandomizedalpn"]}
```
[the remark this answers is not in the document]
tladesignz: Uuuuuh! That's a very good hint. That is a bug in the Apple version, too. Thank you!

orbotservice/src/main/java/org/torproject/android/service/circumvention/Bridge.kt
```diff
@@ -0,0 +1,185 @@
...
+            }
+
+            ice?.let {
+                if (it.isNotEmpty()) params.add("ice=$it")
+            }
+
+            utlsImitate?.let {
+                if (it.isNotEmpty()) params.add("utls-imitate=$it")
+            }
+
```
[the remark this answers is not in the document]
tladesignz: Aaaaah! Again, great catch! Thank you!

tladesignz:
> Will this PR require an updated version of IPtProxy to test meek-azure support?

It shouldn't. But YMMV. I couldn't start it with my setup. Unclear, ATM, what was the reason.

tladesignz:
Ok, late here. Please continue. I will fix all your findings tomorrow and finish Smart Connect!

orbotservice/src/main/java/org/torproject/android/service/circumvention/AutoConf.kt
```diff
@@ -0,0 +1,158 @@
...
+     *
+     * @param country: An ISO 3166-1 Alpha 2 country code to force a specific country.
+     * 	 If not provided, the MOAT service will deduct a country from your IP address (preferred!)
+     * @param cannotConnectWithoutPt: Set to `true`, if you are sure, that a PT configuration *is needed*,
+     * 	 even though the MOAT service says, that in your country none is. In that case, a default configuration will be used.
+     * @return A transport to use and a list of custom bridges, if any could be evaluated.
+     *
+     * @throws MoatApi.MoatError or maybe other IO exceptions.
+     */
+    suspend fun `do`(context: Context, country: String? = null, cannotConnectWithoutPt: Boolean = false): Pair<Transport, List<String>>? {
```
meenbeese, line 23: nit: reformat
```suggestion
    suspend fun `do`(
        context: Context,
        country: String? = null,
        cannotConnectWithoutPt: Boolean = false
    ): Pair<Transport, List<String>>? {
```

app/src/main/res/values/strings.xml
```diff
@@ -310,7 +311,11 @@
...
     <string name="invalid_bridge_format">Invalid bridge format</string>
     <string name="error_asking_tor_for_bridges">Error asking Tor how to connect</string>
+    <string name="enter_bridge_address">Enter bridge address</string>
+    <string name="built_in_bridges_obfs4">Built-in Bridges (obfs4)</string>
+    <string name="built_in_bridges_obfs4_subtitle">Cloaks your traffic. Gets around some Tor blocking. Good if you\'re on public WiFi, but in a country where Tor isn\'t blocked.</string>
     <string name="unlock_app_msg" formatted="true">Please unlock your device in order to use %s</string>
     <string name="pref_password_only_no_biometric">Disallow Biometrics for Orbot Authentication</string>
     <string name="pref_password_only_no_biometric_description">This forces Orbot Authentication to only require your device\'s password even if biometric authentication, such as fingerprint or face unlock, are configured on your device.</string>
+    <string name="run_smart_connect_to_find_the_best_way">Run Smart Connect to find the best way</string>
```
meenbeese, line 320: We should probably shorten this string

app/src/main/java/org/torproject/android/Constants.kt
meenbeese: I vote to move them over to OrbotConstants.

meenbeese (commented):
Great changes so far! Added a few review comments that I saw.

orbotservice/src/main/java/org/torproject/android/service/circumvention/AutoConf.kt
tladesignz: Done.

app/src/main/java/org/torproject/android/ui/connect/ConfigConnectionBottomSheet.kt
```diff
@@ -1,339 +1,280 @@
...
-            }, {
-                Log.wtf(TAG, "Couldn't hit circumvention API... $it")
-                if (isVisible) {
-                    Toast.makeText(
-                        requireContext(),
-                        R.string.error_asking_tor_for_bridges,
-                        Toast.LENGTH_LONG
-                    ).show()
-                    proxy.stop(IPtProxy.MeekLite)
+                    delay(5 * 1000)
```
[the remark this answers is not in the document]
tladesignz: No. On principle. Code UX.

orbotservice/src/main/java/org/torproject/android/service/circumvention/BuiltInBridges.kt
```diff
@@ -0,0 +1,88 @@
...
+
+        fun isOutdated(context: Context): Boolean {
+            val lastModified = try {
+                getUpdateFile(context).lastModified()
+            }
+            catch (_: Throwable) {
+                0
+            }
+
+            return Calendar.getInstance().timeInMillis - lastModified > 2 * 24 * 60 * 60 * 1000
```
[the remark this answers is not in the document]
tladesignz: Really? That's why I leave the calculation to the compiler to make it more clear. But ok, there you are.

app/src/main/java/org/torproject/android/Constants.kt
tladesignz: Then that it is.

app/src/main/java/org/torproject/android/Constants.kt
```diff
@@ -0,0 +1,14 @@
...
+
+    const val emailSubjectAndBody = "get transport"
```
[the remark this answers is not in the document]
tladesignz: Done.

tladesignz:
I'm a little unsure about memory management in the `SmartConnect.handle()` `connectionGuard` `Timer`.

Before I start thinking about how to test this and reading up on Android memory management, I'll ask you folks first:

The created `TimerTask` holds references to the `startTor`, `stopTor` and `completed` lambdas. Does cancelling the `Timer` destroy the `TimerTask` properly, or will it leak, because it still holds the lambda references? Do I need to make them weak somehow?

syphyr:
Regarding smart connect, Tor Browser uses a json file that defines specific bridges based on region. Is that something Orbot will be doing as well?

tladesignz:
> Regarding smart connect, Tor Browser uses a json file that defines specific bridges based on region. Is that something Orbot will be doing as well?

Didn't know about that. As said:
Let's get this merged, and then think about if, where and how to improve!

tladesignz:
> Regarding smart connect, Tor Browser uses a json file that defines specific bridges based on region. Is that something Orbot will be doing as well?

Well, not currently, and not with this PR.

But, as mentioned: I'm happy to further improve, when this is merged.

Can you point us to the relevant pieces in Tor Browser?

syphyr:
> > Regarding smart connect, Tor Browser uses a json file that defines specific bridges based on region. Is that something Orbot will be doing as well?
> 
> Well, not currently, and not with this PR.
> 
> But, as mentioned: I'm happy to further improve, when this is merged.
> 
> Can you point us to the relevant pieces in Tor Browser?

https://gitlab.torproject.org/tpo/anti-censorship/rdsys-admin/-/blob/main/conf/circumvention.json?ref_type=heads#L13

tladesignz:
Well, that is not Tor Browser, but rdsys. That's the backend of the Moat API. If that's what Moat returns in case of showing up with an IP address of one of these countries, we're fine.

syphyr:
I was told by the Tor people that this is what Tor Browser uses. I assume rdsys gets bundled into the Tor Browser package.

tladesignz:
Of course, via the Moat REST API.

bitmold:
Sorry, this has taken so long. Something dismal and unexpected ended up happening in my community. </3 But looking at this now. 

One thing I noticed immediately upon running this on a virtual 8A (x86_64) running API 36 is that gomobile code immediately caused the app to crash. I don't think that running on API 36 has anything to do with the issue at hand here. 

This wasn't a fresh install of Orbot, as in I had already some previously installed version of the app installed with some random bridge connections configured from a past debugging session. 

```
FATAL EXCEPTION: pool-5-thread-1
Process: org.torproject.android.debug, PID: 6202
go.Universe$proxyerror: failed to initialize : no such method
    at IPtProxy.Controller.start(Native Method)
    at org.torproject.android.service.circumvention.Transport.start(Transport.kt:219)
    at org.torproject.android.service.OrbotService$IncomingIntentRouter.run(OrbotService.java:792)
```

```kt
for (transport in transportNames) {
            Log.wtf("abc", transport)
            controller.start(transport, "")
        }
```

Putting a log or breakpoint here, the Set `transportNames` contains both "obfs4" and an empty String, it's passing the empty String that causes Orbot to crash here:

The error occurs here on line 438 of `controller.go` https://github.com/tladesignz/IPtProxy/blob/master/IPtProxy.go/controller.go#L438
```go
default:
		// at the moment, everything else is in lyrebird
		t := transports.Get(methodName)
		if t == nil {
			ptlog.Errorf("Failed to initialize %s: no such method", methodName)
			return fmt.Errorf("failed to initialize %s: no such method", methodName)
		}
```

On this Orbot with old bridge data configured, `Prefs.bridgesList` is as follows (whitespace at the end intentional) 
```
obfs4 159.65.70.223:8443 EC7F2570549489EC706F9C8BC1D47B3EB84C2328 cert=/s6+zDPl0D1PSgArGwQtISdZdgbeMVUrijo405petTOEfpqrYEGEhiwSZnlBQSOeOWQFIQ iat-mode=0
obfs4 85.215.237.174:17930 1B25D91413405D766D0E785B61E90F3D6C88C40E cert=SWgDhdyog59UEL4+lzWQoTlvAiVfLocWXRRLbUo4NVu0MAYKORKld82ygRKSbhYb4iBSDQ iat-mode=0
    
``` 
so `Prefs.bridgesList?.split("\n")?.mapNotNull` retursn an arraylist of `["obfs4", "obfs4", ""]`

You can't just `trim()` `Prefs.bridgesList?` because the user may have inserted whitespace somewhere in the middle of the bridge config and not just at the end like in my case. 

Changing line 99 from `CUSTOM -> Prefs.bridgesList?.split("\n")?.mapNotNull { Bridge(it).transport }?.toSet() ?: emptySet()` to `CUSTOM -> Prefs.bridgesList?.split("\n")?.mapNotNull { Bridge(it).transport }?.filter { !TextUtils.isEmpty(it) }?.toSet()` fixes the bug.

bitmold:
In `ConfigConnectionBottomSheet` if you can't ask tor the following Toast is shown.

```kt
catch(e: Throwable) {
                withContext(Dispatchers.Main) {
                    updateAskTorBt()

                    Toast.makeText(context, e.localizedMessage, Toast.LENGTH_LONG)
                        .show()
                }
            }
```

In the past, a Toast was rendered with the string `R.string.error_asking_tor_for_bridges`. This should be changed back or if you're not satisfied with `R.string.error_asking_tor_for_bridges` a different message should be shown because showing the exception text in the UI is an antipattern. Also with `getLocalizedMessagge()` you're not guaranteed to have anything in all of the languages whereas when we use our string resources we have the community giving us something that is sensible.

Currently, in English, when this exception is thrown the user just gets a Toast with the confusing and unhelpful text "Read timed out". Also these things are rarely, if ever "localized" beyond English, par exemple avec Orbot en français il affiche toujours "Read timed out"

image

app/src/main/java/org/torproject/android/ui/connect/ConfigConnectionBottomSheet.kt
```diff
@@ -1,339 +1,288 @@
...
```
bitmold, line 260: Might be wrong, but as far as I know this is essentially sleeping the thread for 5 seconds. Is that what is happening here, and if so, why? The UI be updated as soon as there is a response from the server.

orbotservice/src/main/java/org/torproject/android/service/circumvention/BuiltInBridges.kt
```diff
@@ -0,0 +1,91 @@
...
+
+        /**
+         * We consider stored updated-bridges.json file to be outdated after 2 days.
+         */
+        fun isOutdated(context: Context): Boolean {
+            val lastModified = try {
+                getUpdateFile(context).lastModified()
+            }
+            catch (_: Throwable) {
+                0
```
bitmold, line 66: Not sure why there's a 0 here. You should log an error to make debugging in this in the future easier incase something ever goes wrong with the call to `getUpdateFile`

orbotservice/src/main/java/org/torproject/android/service/circumvention/AutoConf.kt
```diff
@@ -0,0 +1,160 @@
...
+     */
+    suspend fun `do`(
+        context: Context,
+        country: String? = null,
+        cannotConnectWithoutPt: Boolean = false
+    ): Pair<Transport, List<String>>? {
+        var cannotConnectWithoutPt = cannotConnectWithoutPt
+
+        val done = fun(conf: Pair<Transport, List<String>>?): Pair<Transport, List<String>>? {
+            Transport.controller.stop(IPtProxy.MeekLite)
```
bitmold, line 31: Here's a sneaky one, in the event that an exception is thrown such as when the Toast with the localized exception message is displayed that I was talking about on the main thread, this `done` function is **not** invoked. Meek is only turned off when things go smoothly, but never on errors. Verified this with breakpoints on this line and the line to start meek below.

app/src/main/java/org/torproject/android/ui/connect/ConfigConnectionBottomSheet.kt
```diff
@@ -1,339 +1,288 @@
...
-
-        val moatUrl = OrbotService.getCdnFront("moat-url")
-        val front = OrbotService.getCdnFront("moat-front")
+        lifecycleScope.launch(Dispatchers.IO) {
+            val context = context ?: return@launch
 
-        val pUsername = "url=$moatUrl;front=$front"
-        val pPassword = "\u0000"
+            try {
+                val conf = AutoConf.`do`(context, cannotConnectWithoutPt = true)
```
bitmold, line 214: Why is `cannotConnectWIthoutPt` set to `true` here? When I use smart connect, `AutoConf` tells me I don't need any bridges (because I do not) and uses a Direct connection. However when I click "Ask Tor" since this flag is forced to `true` apparently I "need" obfs4 bridges (which I do not) creating an inconsistent experience between the two UI pathways (who is right? smart connect or "ask tor"?). 

Plus people using obfs4 when they don't need to lowers performance, consumes more network resources, and also might dissuade people from using kindenss mode ("oh Tor says I need a bridge to connect, well in that case I can't help other people connect") 

I don't think this *should* be true, but regardless the user is put in a weird position becuase depending on what they click on "Tor"/"Smart Connect" says something different.

bitmold:
I found a bug in `SmartConnect`

Firstly, I forced Smart Connect to use PTs even though I don't need them by setting the cannot connect flag to true.

```kt
ioScope.launch {
            var conf: Pair<Transport, List<String>>? = null

            try {
                conf = AutoConf.`do`(context, cannotConnectWithoutPt = true)
            }
            catch (_: Throwable) {}
```

Then I put breakpoints on every single call to `Prefs.transprort.start()` and `Prefs.transports.stop()` within the `TimerTask` that's created as the argument for `connectionGuard?.schedule()`.

I ran smart connect, and Orbot decided it needed to use obfs4 as you'd imagine and I connected successfully. So good so far. Then when I disconnected tor, I noticed that every so often my breakpoints on the aforemenioned start and stop calls kept being reached in the debugger... ie the `TimerTask` somehow was still running in the background somewhere!

bitmold:
I noticed that with smart connect if I put a breakpoint in the timertask where `stopConnectionGuard()` is terminated when `progress` is >= 100 that code never gets executed.

```kt
if (progress >= 100) {
                        stopConnectionGuard()
                        Prefs.smartConnect = false

                        mainScope.launch {
                            completed()
                        }

                        return
                    }
```

This is probably why TimerTasks continue when you stop tor after successfully connecting. Also of courses this makes it so that when you disconnect after a successful smart connect, the switch to connect via smart connect remains on instead of off. 

If at the top of the TimerTask I add a log statement to print progress, you can see that this never reaches 100 and it just keeps going on every second. Here's a screenshot of Orbot successfully connected via smart connect forcing obfs4 (which happened very fast) and a ton of progress=0 and eventually progress=30 printouts.

image

bitmold:
On another run testing this, I noticed that `progress` would sometimes make it up to 100 (albeit well after Orbot successfully smart connects) and yet the TimerTask keeps marching on

```
2025-07-13 17:54:21.269 12030-12106 abc                     org.torproject.android.debug         E  0
2025-07-13 17:54:21.282 12030-12109 abc                     org.torproject.android.debug         E  0
2025-07-13 17:54:22.269 12030-12106 abc                     org.torproject.android.debug         E  14
2025-07-13 17:54:22.282 12030-12109 abc                     org.torproject.android.debug         E  14
2025-07-13 17:54:23.269 12030-12106 abc                     org.torproject.android.debug         E  30
2025-07-13 17:54:23.282 12030-12109 abc                     org.torproject.android.debug         E  30
2025-07-13 17:54:24.270 12030-12106 abc                     org.torproject.android.debug         E  30
2025-07-13 17:54:24.282 12030-12109 abc                     org.torproject.android.debug         E  30
2025-07-13 17:54:25.270 12030-12106 abc                     org.torproject.android.debug         E  40
2025-07-13 17:54:25.282 12030-12109 abc                     org.torproject.android.debug         E  40
2025-07-13 17:54:26.270 12030-12106 abc                     org.torproject.android.debug         E  45
2025-07-13 17:54:26.282 12030-12109 abc                     org.torproject.android.debug         E  45
2025-07-13 17:54:27.270 12030-12106 abc                     org.torproject.android.debug         E  45
2025-07-13 17:54:27.283 12030-12109 abc                     org.torproject.android.debug         E  45
2025-07-13 17:54:28.271 12030-12106 abc                     org.torproject.android.debug         E  50
2025-07-13 17:54:28.284 12030-12109 abc                     org.torproject.android.debug         E  50
2025-07-13 17:54:29.271 12030-12106 abc                     org.torproject.android.debug         E  50
2025-07-13 17:54:29.283 12030-12109 abc                     org.torproject.android.debug         E  50
2025-07-13 17:54:30.272 12030-12106 abc                     org.torproject.android.debug         E  50
2025-07-13 17:54:30.284 12030-12109 abc                     org.torproject.android.debug         E  50
2025-07-13 17:54:31.272 12030-12106 abc                     org.torproject.android.debug         E  50
2025-07-13 17:54:31.285 12030-12109 abc                     org.torproject.android.debug         E  50
2025-07-13 17:54:32.272 12030-12106 abc                     org.torproject.android.debug         E  50
2025-07-13 17:54:32.285 12030-12109 abc                     org.torproject.android.debug         E  50
2025-07-13 17:54:33.273 12030-12106 abc                     org.torproject.android.debug         E  50
2025-07-13 17:54:33.285 12030-12109 abc                     org.torproject.android.debug         E  50
2025-07-13 17:54:34.274 12030-12106 abc                     org.torproject.android.debug         E  50
2025-07-13 17:54:34.286 12030-12109 abc                     org.torproject.android.debug         E  50
2025-07-13 17:54:35.274 12030-12106 abc                     org.torproject.android.debug         E  50
2025-07-13 17:54:35.286 12030-12109 abc                     org.torproject.android.debug         E  50
2025-07-13 17:54:36.274 12030-12106 abc                     org.torproject.android.debug         E  50
2025-07-13 17:54:36.286 12030-12109 abc                     org.torproject.android.debug         E  50
2025-07-13 17:54:37.274 12030-12106 abc                     org.torproject.android.debug         E  50
2025-07-13 17:54:37.286 12030-12109 abc                     org.torproject.android.debug         E  50
2025-07-13 17:54:38.274 12030-12106 abc                     org.torproject.android.debug         E  50
2025-07-13 17:54:38.286 12030-12109 abc                     org.torproject.android.debug         E  50
2025-07-13 17:54:39.274 12030-12106 abc                     org.torproject.android.debug         E  50
2025-07-13 17:54:39.286 12030-12109 abc                     org.torproject.android.debug         E  50
2025-07-13 17:54:40.275 12030-12106 abc                     org.torproject.android.debug         E  60
2025-07-13 17:54:40.286 12030-12109 abc                     org.torproject.android.debug         E  60
2025-07-13 17:54:41.275 12030-12106 abc                     org.torproject.android.debug         E  95
2025-07-13 17:54:41.287 12030-12109 abc                     org.torproject.android.debug         E  95
2025-07-13 17:54:42.275 12030-12106 abc                     org.torproject.android.debug         E  95
2025-07-13 17:54:42.287 12030-12109 abc                     org.torproject.android.debug         E  95
2025-07-13 17:54:43.276 12030-12106 abc                     org.torproject.android.debug         E  100
2025-07-13 17:54:44.277 12030-12106 abc                     org.torproject.android.debug         E  100
2025-07-13 17:54:45.277 12030-12106 abc                     org.torproject.android.debug         E  100
2025-07-13 17:54:46.277 12030-12106 abc                     org.torproject.android.debug         E  100
2025-07-13 17:54:47.277 12030-12106 abc                     org.torproject.android.debug         E  100
2025-07-13 17:54:48.277 12030-12106 abc                     org.torproject.android.debug         E  100
2025-07-13 17:54:49.278 12030-12106 abc                     org.torproject.android.debug         E  100
2025-07-13 17:54:50.278 12030-12106 abc                     org.torproject.android.debug         E  100
2025-07-13 17:54:51.279 12030-12106 abc                     org.torproject.android.debug         E  100
2025-07-13 17:54:52.278 12030-12106 abc                     org.torproject.android.debug         E  100
2025-07-13 17:54:53.279 12030-12106 abc                     org.torproject.android.debug         E  100
2025-07-13 17:54:54.279 12030-12106 abc                     org.torproject.android.debug         E  100
2025-07-13 17:54:55.280 12030-12106 abc                     org.torproject.android.debug         E  100
2025-07-13 17:54:56.279 12030-12106 abc                     org.torproject.android.debug         E  100
2025-07-13 17:54:57.279 12030-12106 abc                     org.torproject.android.debug         E  100
2025-07-13 17:54:58.279 12030-12106 abc                     org.torproject.android.debug         E  100
2025-07-13 17:54:59.279 12030-12106 abc                     org.torproject.android.debug         E  100
2025-07-13 17:55:00.280 12030-12106 abc                     org.torproject.android.debug         E  100
2025-07-13 17:55:01.279 12030-12106 abc                     org.torproject.android.debug         E  100
2025-07-13 17:55:02.279 12030-12106 abc                     org.torproject.android.debug         E  100
2025-07-13 17:55:03.280 12030-12106 abc                     org.torproject.android.debug         E  100
2025-07-13 17:55:04.281 12030-12106 abc                     org.torproject.android.debug         E  100
2025-07-13 17:55:05.280 12030-12106 abc                     org.torproject.android.debug         E  100
2025-07-13 17:55:06.281 12030-12106 abc                     org.torproject.android.debug         E  100
2025-07-13 17:55:07.280 12030-12106 abc                     org.torproject.android.debug         E  100
2025-07-13 17:55:08.280 12030-12106 abc                     org.torproject.android.debug         E  100
```

bitmold:
So there are essentially two problems:

1. `progress` isn't reliable, it might not make it to 100
2. and if it does, the task doesn't stop so even though this code is run and the settinsg/UI are updated properly, the task is still marching on in the background. 
```kt
 if (progress >= 100) {
                        stopConnectionGuard()
                        Prefs.smartConnect = false

                        mainScope.launch {
                            completed()
                        }

                        return
                    }
```

I found a fix for the second problem, you don't just want to cancel the timer in `stopConnectionGuard()`, you *also* need to purge it. https://stackoverflow.com/a/6976936

```kt
private fun stopConnectionGuard() {
        connectionGuard?.cancel()
        connectionGuard?.purge()
        connectionGuard = null
    }
```


As for the first issue, I think there's something underlying in `TorControlConnection` which is part of the jtorctl library and not actually implemented as part of Orbot. I've noticed for a long time sometimes the progress doesn't always get emitted to 100. Usually this doesn't matter because the progress was just used as part of a UI progress bar which went away once you connected. 

Perhaps you can continue to read the value of `progress` while also checking if OrbotService's status is `STATUS_ON`. 

In OrbotStatus `mCurrentStatus` is reassigned to new values in the class `ActionBroadcastReceiver` pasted below. Perhaps afterwards you could check if `mCurrentStatus.equals(STATUS_ON)` and if so manually ensure the progress is at 100 `SmartConnect.updateProgress(100)`... It's not the most elegant solution but this combined with purging the timer handles these bugs.
```java
case ACTION_STATUS -> {
                    // hack for https://github.com/guardianproject/tor-android/issues/73 remove when fixed
                    var newStatus = intent.getStringExtra(EXTRA_STATUS);
                    if (STATUS_OFF.equals(mCurrentStatus) && STATUS_STOPPING.equals(newStatus))
                        break;
                    mCurrentStatus = newStatus;
                    if (STATUS_OFF.equals(mCurrentStatus)) {
                        showToolbarNotification(getString(R.string.open_orbot_to_connect_to_tor), NOTIFY_ID, R.drawable.ic_stat_tor);
                    }
                   
                    //// Add this here:
                   if (STATUS_ON.equals(mCurrentStatus)) SmartConnect.updateProgress(100);

                    var localStatus = new Intent(LOCAL_ACTION_STATUS).putExtra(EXTRA_STATUS, mCurrentStatus);
                    LocalBroadcastManager.getInstance(OrbotService.this).sendBroadcast(localStatus); // update the activity with what's new
                }
```

bitmold:
The final bug with SmartConnect that I've found is how it keeps working in the background, even with the `purge()` fixes above applied, is that the timer loop keeps looping if you click cancel before it has a chance to succeed or fail. 

Not only does the Timer continue running, but it doesn't seem to be interrupted after the 30 second timeout either. Cancelling can occasionally get the app into some broken and weird states, one of which is that after you click cancel you might see Orbot attempting to reconnect. 

Eventually, it seems like Orbot may sometimes outright crash. Even with no user interaction. (with my logs printing `progress` from above) 

```
2025-07-13 18:26:38.180 13680-13774 abc                     org.torproject.android.debug         E  5
2025-07-13 18:26:39.152 13680-13770 abc                     org.torproject.android.debug         E  5
2025-07-13 18:26:39.153 13680-13771 abc                     org.torproject.android.debug         E  5
2025-07-13 18:26:39.160 13680-13680 Orbot                   org.torproject.android.debug         D  stopTorAsync
2025-07-13 18:26:39.168 13680-0     Go                      org.torproject.android.debug         E  panic: close of nil channel
2025-07-13 18:26:39.169 13680-13770 libc                    org.torproject.android.debug         A  Fatal signal 6 (SIGABRT), code -6 (SI_TKILL) in tid 13770 (Timer-0), pid 13680 (t.android.debug)
2025-07-13 18:26:39.180 13680-13774 abc                     org.torproject.android.debug         E  5
2025-07-13 18:26:39.306 13680-13770 libc                    org.torproject.android.debug         A  Crash due to signal: failed to wait for crash_dump helper: No child processes
```

Seems like pressing cancel should not only interrupt the timer task properly, but also ensure that any transports which might have been started throughout the dance of SmartConnect, stop.

orbotservice/src/main/java/org/torproject/android/service/circumvention/BuiltInBridges.kt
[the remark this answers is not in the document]
tladesignz: Resolved, since no reply for 2 weeks.

orbotservice/src/main/java/org/torproject/android/service/circumvention/BuiltInBridges.kt
```diff
@@ -0,0 +1,88 @@
...
+    companion object {
+
+        const val FILE_NAME = "builtin-bridges.json"
+
+        const val UPDATE_FILE_NAME = "updated-bridges.json"
+
+        private var instance: BuiltInBridges? = null
+
+
+        fun getInstance(context: Context? = null): BuiltInBridges? {
```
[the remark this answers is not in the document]
tladesignz: Resolved, since no reply for 2 weeks.

orbotservice/src/main/java/org/torproject/android/service/circumvention/BuiltInBridges.kt
tladesignz: The only thing this catches is when `lastModified` will throw, because a `SecurityManager` is throwing a `SecurityException`.

That is something that should never happen, and if it does, the only remedy I see is trying to re-load the latest Moat data and hope, we can at least overwrite the file.

I don't see any benefit in logging this, and I'm pretty sure, nobody would ever see such a log entry, anyway.

app/src/main/java/org/torproject/android/ui/connect/ConfigConnectionBottomSheet.kt
tladesignz: No, you're not wrong. This is sleeping the special `withContext(Dispatchers.Main)` thread for 5 seconds.

The logic is:
- When "Ask Tor" succeeds, show green checkmark and bridge type on button.
- Wait 5 seconds, so users can read the visual feedback.
- Finally reset the button to mirror its function again: remove checkmark, set text to "Ask Tor" again.

The latter part was forgotten, when somebody copied the UI from Orbot iOS the first time. Instead, the button showed "Snowflake ✅" until users closed the sheet, but the button functionality stayed the same.

tladesignz:
> so `Prefs.bridgesList?.split("\n")?.mapNotNull` retursn an arraylist of `["obfs4", "obfs4", ""]`

Uuuuh. This is a great catch! Thank you!

I added code to make sure, the bridge list is (more) valid.

> Changing line 99 from `CUSTOM -> Prefs.bridgesList?.split("\n")?.mapNotNull { Bridge(it).transport }?.toSet() ?: emptySet()` to `CUSTOM -> Prefs.bridgesList?.split("\n")?.mapNotNull { Bridge(it).transport }?.filter { !TextUtils.isEmpty(it) }?.toSet()` fixes the bug.

You seriously need to stop using `TextUtils` in Kotlin. 😜

orbotservice/src/main/java/org/torproject/android/service/circumvention/AutoConf.kt
tladesignz: Uuuh. Yeah, that is also a great catch.
I got myself lulled by Retrofit. But, of course, the Retrofit API calls can throw all sorts of exceptions.
I'll catch them and make sure, the Meek connection is closed. Thank you!

tladesignz:
> In `ConfigConnectionBottomSheet` if you can't ask tor the following Toast is shown.
> 
> ```kotlin
> catch(e: Throwable) {
>                 withContext(Dispatchers.Main) {
>                     updateAskTorBt()
> 
>                     Toast.makeText(context, e.localizedMessage, Toast.LENGTH_LONG)
>                         .show()
>                 }
>             }
> ```
> 
> In the past, a Toast was rendered with the string `R.string.error_asking_tor_for_bridges`. This should be changed back or if you're not satisfied with `R.string.error_asking_tor_for_bridges` a different message should be shown because showing the exception text in the UI is an antipattern. Also with `getLocalizedMessagge()` you're not guaranteed to have anything in all of the languages whereas when we use our string resources we have the community giving us something that is sensible.

You seem to have quite strong opinions for matters, which aren't that clear from my point of view.

The mentioned string reads (in English): "Error asking Tor how to connect"

There are 2 pieces of data in it:
1. An error happened
2. when I (the user) tried to ask Tor how to connect.

I say "data", not information, because the second one is actually not information. I pressed that button. Users mostly don't forget that within 20 seconds and they also don't expect any other random error messages to show up from other parts of the system.

So the only information you give them with that is "an error happened".

I, as a user, are regularly pissed at software, which tells me "Ups. Error. Shrug." and no information *at all* what might be the underlying cause or any other hint on how to fix it.

I consider **that** a UX antipattern and developer laziness.

Of course, the best way of solving this would be, to classify each possible error, add explanation texts for each error class and add instructions on how to remedy it.

But, as both of us are lazy (and it seriously is a crazy amount of work, which only the biggest of projects undergo and even then not for *all* cases), that's not, what's going to happen.

Alas, "read timed out" is also not a great message for the uninitiated (although this project has a lot of pro users), I give you that.

So, how about I add your string back, but *also* output the raw error message? That should satisfy both of us.

app/src/main/java/org/torproject/android/ui/connect/ConfigConnectionBottomSheet.kt
tladesignz: Yeah. I know. *sigh*.

This was my workaround for the designer dropping my "I know I cannot connect without a bridge" toggle, which I initially had.

We ran out of time, and it's still unclear to me, why the designer dropped that feature. I guess, they never understood, what that was for, and of course designers want to reduce UI complexity as much as possible.

My toggle solution wasn't good, that is clear to me. But it is also a crucial feature, I think. Because Moat does have different answers in that case, and that might be helpful to some users in certain situations. (Like libraries, where they block Tor, but don't go the long way to block bridges, too?)

So, that was my solution back then:

In case, the users use "Smart Connect", I would use the default answer and work from there. After all, the algo would also try different bridges anyway, if connection wouldn't happen.

But if the user explicitly asked for bridges, they might have a deeper reason for that, so I rather give them a bridge, then tell them "all is fine" in situations, where not all might be fine. After all, a user explicitly asked for it.

That's, of course, not your situation when you try to test everything and know very well, what and how your network is set up. But that's probably not the situation a typical user has how comes to that scene.

What's your opinion on that reasoning?

tladesignz:
@bitmold , your debugging around `SmartConnect` was super helpful! Thank you very much!

That `purge` is a great find! I wouldn't have expected that.

I added your suggestions, and made sure, the `SmartConnect` `TimerTask` also gets cancelled, when the user cancels connecting.

> Seems like pressing cancel should not only interrupt the timer task properly, but also ensure that any transports which might have been started throughout the dance of SmartConnect, stop.

This should already happen in `OrbotService.stopTorAsync()`:
https://github.com/tladesignz/orbot/blob/bridge_refurbish/orbotservice/src/main/java/org/torproject/android/service/OrbotService.java#L205-L227

Which gets called in the `ACTION_STOP` handler:
https://github.com/tladesignz/orbot/blob/bridge_refurbish/orbotservice/src/main/java/org/torproject/android/service/OrbotService.java#L808-L813

bitmold:
My point was with the supposedly localized message, "Read timed out" is always in english, and not exactly clear that it even has anything to do with "asking tor" to a lay user. 

I hate vague error messages too, but using a string that conveys the thing you are trying to do couldnt work is slightly better, especially if it is localized into non english languages 

But ya glad its changed, thanks. I would be slightly more tolerant if the "localized" messages actually were in more than 1 language. Localized here means localize if possible, its a misleading method name

bitmold:
image

I've never used Telegram for obfs4 or anything else so this might be a dumb question. Just wondering, does the bot require writing the word 'start' in English? I'd assume no. Only asking this here because translators may translate the start in instructions into other languages.

bitmold:
I'm just being thorough here, I'm not even necessarily sure this was introduced as part of your PR. but how would a user scan "the QR code" to add bridges?

image

I don't think any QR reading functionality has existed in the app for years. Maybe I'm ignorant about this, though (@n8fr8 do you know?)

bitmold:
I can't reproduce any of those bugs anymore, and other stress testing stuff seems fine. 

Aside from the ongoing issue of if "Not Sure? Ask Tor" should force a bridge everything else here looks great

tladesignz:
> image
> I've never used Telegram for obfs4 or anything else so this might be a dumb question. Just wondering, does the bot require writing the word 'start' in English? I'd assume no. Only asking this here because translators may translate the start in instructions into other languages.

Translation string placeholders are our friend:
https://github.com/tladesignz/orbot/blob/bridge_refurbish/app/src/main/res/values/strings.xml#L238

tladesignz:
> I'm just being thorough here, I'm not even necessarily sure this was introduced as part of your PR. but how would a user scan "the QR code" to add bridges?
> 
> image
> I don't think any QR reading functionality has existed in the app for years. Maybe I'm ignorant about this, though (@n8fr8 do you know?)

Oh, my. You got me there. Forgot to add that part. :-/ More work!

tladesignz:
Sooooo. Finally! I think, we have it together now. @bitmold, can you give the QR code scanning on last test drive?

It brought back the Zxing `IntentIntegrator`! :-) Turns out, besides implementing your own QR code scanning (of course, preferably with Google's MLKit, of course integrated in Google Play...), the Zxing thing became a quasi-standard, and a lot of apps seem to react to that intent request.

Of course, that's brittle, but that's better then the alternatives. @n8fr8 thinks so, too.

bitmold:
It's good, definitely ready for a beta test

PR #1363, opened 2025-07-02
Update dependencies and remove unused ones

meenbeese:


bitmold:
Thanks, closing because a lot of other dependency updates (including these) have been put into #1371 which I've tested

Issue #1369, opened 2025-07-26
[BUG] Invalid bridge format for Tor-relay bridges


ZdenekBachy:
**Describe the Bug**
For bridges like this: 217.83.162.43:9001 6F1D1B83152D6181CAC2C726EB7547B7B9BB7B7F a window "Invalid bridge format" appears erroneously, which prevents them from working.

**To Reproduce**
1. Open Orbot (in my case Samsung nonroot)
2. Choose Custom bridges
3. Place bridge
4. See window "Invalid bridge format" 

**Expected Behavior**
No window "Invalid bridge format". This format is correct. It work in Tor browser, for example. 
Additionally, I think such UI is incorrect in general, especially when it limits functionality.

**Screenshots**

![Image](https://github.com/user-attachments/assets/abc63d31-2902-4b7f-9747-ba6712455266)

**Smartphone**
- Device: Samsung M52
- OS: Android 12
- Version: 17.5.0-RC-1-tor-0.4.8.16

syphyr:
The format is invalid because it's not specifying the bridge type at the beginning of the line.

ZdenekBachy:
Okey, but why such a type successfully worked on the previous version of Orbot? What should I write before bridge to make it work?

syphyr:
What type of bridge is it?

ZdenekBachy:
I'm a regular user, not a network engineer. Vanilla, maybe? I took it from [Tor-relay-scanner](https://github.com/ValdikSS/tor-relay-scanner), and it worked. Today Tor-relay-scanner is the only way in my country.
I understand that there are many types of obfuscation: meek, webtunnel, snowflake, obfs4, but which one is this?

syphyr:
Try putting obfs4 at the beginning.

ZdenekBachy:
I get your point. 
Obfs4 gives me:
<details><summary>Log</summary>
<p>

ClientTransportPlugin obfs4 socks5 127.0.0.1:35313
Bridge obfs4 79.43.84.251:9001 36FD1196BDCF2ABF5A9533DCCADD2FD2433544B1
GeoIPFile /data/data/org.torproject.android/files/geoip
GeoIPv6File /data/data/org.torproject.android/files/geoip6
StrictNodes 0



Orbot is Starting…
binding tor service

</p>
</details> 

Webtunnel gives simillar log.

Snowflake, vanilla, meek, meek-azure and some other lead to "Invalid bridge format".

[...]

PR #1374, opened 2025-07-30
Add validation support for MeekLite and Vanilla custom bridges

syphyr:
Add regex support for MeekLite and Vanilla custom bridge validation. Tor supports a "vanilla bridge" with `Bridge IP:PORT <FINGERPRINT>` which can be helpful in specific regions.  Also, only enable the "Connect" button for custom bridges after a successful format validation.

Fixes: #1369

### DIFF
--- a/app/src/main/java/org/torproject/android/ui/connect/CustomBridgeBottomSheet.kt
+++ b/app/src/main/java/org/torproject/android/ui/connect/CustomBridgeBottomSheet.kt
@@ -1,6 +1,8 @@
 package org.torproject.android.ui.connect
 
 import android.app.AlertDialog
+import android.content.res.ColorStateList
+import android.graphics.Color
 import android.os.Bundle
 import android.text.Editable
 import android.text.TextWatcher
@@ -116,7 +118,16 @@ class CustomBridgeBottomSheet(private val callbacks: ConnectionHelperCallbacks) 
 
     private fun updateUi() {
         val inputText = binding.etBridges.text.toString()
-        binding.btnAction.isEnabled = inputText.isNotEmpty() && isValidBridge(inputText)
+        val isValid = inputText.isNotEmpty() && isValidBridge(inputText)
+
+        binding.btnAction.isEnabled = isValid
+        binding.btnAction.backgroundTintList = ColorStateList.valueOf(
+            if (isValid) {
+                requireContext().getColor(R.color.orbot_btn_enabled_purple)
+            } else {
+                Color.DKGRAY
+            }
+        )
 
         if (!isValidBridge(inputText)) {
             binding.etBridges.error = requireContext().getString(R.string.invalid_bridge_format)
--- a/app/src/main/java/org/torproject/android/ui/connect/CustomBridgeBottomSheet.kt
+++ b/app/src/main/java/org/torproject/android/ui/connect/CustomBridgeBottomSheet.kt
@@ -23,9 +23,10 @@ class CustomBridgeBottomSheet(private val callbacks: ConnectionHelperCallbacks) 
 
     companion object {
         const val TAG = "CustomBridgeBottomSheet"
-        private val bridgeStatement = Regex("(obfs4|meek|webtunnel)")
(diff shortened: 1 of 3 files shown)


syphyr:
The regex for `bridgeStatement` was changed to look for an ip address in the custom bridge lines  because vanilla bridges do not use a specific bridge type and the one thing in common for all custom bridges is that they define an ip address.

tladesignz (changes requested):
Sorry, we cannot just let any bridge statement through. It doesn't work this way.
To use a bridge of a specific type, Orbot needs to make sure, that the resp. bridge type code is also started and listens for Tor to connect to.

That happens in `Transport.start(Context)`. For custom bridges, the list of bridge types is extracted from the custom bridge strings and started respectively. If you would be able to add types, which we don't understand (currently Snowflake, Meek, Obfs4 and Webtunnel), the code would throw.

tladesignz:
Uuuh. I just read through the #1369 issue. I'm working on this stuff now for 8 years, and I actually never heard of "vanilla" bridges.
Ok, if we want to support this, we will probably need to extend the `Transport` enum, so it actually supports that.
But let me clarify this first with some more knowledgable people.

syphyr:
> Sorry, we cannot just let any bridge statement through. It doesn't work this way. To use a bridge of a specific type, Orbot needs to make sure, that the resp. bridge type code is also started and listens for Tor to connect to.
> 
> That happens in `Transport.start(Context)`. For custom bridges, the list of bridge types is extracted from the custom bridge strings and started respectively. If you would be able to add types, which we don't understand (currently Snowflake, Meek, Obfs4 and Webtunnel), the code would throw.

From my understanding, vanilla bridges do not require any pluggable transports. So, Orbot only needs to add the `UseBridges 1` to the config without starting any PT's.

syphyr:
Also, we are not just letting any custom bridge statement through because of the additional regex check.

bitmold:
Assuming the regex validates correctly this looks good to me

syphyr:
> Assuming the regex validates correctly this looks good to me

I used this vanilla bridge to test regex.

```
217.83.162.43:9001 6F1D1B83152D6181CAC2C726EB7547B7B9BB7B7F
```

The fingerprint is always 40 characters. Also, both ipv4 and ipv6 are supported by the regex check.

tladesignz:
@bitmold, the `Transport` code didn't break apart when you tested it?

syphyr:
> @bitmold, the `Transport` code didn't break apart when you tested it?

You are right, it fails because it thinks that `217.83.162.43:9001` is the transport type. So, how could I fix that?

I need to change it so that the transport type is blank if it is not one of the known transports.

syphyr:
@tladesignz  Maybe this could be fixed by changing the filter on custom transports?

https://github.com/guardianproject/orbot-android/blob/3f24f6e62f1eace6946ccb97f42dd418a50c7c8a/orbotservice/src/main/java/org/torproject/android/service/circumvention/Transport.kt#L102

syphyr:
I fixed the problem with the invalid custom transport name by adding an additional filter. This solved the crash.

syphyr:
logcat now shows that `UseBridges 1` is set correctly with custom vanilla bridges and Orbot does not crash trying to set an unknown transport.  Also, obfs4 and webtunnel still work as usual. I do not have an actual working vanilla bridge to connect to, but the logs look correct.

bitmold:
I'll test this later this week and merge if correct, thank you

tladesignz:
@syphyr, thanks for pushing that whole vanilla bridge thing. Shame on me, that I didn't know about it. I'll bring support to Apple platforms, too!
I improved the parsing of bridge lines now, so everything should work as expected.

I tested, and nothing seems to break and the torrc log looks fine.

@bitmold, can you please confirm, and if good from your perspective, merge?
Thank you!

bitmold:
Tested it and it's working with vanilla bridges obtained from [tor-relay-scanner](https://github.com/ValdikSS/tor-relay-scanner) 

The vanilla bridges play nicely with obfs4 custom bridges too. Merging.

bitmold:
The commits have been merged, I rebased this branch on `master` so merged it by hand. Closing this PR...

PR #1379, opened 2025-08-01
Fix landscape orientation for tablets

ezhd:
SCREEN_ORIENTATION_LANDSCAPE always turn screen clockwise. 
However some tablet users might have their screens locked in counterclockwise orientation. 
This situation leads to Orbot activities being locked upside-down. 

SCREEN_ORIENTATION_USER_LANDSCAPE respects the orientation picked by the user. 
Rotation does not trigger restart of the activity.


bitmold:
Great thank you

PR #1391, opened 2025-08-13
Implement a new floating navbar

meenbeese:
The current navbar feels out of place ever since edgeToEdge has been enabled by default on Android. Moreover, it is stretched out and is poorly optimized for larger devices and on landscape. I thus decided to wrap it inside a cardview to turn it into a floating component which has the benefit of feeling more material-like and more modern. The logic remains unchanged. Open to tweaking the margins/dimensions/colors better as I chose some initial defaults that made sense.

<table>
    <tr>
        <th>Before</th>
        <th>After</th>
    </tr>
    <tr>
        <td>
</td>
        <td>
</td>
    </tr>
</table>
<table>
    <tr>
        <th>Before</th>
        <th>After</th>
    </tr>
    <tr>
        <td></td>
        <td></td>
    </tr>
</table>

### DIFF
--- a/app/src/main/res/layout/activity_orbot.xml
+++ b/app/src/main/res/layout/activity_orbot.xml
@@ -13,23 +13,18 @@
         android:layout_width="match_parent"
         android:layout_height="match_parent"
         app:defaultNavHost="true"
-        app:layout_constraintBottom_toTopOf="@id/bottom_navigation"
+        app:layout_constraintBottom_toTopOf="@id/floating_bottom_nav"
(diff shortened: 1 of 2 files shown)


bitmold:
This is a good idea, do you know how it looks on API 24 (the minimum)

meenbeese:
Tweaked the width and height a bit, I will test on API 24.

syphyr:
I just testing on my tablet running Nougat and it looks great. Good job!

bitmold:
This came out so nice, great work, thank you!

PR #1395, opened 2025-08-14
Update to latest Gradle and AGP

meenbeese:


bitmold:
very nice, thank you. getting closer to kotlin dsl

PR #1396, opened 2025-08-14
Remove unused strings with no default value

syphyr:
warn: removing resource org.torproject.android.nightly:string/bridge_snowflake without required default value.
warn: removing resource org.torproject.android.nightly:string/btn_use_custom_bridge without required default value.
warn: removing resource org.torproject.android.nightly:string/configure_header without required default value.
warn: removing resource org.torproject.android.nightly:string/in_a_browser without required default value.
warn: removing resource org.torproject.android.nightly:string/menu_help_others without required default value.


bitmold:
Thanks, this must have escaped me when I was refactoring

Issue #1400, opened 2025-08-19
[BUG] Orbot not respecting Outbound Network Proxy I gave


SeeleVolleri:
I tried set an outbound proxy that I built, but it didn't try to use it.

**To Reproduce**
1. Go to 'More'
2. Click on 'Orbot Settings'
3. Scroll down to 'Outbound Network Proxy'
4. Outbound Proxy Type input 'SOCKS5'
5. Outbound Proxy Host input a random hostname or IP
6. Outbound Proxy Port input a random port
7. fire up wireshark and see nothing

**Expected Behavior**
It should respect my opinion during direct tor connect, and proxy setting should being more friendly

**Smartphone (please complete the following information):**
- Device: [e.g. random android]
- OS: Android 11
- Version: v17.6.0-BETA-1-tor-0.4.8.17

syphyr:
Can you include a logcat please.

syphyr:
Entering a "random hostname or IP" will not work without also configuring the network settings of your android device to match that same hostname or IP.
Also, the port that is used must not be within the system port range (1-1023) since Orbot does not have root permissions to run Tor on system ports.

SeeleVolleri:
I'm sorry, but I assume this feature was relevant as tor's `Socks5Proxy example.com:1234` feature. If not, what's the use case?

syphyr:
The most common use case is probably sharing the Tor service on your local network with a private IP address. Although, it could be any address as long as your device is setup to have that address.

SeeleVolleri:
> The most common use case is probably sharing the Tor service on your local network with a private IP address. Although, it could be any address as long as your device is setup to have that address.

So it should be the `Socks5Proxy` feature?

I tried to set up a dante socks5 server on my local network listening on 7890, and can confirm it is reachable to Internet and the device running Orbot,
But I didn't see anything like `conn_proxy` which like tor did:

```
$ echo 'Socks5Proxy proxy.lan:7891' >torrc 
$ tor -f torrc 
Aug 20 19:53:17.552 [notice] Tor 0.4.8.16 running on Linux with Libevent 2.1.12-stable, OpenSSL 3.5.1, Zlib 1.3.1, Liblzma 5.4.5, Libzstd 1.5.6 and Glibc 2.41 as libc.
Aug 20 19:53:17.552 [notice] Tor can't help you if you use it wrong! Learn how to be safe at https://support.torproject.org/faq/staying-anonymous/
[...]

PR #1401, opened 2025-08-20
Remove redundant VPN check for outbound proxy config

SeeleVolleri:
Tor's outbound proxy does not conflict with VPN mode, so remove this check.

### DIFF
--- a/orbotservice/src/main/java/org/torproject/android/service/tor/TorConfig.kt
+++ b/orbotservice/src/main/java/org/torproject/android/service/tor/TorConfig.kt
@@ -74,26 +74,23 @@ object TorConfig {
         if (transport == Transport.NONE) {
             conf.add("UseBridges 0")
 
-            if (!Prefs.useVpn()) {
-                // Set the proxy here if we aren't using a bridge.
-                val proxyType = prefs?.getString("pref_proxy_type", null)
+            val proxyType = prefs?.getString("pref_proxy_type", null)
 
-                if (!proxyType.isNullOrEmpty()) {
-                    val proxyHost = prefs.getString("pref_proxy_host", null)
-                    val proxyPort = prefs.getString("pref_proxy_port", null)
-                    val proxyUser = prefs.getString("pref_proxy_username", null)
-                    val proxyPass = prefs.getString("pref_proxy_password", null)
+            if (!proxyType.isNullOrEmpty()) {
+                val proxyHost = prefs.getString("pref_proxy_host", null)
+                val proxyPort = prefs.getString("pref_proxy_port", null)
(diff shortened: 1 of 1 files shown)


SeeleVolleri:
#1400

SeeleVolleri:
[Snowflake and obfs4 was unreachable on 2025/08/20 in China Mainland](https://gfw.report/blog/gfw_unconditional_rst_20250820/en/)
So I tried connect tor using a custom SOCKS5 proxy. It didn't work as expected. All broker were blocked.

syphyr:
@tladesignz @bitmold  Is there an actual conflict with enabling VPN mode and enabling outbound proxy connections at the same time? Or, is the solution just to enable Power User mode?

SeeleVolleri:
From my tests, the domain resolved successfully and connected via IPv6 without issues.

syphyr:
> It shouldn't check if we're using VPN since there already checked if we're not using bridge.

The wording of the description here can be confusing for someone that is reviewing this MR. Checking if a bridge is being used should not have anything to do with if VPN mode is being used.  The actual description should say that Tor's  Outbound Proxy does not actually conflict with VPN mode, so this check should be removed. Other than the description being confusing, this change seems ok to me.

syphyr:
I changed the description to this:

```
    Remove VPN check for outbound proxy config
    
    Tor's outbound proxy does not conflict with VPN mode, so remove this
    check.
```
https://github.com/syphyr/orbot/commit/44a8fc10f100176765bfce22c392de2d0b324fa1

SeeleVolleri:
> I changed the description to this:
> 
> ```
>     Remove VPN check for outbound proxy config
>     
>     Tor's outbound proxy does not conflict with VPN mode, so remove this
>     check.
> ```
> 
> [syphyr@44a8fc1](https://github.com/syphyr/orbot/commit/44a8fc10f100176765bfce22c392de2d0b324fa1)

Appreciate!

SeeleVolleri:
I noticed it was a mirror outside Github. Does it means the commit will go upstream, or I should leave this PR open?

syphyr:
Leave open so it can be reviewed.

syphyr:
@SeeleVolleri I suggest to change the subject and description of this MR to:

```
Allow outbound proxy to work in VPN mode

Tor's outbound proxy does not conflict with VPN mode, so remove this restriction
```

SeeleVolleri:
> @SeeleVolleri I suggest to change the subject and description of this MR to:
> 
> ```
> Allow outbound proxy to work in VPN mode
> 
> Tor's outbound proxy does not conflict with VPN mode, so remove this restriction
> ```

Yes, I agree it is a great description. Please give me some time on how to do it :)

PR #1408, opened 2025-08-27
Fix string with wrong default dns port

syphyr:
The default DNS port is now 9053.


bitmold:
awesome, thanks, going to check for other strings

PR #1416, opened 2025-09-05
Added support for UPnP IGD to automatically open up ports for Snowflake Proxy…

tladesignz:
… if available.

Still needs an update to IPtProxy and the outlined change, when available.


app/src/main/AndroidManifest.xml
```diff
@@ -41,6 +41,7 @@
         android:configChanges="locale|orientation|screenSize"
         android:description="@string/app_description"
         android:hasFragileUserData="false"
+        android:usesCleartextTraffic="true"
```
bitmold, line 44: can you please explain why this is necessary, my gut reaction is to be adverse to this

tladesignz: 😆 I would have had the same reaction. Guess, this needs a comment.
It is needed, because the UPnP IGD protocol is unencrypted HTTP.

Would be happy, if we could limit that to local networks, but I doubt this is possible.

bitmold: Actually, I think it is. I can find some documentation when I get on my computer later, but I've seen this sort of thing before.

orbotservice/src/main/java/org/torproject/android/service/circumvention/SnowflakeProxyWrapper.kt
```diff
@@ -22,37 +32,59 @@ class SnowflakeProxyWrapper(private val context: Context) {
...
+                brokerUrl = OrbotService.getCdnFront("snowflake-target-direct")
+                capacity = 1L
+                pollInterval = 120L
+                stunServer = stunUrl
+                relayUrl = OrbotService.getCdnFront("snowflake-relay-url")
+                natProbeUrl = OrbotService.getCdnFront("snowflake-nat-probe")
+                clientConnected = SnowflakeClientConnected { onConnected() }
+
+// TODO: Activate, when new IPtProxy is available.
+//            ephemeralMinPort = mappedPorts.firstOrNull()
```
bitmold, line 64: will IPtProxy be ok with getting null as an argument, just because if mappedPorts is < 3 the ports are released and then mappedPorts is reassigned to an empty list, meaning that `ephemeralMinPort` and `ephemeralMaxPort would in fact be assigned to null here

orbotservice/src/main/java/org/torproject/android/service/circumvention/SnowflakeProxyWrapper.kt
```diff
@@ -22,37 +32,59 @@ class SnowflakeProxyWrapper(private val context: Context) {
...
+        CoroutineScope(Dispatchers.IO).launch {
+            val start = Random.nextInt(49152, 65536 - 2)
+
+            for (port in (start..start + 2)) {
+                if (UPnP.openPortUDP(port, OrbotConstants.TAG)) {
+                    mappedPorts.add(port)
+                }
+            }
+
+            if (mappedPorts.size < 3) {
```
bitmold, line 45: i might just be missing some context, but if you are only using two ports out of `mappedPorts`, the ephemeral min and max ports, than why does the size need to be less than 3. at a glance, this looks like it should be less than 2, since you're only needing 2 ports here.

app/src/main/AndroidManifest.xml
tladesignz: It's not about talking to localhost (same device), it's about talking to some router on the local network, which could potentially be on *any* IP address, but typically is on a private IP address only. https://en.wikipedia.org/wiki/Private_network

orbotservice/src/main/java/org/torproject/android/service/circumvention/SnowflakeProxyWrapper.kt
tladesignz: Yes, that's why this is done. Setting max/min ports to null is like setting it not at all.
Needs another comment, I guess.

orbotservice/src/main/java/org/torproject/android/service/circumvention/SnowflakeProxyWrapper.kt
tladesignz: We need 3 consecutive ports, because that's what Snowflake Proxy wants to function properly when having Capacity set to 1.
In fact, it's Capacity * 2 + 1.
Seems, another comment is necessary. Will add a commit now with all the comments.

orbotservice/src/main/java/org/torproject/android/service/circumvention/SnowflakeProxyWrapper.kt
```diff
@@ -22,37 +32,63 @@ class SnowflakeProxyWrapper(private val context: Context) {
...
+                capacity = 1L
+                pollInterval = 120L
+                stunServer = stunUrl
+                relayUrl = OrbotService.getCdnFront("snowflake-relay-url")
+                natProbeUrl = OrbotService.getCdnFront("snowflake-nat-probe")
+                clientConnected = SnowflakeClientConnected { onConnected() }
+
+// TODO: Activate, when new IPtProxy is available.
+                // Setting these to null or 0 is equivalent to not setting this at all.
+//                ephemeralMinPort = mappedPorts.firstOrNull()
```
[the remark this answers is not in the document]
syphyr: I'm not sure if this is the right way, but the problem is fixed by doing this:

-                ephemeralMinPort = mappedPorts.firstOrNull()
-                ephemeralMaxPort = mappedPorts.lastOrNull()
+                val ephemeralMinPort = mappedPorts.firstOrNull()
+                val ephemeralMaxPort = mappedPorts.lastOrNull()

PR #1426, opened 2025-09-10
Refactor Tor/VPN actions to use ViewModel events

meenbeese:
This PR replaces direct method calls for starting/stopping Tor and VPN, refreshing menu, and sending Newnym signal with event-based triggers in `ConnectViewModel`. This improves separation of concerns and centralizes UI event handling, making the codebase more maintainable and testable.

Tested on Pixel 8 API 35.


syphyr:
I've been testing this for a few weeks without any issues.

syphyr:
I just noticed that when these changes are applied, then I have to select the stop button twice before tor will disconnect.

meenbeese:
> I just noticed that when these changes are applied, then I have to select the stop button twice before tor will disconnect.

Tapping on the stop button does take a bit of time, but I can get it to work without tapping twice.

bitmold:
Not merging for reason I had given on the other PR

PR #1428, opened 2025-09-11
Removes ACTION_START_VPN and ACTION_STOP_VPN intent actions from OrbotService

bitmold:
`ACTION_START_VPN` and `ACTION_STOP_VPN` were added when Orbot became a VPN app. They've stuck around in the app logic ever since - causing duplicate blocks of code in `OrbotService` and are no longer needed. 

`ACTION_START` and `ACTION_STOP` do everything `ACTION_START_VPN` and `ACTION_STOP_VPN` do if and only if there is actually a VPN session, making these commands entirely superfluous. 

```java

                case ACTION_START -> {
                    var transport = Prefs.getTransport();
                    transport.start(OrbotService.this);

                    startTor();
                    replyWithStatus(mIntent);
                    if (Prefs.useVpn()) {
                        if (mVpnManager != null && (!mVpnManager.isStarted())) { // start VPN here
                            Intent vpnIntent = VpnService.prepare(OrbotService.this);
                            if (vpnIntent == null) { //then we can run the VPN
                                mVpnManager.handleIntent(new Builder(), mIntent);
                            }
                        }
                        if (mPortSOCKS != -1 && mPortHTTP != -1)
                            sendCallbackPorts(mPortSOCKS, mPortHTTP, mPortDns, mPortTrans);
                    }
                }

                //// handling of other cases ...

                case ACTION_START_VPN -> {
                    if (mVpnManager != null && (!mVpnManager.isStarted())) {
                        //start VPN here
                        var vpnIntent = VpnService.prepare(OrbotService.this);
                        if (vpnIntent == null) { //then we can run the VPN
                            mVpnManager.handleIntent(new Builder(), mIntent);
                        }
                    }
                    if (mPortSOCKS != -1 && mPortHTTP != -1)
                        sendCallbackPorts(mPortSOCKS, mPortHTTP, mPortDns, mPortTrans);
                }
```

PR #1429, opened 2025-09-11
Snowflake Proxy runs as its own Foreground Service 

bitmold:
- Based code off of https://github.com/guardianproject/orbot-android/pull/1428 since that also removes a lot of cruft from `OrbotService`, that'll need to be merged first 
- Removes all starting + stopping of Kindness Mode from `OrbotService`
- Power + WiFi Only Settings work 
- When the UI says you're in kindness mode, you always actually are in kindness mode. 
- Been serving a lot more snowflakes with this running than previously


bitmold:
Initially I had taken that approach with the `systemExempted` foreground service permission that OrbotService uses (the one that `VPNService`s use) but then declared the new snowflake proxy service as a different kind of foreground service, remote messaging: 
https://developer.android.com/about/versions/14/changes/fgs-types-required#remote-messaging

If Google Play is fine with this, we won't need that alarms permission for snowflake

tladesignz (approved):

tladesignz:
Please merge yourself when you see fit!

PR #1430, opened 2025-09-13
Use non-deprecated window API on API 30+

meenbeese:


bitmold:
switching from `activity` to `requireActivity()` would cause a crash if you aren't attached since requireActivity would throw a run time exception. 

that's why the method type was `Activity?` and it's wrapped in an `activity?.let` to only set the height if attached 

not merging since this reintroduces this bug/crash scenario

PR #1443, opened 2025-09-28
Allow custom snowflake

ixiumu:
In regions where Tor is blocked/censored, people often use self-hosted or reverse proxy Snowflake servers. They sometimes also need to modify the frontend domain, but overly strict regular expressions prevent the app from connecting at all.

### DIFF
--- a/app/src/main/java/org/torproject/android/ui/connect/CustomBridgeBottomSheet.kt
+++ b/app/src/main/java/org/torproject/android/ui/connect/CustomBridgeBottomSheet.kt
@@ -23,28 +23,22 @@ class CustomBridgeBottomSheet() :
 
     companion object {
         const val TAG = "CustomBridgeBottomSheet"
-        private val bridgeStatement =
-            Regex("""(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}|\[[0-9a-fA-F:]+])""")
-        private val meekLiteRegex =
-            Regex("""^meek_lite\s+(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}|\[[0-9a-fA-F:]+]):\d+\s+url=https?://\S+\s+front=\S+\s+utls=\S+$""")
-        private val obfs4Regex =
-            Regex("""^obfs4\s+(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}|\[[0-9a-fA-F:]+]):\d+\s+[A-F0-9]{40}(\s+cert=[a-zA-Z0-9+/=]+)?(\s+iat-mode=\d+)?$""")
-        private val vanillaRegex =
-            Regex("""(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}|\[[0-9a-fA-F:]+]):\d+\s+[A-F0-9]{40}?$""")
-        private val webtunnelRegex =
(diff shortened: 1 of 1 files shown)


bitmold:
@ixiumu @tladesignz @n8fr8 

is there a good resource to get a self hosted snowflake server to test this with? admittedly, I haven't used snowflake much outside of the default settings used to get snowflakes from the broker. 

at a glance, my gut instinct is (perhaps wrongly) telling me that this regex went from being too restrictive to  **too permissive**. So yeah, we just wanna make sure this regex change doesn't also allow for invalid/formatted data. 

but in any case, we're excited to get this into our next release - im always grateful to receive insights into what users in different censored regions are doing to get around their censors and we're enthusiastic to ship fixes like this that make things easier for use-cases that we haven't fully considered.

tladesignz:
So, valid snowflake bridge lines look like this:

```
snowflake 192.0.2.3:80 A552D70428813177B25083D0196085536A6EBF04 fingerprint=A552D70428813177B25083D0196085536A6EBF04 url=https://example.org/ fronts=www.example.com,www.example.org ice=stun:stun.example.com:3478,stun:stun.example.org:3478 utls-imitate=hellorandomizedalpn
```

I would suggest to at least add a check of the IP and the fingerprint to the regex.

### Notes

- The IP address is a dummy adress, always IPv4 and always in the 192.0.2.0/24 range. This is by convention for bridge types which don't actually use that IP address. (Like Snowflake, as the Snowflake proxies are taken dynamically from a pool.)

- The fingerprint is doubled, as one is for Tor, the other one for the Snowflake client binary. If you would be hardcore, your regex would make sure that both are identical.

- Arguments like `utls-imitate` are optional. There are other optional arguments.

ixiumu:
```
snowflake 192.0.2.4:80 8838024498816A039FCBBAB14E6F40A0843051FA fingerprint=8838024498816A039FCBBAB14E6F40A0843051FA url=https://1098762253.rsc.cdn77.org/ fronts=www.example.com ice=stun:stun.example.com:3478,stun:stun.example.org:3478 utls-imitate=hellorandomizedalpn
snowflake 192.0.2.3:80 A552D70428813177B25083D0196085536A6EBF04 fingerprint=A552D70428813177B25083D0196085536A6EBF04 url=https://www.example.com/ ice=stun:stun.example.com:3478,stun:stun.example.org:3478 utls-imitate=hellorandomizedalpn
snowflake 192.0.2.3:80 A552D70428813177B25083D0196085536A6EBF04 fingerprint=A552D70428813177B25083D0196085536A6EBF04 utls-imitate=hellorandomizedalpn url=https://www.example.com/ ice=stun:stun.example.com:3478,stun:stun.example.org:3478
```

I've come across the three bridge line formats above. The second and third differ only by the parameter sequence and are both missing the fronts field. It might be easier to simplify the regex by mandating a fixed parameter order.

tladesignz:
Parameters to the Snowflake client shouldn't be checked. Esp. Snowflake has *lots* of them and many or optional. Just the basic Tor Bridge line parameters at the beginning.

tladesignz:
Ok. Thanks for pushing on this matter!

I looked hard at the current solution, and I seriously wasn't satisfied with it.
Some of the stuff was too specific (bridge arguments are subject to change and order can vary), and there was too much duplication. Additionally, they were underdocumented.

I gathered a bunch of bridge lines of all sorts and the resulting regex should be stable for the foreseeable future. (If there's not a wild special character out there which didn't make the list.)

@ixiumu, can you confirm this works with what you're up to?

ixiumu:
@tladesignz I compiled and test-ran it using your pull, and it works correctly. Thank you!

Issue #1440, opened 2025-09-21
[BUG] tor_raw_abort_ when switching between Bridges


kasia-de:
**Describe the Bug**

**To Reproduce**
Quickly switch back and forth between direct and snowflake, disconnecting in between, until Orbot crashes.

1. Open Orbot (cold boot).
2. Click "how to connect"
3. Click direct
4. Click connect
5. Click cancel
6. Click "how to connect"
7. Click snowflake
8. Click connect
9. Click disconnect
10. goto step 2

**Expected Behavior**
No crash

**What Custom Configuration Do You Use?**

**Screenshots**

**Smartphone (please complete the following information):**
- Device: Google Pixel Tablet
- OS: GrapheneOS 2025091000
- Version: 17.5.0-RC-2-tor-0.4.8.17

**Crash Logs (Advanced)**
```
09-13 20:19:59.826 15741 19927 F libc    : Fatal signal 6 (SIGABRT), code -1 (SI_QUEUE) in tid 19927 (tor), pid 15741 (project.android)
09-13 20:20:00.394 19931 19931 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
09-13 20:20:00.394 19931 19931 F DEBUG   : Build fingerprint: 'google/tangorpro/tangorpro:16/BP2A.250805.005/2025091000:user/release-keys'
09-13 20:20:00.394 19931 19931 F DEBUG   : Revision: 'MP1.0'
09-13 20:20:00.394 19931 19931 F DEBUG   : ABI: 'arm64'
09-13 20:20:00.394 19931 19931 F DEBUG   : Timestamp: 2025-09-13 20:20:00.008272133+0000
09-13 20:20:00.394 19931 19931 F DEBUG   : Process uptime: 1300s
09-13 20:20:00.394 19931 19931 F DEBUG   : Cmdline: org.torproject.android
09-13 20:20:00.394 19931 19931 F DEBUG   : pid: 15741, tid: 19927, name: tor  >>> org.torproject.android <<<
09-13 20:20:00.394 19931 19931 F DEBUG   : uid: 2510182
09-13 20:20:00.394 19931 19931 F DEBUG   : tagged_addr_ctrl: 0000000000000001 (PR_TAGGED_ADDR_ENABLE)
09-13 20:20:00.394 19931 19931 F DEBUG   : signal 6 (SIGABRT), code -1 (SI_QUEUE), fault addr --------
[...]

PR #1450, opened 2025-10-09
Full proxy support + reconf instead of restart on smart connect

tladesignz:
No UI improvements, yet.

Should we add these?

- [x] Explanation, that the proxy will not work with Snowflake?
- [ ] Trash option menu button to quickly delete proxy settings again?
- [x] Improve settings to show entered values directly?
- [x] Only allow https, socks4 and socks5 schemes through predefined values selector?

I'll add a refactoring to support at-runtime-reconfiguration, too. Important for smart connect to reduce crashes.

### DIFF
--- a/app/src/main/java/org/torproject/android/service/OrbotService.java
+++ b/app/src/main/java/org/torproject/android/service/OrbotService.java
@@ -353,6 +353,20 @@ public class OrbotService extends VpnService {
             }
 
             return null;
+        }, () -> {
+            if (conn == null) return false;
+
+            try {
+                conn.resetConf(Arrays.asList("UseBridges", "ClientTransportPlugin", "Bridge"));
+                conn.setConf(Prefs.getTransport().getTorConfig(this));
+            }
+            catch (IOException e) {
+                logNotice(e.getLocalizedMessage());
+
+                return false;
+            }
+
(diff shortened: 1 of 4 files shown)


tladesignz:
Oh, well. It's that Preferences thing which is deprecated. I guess the nice trash button needs to wait until we replaced preferences completely.

tladesignz:
Ok, reconfiguring instead of restarting was easier than I thought. At least for smart connect.

Do you know if there's other places where we could leverage reconfiguring instead of restarting, @bitmold?

Any objections with this PR? Any improvement suggestions?

tladesignz:
Will mitigate #1440

bitmold:
> Do you know if there's other places where we could leverage reconfiguring instead of restarting, @bitmold? 

Off the top of my head, no. But can investigate this further. Been afk for a little bit. 

> Oh, well. It's that Preferences thing which is deprecated. I guess the nice trash button needs to wait until we replaced preferences completely.

This has been long overdue. Part of refactoring `orbotservice` into `app` was so that migrating to the newer Preference framework would be easier to do.


Otherwise, for now, this is looking good to me.

hosein203:
مرورگرconet android

PR #1452, opened 2025-10-15
Reduce Orbi animation cycles

syphyr:
Orbi's animation currently takes about 5 seconds to complete. This reduces the duration of Orbi's animation to about 2 seconds.  If Orbi's animation has not completed its cycles before changing views, the animation will resume when switching back to the previous view until it has completed all of its cycles.

### DIFF
--- a/app/src/main/java/org/torproject/android/ui/connect/ConnectFragment.kt
+++ b/app/src/main/java/org/torproject/android/ui/connect/ConnectFragment.kt
@@ -340,13 +340,11 @@ class ConnectFragment : Fragment(),
 
         binding.ivStatus.setImageResource(R.drawable.torstarting)
         val animHover = AnimationUtils.loadAnimation(context, R.anim.hover)
-        animHover.repeatCount = 3
         animHover.repeatMode = Animation.REVERSE
         binding.ivStatus.animation = animHover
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/torproject/android/ui/connect/ConnectFragment.kt
```diff
@@ -340,13 +340,13 @@ class ConnectFragment : Fragment(),
...
         binding.ivStatus.setImageResource(R.drawable.torstarting)
         val animHover = AnimationUtils.loadAnimation(context, R.anim.hover)
-        animHover.repeatCount = 7
+        animHover.repeatCount = 3
         animHover.repeatMode = Animation.REVERSE
         binding.ivStatus.animation = animHover
         animHover.start()
 
         val animShadow = AnimationUtils.loadAnimation(context, R.anim.shadow)
-        animShadow.repeatCount = 7
```
bitmold: wondering if these assignments are necessary or if this can all be done in XML. Same with the `repeatMode` assignment too. I'm in favor of this PR and can merge, but would like to get rid of this kotlin if it's unnecessary

bitmold:
hell yeah, thank you

PR #1461, opened 2025-10-22
Show detected custom bridge type as a chip

meenbeese:
This is more of a UX change that can help beginner users by informing them of the bridge type they entered. Obviously the phrase I used is for debugging purposes. Also, the UI can potentially be improved for this bottom sheet but I don't have any great ideas yet. Open to suggestions.

Screenshot_20251021_222527

Tested on Pixel 8 API 35.


n8fr8:
This is a nice idea. Will review asap. Thanks!

meenbeese:
Rebased and fixed the implementation for #1443.

syphyr:
~Perhaps unrelated to what this PR wants to achieve, but I've also noticed that Orbot's main view (with Orbi) displays the bridge type being used when in VPN mode, but not when in power user mode.~ This is fixed now.

PR #1462, opened 2025-10-22
Refactor MoreFragment logic into MoreViewModel

meenbeese:
Moved status text and menu event handling from MoreFragment to a new MoreViewModel using StateFlow and event channels. This improves separation of concerns and prepares the UI for better lifecycle management and testability.

Tested on Pixel 8 API 35.


bitmold:
Currently not interested in these minor rewrites and refactoring. Its a lot to thoroughly test everything and ensure there are no regressions.  If theres something that seems especially worth refactoring open a ticket and we can talk about it.

Issue #1467, opened 2025-10-28
Logo Assets


luanafaustini:
Here are the assets for logo update at app stores and possible customized versions. Thanks!

![Image](https://github.com/user-attachments/assets/db55ce69-2bdb-4d4a-8f70-9b5029c7f467)
![Image](https://github.com/user-attachments/assets/c80e0d7c-54ce-4484-90cf-81a0c6a87450)
![Image](https://github.com/user-attachments/assets/1c67d171-ce3d-4983-b627-23dc722d960b)
![Image](https://github.com/user-attachments/assets/f19e5309-d635-4b8f-9976-47d947e551c5)
![Image](https://github.com/user-attachments/assets/82f9b3b1-83b9-4db0-8812-f07fa3ba4932)
![Image](https://github.com/user-attachments/assets/70fee949-7522-47d4-8253-0378b0772a3a)
![Image](https://github.com/user-attachments/assets/416706c4-70e5-439e-9be4-b94b3b2aed97)
![Image](https://github.com/user-attachments/assets/911d5fad-f1a8-4827-b7eb-49fdf634698e)
<img width="1024" height="1024" alt="Image" src="https://github.com/user-attachments/assets/1e54626a-aaa4-4225-bf0a-c5712b11e7bc" />
![Image](https://github.com/user-attachments/assets/5cbc1b96-bd1a-4da5-a4ac-06a95f2f7d92)
<img width="512" height="512" alt="Image" src="https://github.com/user-attachments/assets/4ca9c4b2-83de-4600-aee6-0b22060fdc7d" />
![Image](https://github.com/user-attachments/assets/657df1e7-8280-4d2c-9cd0-18461dbc861a)

n8fr8:
Thanks. Can you remind me which is the official one for the main icon and where the SVG is?

n8fr8:
SVG files here

[App Logos Customized-20251104T141145Z-1-001.zip](https://github.com/user-attachments/files/23337646/App.Logos.Customized-20251104T141145Z-1-001.zip)

n8fr8:
the "customized" are for the alt icon camouflage feature

here are the main app icons

[App Store Logos-20251104T141318Z-1-001.zip](https://github.com/user-attachments/files/23337664/App.Store.Logos-20251104T141318Z-1-001.zip)

bitmold:
Thanks for these, they've been integrated into Orbot so closing this ticket

PR #1472, opened 2025-11-04
icon and design updates to match new brand guidlines and iOS

n8fr8:
for issue #1467 and others


bitmold:
Looks great, there are a few screens where maybe we'd also want to apply the new gradient background (app selection, settings, onion sevices). Not sure if this was intentional, but if it wasn't I can make the updates to make the theme uniform across all activities/screens. 

And I agree with TLA, it looks really crisp on my irl phones too

bitmold:
@n8fr8 I can rerun the script to regenerate (app store) screenshots in various languages...

PR #1476, opened 2025-11-08
Rewrite and unify navigation with custom wrapper

meenbeese:
This PR removes the legacy Android Navigation Component and `nav_graph.xml`, replacing it with a unified `FragmentNavigator` utility for handling both fragment and activity navigation. These changes not only reduce and simplify boilerplate code, but also unlock future developments such as rewriting activities as fragments and switching to Compose.

Tested on Pixel 8 API 35.


bitmold:
For now, not merging this. We still use actvities for a large part of the app, and are switching more towards nav controller #1483 as an interim between activities and something more modern.

PR #1483, opened 2025-11-13
Fix Activity UI Issues

bitmold:
This PR fixes a lot of issues that were introduced when we recently upgraded our target API. Previously the AppManagerActivity and SettingsActivity would create new tasks in the app switcher, even after closing. Polluting the user's app switcher. I didn't realize how bad this had gotten, and it's pretty inexcusable so prioritized working on this for our next release. 

Tested on API 24 + 35.

The simplest thing was to make the app picker activity and settings activity fragments, this is generally what we ought to have been doing for a long time now anyway.


app/src/main/java/org/torproject/android/OrbotActivity.kt
```diff
@@ -262,12 +268,6 @@ class OrbotActivity : BaseActivity() {
         super.onActivityResult(requestCode, resultCode, data)
         if (requestCode == REQUEST_CODE_VPN && resultCode == RESULT_OK) {
             connectViewModel.triggerStartTorAndVpn()
-        } else if (requestCode == REQUEST_CODE_SETTINGS && resultCode == RESULT_OK) {
-            Prefs.defaultLocale = data?.getStringExtra("locale") ?: Locale.getDefault().language
```
bitmold: this logic, which only ran on legacy androids when locale changes were made in the settings app, can now just be handled in the settings fragment

tladesignz:
Wohooo! Jony Ive commemorative edition! 🤪

Looks all good to me!

I'm in the mood for merging today, so I'm going to bring this straight to master.

Great work, @bitmold!

PR #1485, opened 2025-11-13
Update built-in bridges

syphyr:
Run update-bridges.sh


syphyr:
It does look like `"meek":` is new

tladesignz:
> It does look like `"meek":` is new

Yep. It was decided to rename "meek-azure" to "meek". For a grace period, it actually contains both.
Anyway, good reminder to switch the deserialization instructions!

And your PR is another good reminder for cleaning this up even better:

This file should not be part of the repo anymore, but being fetched on build. I can extend the script, so it only does this when the file isn't there ore once per day.

But I have no idea where and how to hook that in the Gradle build scripts.

@syphyr, @bitmold, any suggestions?

tladesignz:
I pushed a commit which does the change from "meek-azure" to "meek", and I needed a `builtin-bridges.json` file for that.
So please don't try to merge this anymore.

However, I want to have this discussion, so let's keep this open until it is clarified how we proceed.

syphyr:
Make this task depend on a new task that runs the script.
https://github.com/guardianproject/orbot-android/blob/7de255cd657bc919c216811b357afd9e471ac9ec/app/build.gradle.kts#L180

meenbeese:
> > It does look like `"meek":` is new
> 
> 
> 
> Yep. It was decided to rename "meek-azure" to "meek". For a grace period, it actually contains both.
> 
> Anyway, good reminder to switch the deserialization instructions!
> 
> 
> 
> And your PR is another good reminder for cleaning this up even better:
> 
> 
> 
> This file should not be part of the repo anymore, but being fetched on build. I can extend the script, so it only does this when the file isn't there ore once per day.
> 
> 
> 
> But I have no idea where and how to hook that in the Gradle build scripts.
> 
> 
> 
> @syphyr, @bitmold, any suggestions?

I can work on integrating it as a Gradle build task that runs on every time the app is built or on any custom interval.

bitmold:
@tladesignz Doing something like @meenbeese did in their PR is how you'd hook into the gradle build process. 

Though I'm not sure I like the idea of this as part of the build process...

 If I wanted to build an older release of Orbot, I'd ultimately get different APKs depending on when I build the release (whatever the torproejct endpoint gives out at build time). I'm not a purist about reproducible builds with this project, but I think there's an important distinction between having two APKs that are functionally identical but have binary diffs because they were built differently/at different times and having two APKs that have different data in them that the program interacts with at runtime. 

Also if for whatever reason that URL goes down, moves, has a problem, can't be reached, etc etc you'd be unable to completely build Orbot. Could also see it being frustrating to people who have the code on their machine, but can't build a functional APK because their ISP at build time gets in the way of them making an HTTP request to torproject.org

So I think it _should_ be part of the repository, but fetched at release time - in such a way where you can't make an Orbot release without this script being run. Then when @n8fr8 or whomever is tagging a release, the JSON can be committed to git. Implementing this would be very similar to the type of gradle implementation meenbeese used in #1488.

C tor does something similar with `geoip` and `geoip6` - every `tor` release updates the geoip data, commits it to git, and everyone who builds that version of `tor` uses the same data no matter when they decide to build it. This type of thing could even cause us grief as maintainers, imagine a scenario where a google play APK and an f droid APK of the same version of Orbot, but since they were built at different times one has different PT config in it than the other. Or if we get a bug report, we'd never be able to reproduce exactly what the user had.

This way you don't need to ping torproject.org and be dependent on it successfully returning you JSON whenever you want to build the app, but also every release would have fresh JSON configuration.

bitmold:
IDK that was kind of a mouthful, sorry. Unfortunately I have another mouthful to say 🙃 

Personally, I think there's a lot of value to be able to clone a project - and if I have everything on my machine that's needed to build it - later go about building it offline/in restrictive interent conditions. if for instance I had the repository, built it (downloading gradle dependencies for the given commit if needed), deleted the local repo, and later recloned the repo I should be able to build it 100% for as long as I retain my gradle cache. This is something that we have now, and that we'd lose getting the JSON out of git.

Also if someday tor shuts down this endpoint in favor of something else or people are looking at this codebase 15+ years from now, the librarian in me thinks there's a lot of archivist/educational value for people interested in censorship circumvention tech to be able to clone some old project and understand how it went about tackling the problem in its day and age.

Anyway...

There's this script that @n8fr8 runs to make release APKs for distribution. https://github.com/guardianproject/orbot-android/blob/master/scripts/make-release-build.sh

Maybe instead of gradle, it'd be better to grab this file in this script and if it's different than last time commit it to git. You can do this for geoip too. This in theory solves the problem of someone forgetting to update this data for long periods of time between releases. 

If there's a long time between releases, developers could still run the script and submit a PR with the new JSON to get the latest configuration. But no one ever **needs** to run it, especially people who have gotten their hands on this code in censored/no internet areas and can't for whatever reason reliably hit torproject.org in order to compile a complete APK. 

I could see there being value in users opting in to update this data at runtime, maybe a preference where if the user is able to successfully connect to torproject.org over tor and it's been 2 weeks grab the JSON and update the local copy. This way users aren't left with senseless builtin defaults if they can't go about getting an update, we take a long time to get updates out, etc etc.

syphyr:
~Maybe this release script could check for an environment variable before trying to update these files.~
That wont work if these files are not being tracked.

Edit: Actually, why not keep tracking these files and the release script would simply create a new commit if there are changes.

bitmold:
> That wont work if these files are not being tracked.0

Right, I'm advocated for keeping it tracked in git, while also adding something at release time that runs the script no matter what.

My understanding is that @tladesignz is trying to reduce developer overhead - it's unacceptable that there have been releases of Orbo where this script hasn't been refreshed and by fetching the file at build time developers would at least in theory be way less likely to make that mistake again.

🤷‍♀️🤷‍♀️🤷‍♀️🤷‍♀️🤷‍♀️ Sorry, mega-rambly comments above really just reduce to the following short point:  by making the script be part of the release process for publihing APKs, you'd automate the problem of this script not being run consistently and reliably, but in such a way where you keep the many benefits of preserving the file in git.

Maybe there are other reasons to get rid of the file from the git repo, but I can't think of any besides overhead...

syphyr:
Instead of a gradle task, I think this would all be easier to accomplish within a separate bash script (ie; make_release).

```
./do_release_housekeeping.sh
./gradlew assembleRelease
```

You could even add ability to automatically create commits when housekeeping script is executed.

meenbeese:
> Instead of a gradle task, I think this would all be easier to accomplish within a separate bash script (ie; make_release).
> 
> ```
> ./do_release_housekeeping.sh
> ./gradlew assembleRelease
> ```
> 
> You could even add ability to automatically create commits when housekeeping script is executed.

I'm not a huge fan of writing more and more bash scripts, please check the new solution at #1485 that addresses these issues.

syphyr:
> I'm not a huge fan of writing more and more bash scripts, please check the new solution at #1485 that addresses these issues.

good work! I like what you did.

tladesignz:
Oh, my. I triggered something there.

I'm of the opinion that I don't want to have files in the repo which are external and change regularly.

That's why I suggested a build script, which only checks for updates once a day. (Using the file last-modified timestamp) And which should break the build if the file can't be acquired.

That way, there's no commits, no storage of foreign content, and always fresh data on production builds. Seems like the easy solution.

But, yeah, there might be people out there who want to build this behind a censor. And there's the reproducibility argument (which we don't have anyway, but ok.)

Then everything gets complicated.


So, ok I'm willing to bow to your wishes.

But which suggestion should it be now?  @meenbeese 's #1488 is declined now, do I get that right?

> I'm not a huge fan of writing more and more bash scripts, please check the new solution at https://github.com/guardianproject/orbot-android/pull/1485 that addresses these issues.

But this links to this exact PR. And I don't see any more commits here. I'm confused.

So what should it be now?

syphyr:
@tladesignz please take a look at the latest comments here https://github.com/guardianproject/orbot-android/pull/1488

I think most of the issues there have been resolved. There is just one additional suggestion I have which is currently being discussed in the other thread that I think should be implemented (about how often the new json should be downloaded).

tladesignz:
Ok, so we want the #1488 solution, but with a check, how old the file on disk is?

Good. I'd prefer a 1-day limit. (Mentioned that in #1488, too.)

tladesignz:
Great! After a lot of discussion, we probably found the optimal solution in #1488. Thank you, @bitmold, @syphyr and @meenbeese!

PR #1488, opened 2025-11-14
Automate builtin-bridges.json fetch in Gradle

meenbeese:
As discussed in #1485. @tladesignz please review.


app/build.gradle.kts
```diff
@@ -176,9 +177,33 @@ dependencies {
     androidTestUtil(libs.androidx.orchestrator)
 }
 
-tasks.named("preBuild") { dependsOn("copyLicenseToAssets") }
-tasks.register<Copy>("copyLicenseToAssets") {
+tasks.named("preBuild") {
+    dependsOn(copyLicenseToAssets)
+    dependsOn(fetchBuiltinBridges)
```
bitmold: Haven't had a chance to run this yet, but doesn't this happen every single time someone builds Orbot? Seems like doing something to limit the file is important. Copying the LICENSE file to the disk everytime a build runs doesn't bother me in the same way because it's a tiny file that's already on the disk.

syphyr:
I would still like the ability to run these scripts without having to start a build.

Whoever does the releases, could possibly have some kind of environment variable that the scripts checks for and if it exists then it will also try to update the built-in bridges and geoip files automatically.   Otherwise, that release process would be skipped for everyone else. 

Edit: Sorry, I guess that will not work since `builtin-bridges.json` is no longer being checked in to the repo with this PR.
Edit2: Why not keep tracking these files and just automatically create a commit if there are changes.  If there are changes, stop the build process and commit the change first.

meenbeese:
Made some changes to address the rightful concerns from the discussion:

1. The bridge task failing no longer blocks the build.
2. The bridge task is only built for the release flavor.
3. The json file is still tracked and is updated as a task.

syphyr:
These current changes look very good. Nice work.

app/build.gradle.kts
```diff
@@ -176,9 +177,87 @@ dependencies {
...
+
+    val bridgesDir = layout.projectDirectory.dir("src/main/assets")
+    val bridgesFile = bridgesDir.file("builtin-bridges.json").asFile
+
+    inputs.file(bridgesFile)
+
+    doLast {
+        val diffOutput = try {
+            providers.exec {
+                commandLine("git", "diff", "--name-only", bridgesFile.path)
```
syphyr: The problem here is that this will think the builtin-bridges.json needs updating if the repo is just  cleaned.

syphyr: Every time builtin-bridges.json is downloaded after the repo is cleaned, the obfs4 bridges are randomly generated.

syphyr:
Since `builtin-bridges.json` is randomly updated each time it is downloaded, using `git diff` will always be true, so I have changed the task to check if the bridges have been updated within the past 3 weeks instead.  Also, only download a new copy of `builtin-bridges.json` if it is older than 3 weeks. This also fixes the problem with starting with a clean build. 

https://github.com/syphyr/orbot/commit/91eca3e45079981440e311f15261a53b5ff469cd

meenbeese:
> Since `builtin-bridges.json` is randomly updated each time it is downloaded, using `git diff` will always be true, so I have changed the task to check if the bridges have been updated within the past 3 weeks instead. Also, only download a new copy of `builtin-bridges.json` if it is older than 3 weeks. This also fixes the problem with starting with a clean build.
> 
> [syphyr@91eca3e](https://github.com/syphyr/orbot/commit/91eca3e45079981440e311f15261a53b5ff469cd)

Thanks, I implemented your changes while keeping the two tasks separated. PTAL.

syphyr:
> Thanks, I implemented your changes while keeping the two tasks separated. PTAL.

The idea behind combining the tasks is that the bridge file should only be downloaded when its older than 3 weeks.  Otherwise, the download task always happens.

The json should only be downloaded right before the exception is thrown for being too old.

tladesignz:
I would actually suggest to freshly download, when the file is older than 1 day. These infos are changing quite often these days. And I'd rather don't want outdated configuration in the build.

syphyr:
But the json should not be downloaded if it is not outdated, otherwise every release build will always contain a different json.

tladesignz:
That's why I suggest to download it only once per day. And, yes. Every release build actually **should** contain a different JSON, because a newer release should also contain a newer configuration.

If the configuration didn't change, the download shouldn't change anything, anyway.

If this is about re-building an old release, then we would need to have another mechanism for that. You could e.g. just update the last-modified timestamp of the file, then there would be no download. Or a special build flag ("--reproducible" or something, which makes the download override.)

Not sure, what's the best way there. But we don't have reproducible builds anyway, or do we?

What **shouldn't** happen anymore, is, that release builds are created, which contain months-old configuration files.

syphyr:
> 
> If the configuration didn't change, the download shouldn't change anything, anyway.
> 

Every download of the json is randomly different. Currently, the time limit placed on the json is only when it is required be committed to the repo. But, when gradle downloads a new json each time a release is started, it's possible the json being used is no longer being committed, let's say for example the following scenario:

1) a release build is started and it throws an exception because the json is too old.
2) the new json that is always automatically downloaded is then committed and the release is started again.
3) the release is started again but another new json file is automatically downloaded and now the release includes a json that is not committed anywhere.

Preventing this is simple. Only download the
json if it exceeds the established time limit.

syphyr:
Also, the script used to manually update the json should still be available so the person doing the release can still manually update the json and commit it before starting the release build, to avoid any build failure.

tladesignz:
Ah. Yes. This should not happen, of course.

@meenbeese, are you able to adapt this script, so it does:

- Download a new JSON, when the current file is older than one day, otherwise *don't* download anything?
- Uses the last-modified timestamp of the file for that comparison.
- The git check could then just fail, if the repo is dirty, I'd say?

syphyr:
Combining the download task with the task that checks how old the commit is will solve all of these issues.

syphyr:
Well, the tasks actually dont *need* to be combined. If they are not combined, then the gradle output looks nicer (even though there is more code duplication).  Maybe it is actually better to keep the tasks separate after all. @meenbeese here is my second try.  This solution also fixes all the mentioned issues.

https://github.com/syphyr/orbot/commit/3c2a870523ae2e5ef4071522834e403ae4d22359

meenbeese:
> Well, the tasks actually dont _need_ to be combined. If they are not combined, then the gradle output looks nicer (even though there is more code duplication). Maybe it is actually better to keep the tasks separate after all. @meenbeese here is my second try. This solution also fixes all the mentioned issues.
> 
> [syphyr@3c2a870](https://github.com/syphyr/orbot/commit/3c2a870523ae2e5ef4071522834e403ae4d22359)

I actually would rather avoid code duplication. I was trying to keep the separation of concerns between the tasks clear, but I have now combined them into one, as well as changed the fetching interval to 1 day. I am trying to follow the discussion but I might have missed something, so please let me know if I skipped any of your discussed changes. Thanks for the suggestion though @syphyr @tladesignz!

The only point I don't agree on is the one on keeping the bash script. I feel like we should start relying more on Gradle tasks and perfecting said tasks than running scripts for different needs every time something pops up.

syphyr:
The point of keeping the bash script to update the bridges is to not break the ability to manually run the housekeeping. The gradle task is there just to make sure housekeeping is happening. But it's nice to be able to have the option to avoid the gradle errors.

tladesignz:
Ok. I like the file timestamp check and everything.

What I don't like is the overly specific test, if there was a commit not older than 1 day which contains the file.

Why should it be important, that there's a commit not older than one day?

The important question is, if `builtin-bridges.json` is fresh. 

If we just fetched it, and git still says "all fine" (in other words: the workspace isn't dirty, nothing changed which needs a commit), then we're good. Regardless how old the last commit of that file is.

Can you replace this check with a `git status` check instead? And having it in another task is actually a good idea.

There shouldn't be a release build done from dirty files, really.

syphyr:
Yes, the download check should still be checking the git log if it's older than one day. We are concerned about how old the commit is, not when it was last modified.

tladesignz:
> Yes, the download check should still be checking the git log if it's older than one day. We are concerned about how old the commit is, not when if was last modified.

No, we're not concerned about the commit date, we're concerned about how old the **actual data** is.

Just to be clear: The discussion about this specific topic ends here.

syphyr:
I'll just say, only checking the modified date does not ensure the file is tracked and then we have the same problem with including files in the release that are not committed anywhere.

The condition for when to download the file and when to throw the exception have to be the same test and that test is:

1) was the file recently updated AND committed.

meenbeese:
> Ok. I like the file timestamp check and everything.
> 
> What I don't like is the overly specific test, if there was a commit not older than 1 day which contains the file.
> 
> Why should it be important, that there's a commit not older than one day?
> 
> The important question is, if `builtin-bridges.json` is fresh.
> 
> If we just fetched it, and git still says "all fine" (in other words: the workspace isn't dirty, nothing changed which needs a commit), then we're good. Regardless how old the last commit of that file is.
> 
> Can you replace this check with a `git status` check instead? And having it in another task is actually a good idea.
> 
> There shouldn't be a release build done from dirty files, really.

Replaced with a `git status` check, thanks! Also, I think keeping it in the same task is fine here.

syphyr:
Checking the last modified date and then calling `git status` works perfectly. This ensures a clean build and all files are tracked.  Thanks for all the work.  It's much better this way.

tladesignz:
Thank you very much @meenbeese and @syphyr for the discussion and all the work!

PR #1503, opened 2025-12-03
Fixed notification action buttons.

tladesignz:
On newer Android, it seems that `PendingIntent`s with implicit `Intent`s to a class-private `BroadcastReceiver` don't work anymore.

The easiest solution seem to be sending the intents to the service itself, instead.

This definitely works and fixes the problem on my Android 15.

But do I overlook something?


bitmold:
This is excellent, I had tried in vain sometime in the past few weeks to address this. Thank you.

Issue #1479, opened 2025-11-11
[BUG] [base.apk] tor_cond_wait crash pthread_mutex_lock called on a destroyed mutex


n8fr8:
**Describe the Bug**
increase crash related to pthread and [base.apk] tor_cond_wait

**To Reproduce**
unshure

**Expected Behavior**
no crash

**What Custom Configuration Do You Use?**
none

**Smartphone (please complete the following information):**

1770300102 (17.7.0-RC-1-tor-0.4.8.19)
1770300101 (17.7.0-RC-1-tor-0.4.8.19)
1770300105 (17.7.0-RC-1-tor-0.4.8.19)
By Android version
Android 15 (SDK 35)
Android 14 (SDK 34)
Android 16 Beta (SDK 36)
Android 13 (SDK 33)
Others
By device
samsung e3q
samsung pa3q
samsung dm3q
samsung a15
Others
By issue visibility
Foreground

**Crash Logs (Advanced)**
Stack trace
1
Sample attributes:
[Samsung Galaxy A52s 5G](https://play.google.com/console/u/0/developers/6147505171125588874/app/4973609425637044610/vitals/crashes/197ce84281ace1223674b1a747b2a81d/details?days=28&versionCode=1770300105%2C1770300104%2C1770300102%2C1770300101&isUserPerceived=true#)
[Android 14 (SDK 34)](https://play.google.com/console/u/0/developers/6147505171125588874/app/4973609425637044610/vitals/crashes/197ce84281ace1223674b1a747b2a81d/details?days=28&versionCode=1770300105%2C1770300104%2C1770300102%2C1770300101&isUserPerceived=true#)
Version: 1770300102 (17.7.0-RC-1-tor-0.4.8.19)Occurred: 19 days ago
Thread
FORTIFY: pthread_mutex_lock called on a destroyed mutex (0x<sanitized>)
*** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
pid: 0, tid: 8602 >>> org.torproject.android <<<

backtrace:
  #00  pc 0x000000000005b6f0  /apex/com.android.runtime/lib64/bionic/libc.so (abort+168)
  #01  pc 0x000000000005d624  /apex/com.android.runtime/lib64/bionic/libc.so (__fortify_fatal(char const*, ...)+128)
[...]

PR #1511, opened 2025-12-09
Feature/service process

tladesignz:
I just put `OrbotService` and `TorService` in another process together.

The interface between OrbotService and the rest of the app seems just to be the broadcasts, which I could switch from local broadcasts to normal ones.

This *seems* to work on first glance.

Is this a bad idea? Why? @syphyr, what do you think?

related: #1479

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -116,6 +116,7 @@
             android:name=".service.OrbotService"
             android:enabled="true"
             android:exported="true"
+            android:process=":tor"
             android:foregroundServiceType="systemExempted"
             android:permission="android.permission.BIND_VPN_SERVICE"
             android:stopWithTask="false">
@@ -123,6 +124,11 @@
                 <action android:name="android.net.VpnService" />
             </intent-filter>
         </service>
+
+        <service android:name="org.torproject.jni.TorService"
+            android:process=":tor"
+            tools:node="merge"
+            tools:replace="android:process" />
 
         <service
             android:name=".ui.OrbotTileService"
--- a/app/src/main/java/org/torproject/android/OrbotActivity.kt
+++ b/app/src/main/java/org/torproject/android/OrbotActivity.kt
@@ -13,7 +13,6 @@ import android.util.Log
 import android.view.View
 import android.view.WindowInsetsController
 import androidx.activity.addCallback
-
 import androidx.activity.result.contract.ActivityResultContracts
 import androidx.activity.viewModels
 import androidx.annotation.RequiresApi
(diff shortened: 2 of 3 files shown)


syphyr:
What happens if the owning process dies? Does Tor need to use `--__OwningControllerProcess` to check if the owning process still exists?

tladesignz:
> What happens if the owning process dies? Does Tor need to use `--__OwningControllerProcess` to check if the owning process still exists?

I don't see why that would be necessary. The main process is the app. The forked process is `OrbotService` (the `VpnService` implementation) together with `TorService` (the actual Tor).

The `VPNService` and the app can work without each other just fine. If the app dies, Tor doesn't need to worry about that. If Tor and along with it the `VPNService` dies, that's a state change of the VPN, so the app should get informed about that by the system. (Or am I assuming too much and in fact it doesn't?)

Main (sought after) benefit: the app keeps running, if the service process dies, instead of getting torn down with it. (Or doesn't it?)

syphyr:
@tladesignz this comment seems relevant https://github.com/guardianproject/tor-android/pull/61#discussion_r771310086

tladesignz:
I can't see what exactly you consider relevant from that discussion.

Our main concern is Tor not crashing the app and vice versa.

Goal seems achieved.

If you think not, please explain how not!

syphyr:
I was referring to the part of the discussion that says:
"Android doesn't guarantee that a service will be destroyed if a bound client in another process crashes"

tladesignz:
Well, perfect. The VPN process *shouldn't* be destroyed, when the app process crashes.

It seems, this part is still not clear:

I put the OrbotService (which implements the foreground VpnService with all the Tor config/start/stop code) **together** with Tor itself in one process, which is **another** process then where the app is running.

If Tor crashes, then also the Orbot foreground service dies with it. And vice versa. The rest of the app stays alive.

If the *app* crashes, the Orbot foreground service together with Tor stays alive.

That sounds pretty sound to me.

The only question is: is there an edge case where things go off the rails?

syphyr:
So far, I have not seen any issues with these changes.

n8fr8:
Is there a privacy concern with moving to system-wide broadcasts vs local broadcasts? Does this mean any app on the device can register to receive info about the user's tor session/circuits/etc?

tladesignz:
> Is there a privacy concern with moving to system-wide broadcasts vs local broadcasts? Does this mean any app on the device can register to receive info about the user's tor session/circuits/etc?

Well, yeah. Since `OrbotService` has set `android:exported="true"`, because it already sends public broadcasts.

PR #1513, opened 2025-12-12
Implementing a new pill navbar layout

meenbeese:
This PR is the spiritual successor to #1391, as it further tweaks the navbar design by adopting a modern pill layout. It is almost like a mix of Android's M3E and iOS' Liquid Glass, which I think feels quite at home on Android. I implemented a smooth animation that reveals the labels of elements only when they are selected, as was done before. And as always, the core logic remains unchanged. I feel quite good about the margins and colours after quite a few tweaks, however implementing animations has been hell, so I am only open to code contributions there if you are actually able to improve the animation smoothness.

<table>
    <tr>
        <th>Before</th>
        <th>After</th>
    </tr>
    <tr>
        <td>
</td>
        <td>
</td>
    </tr>
</table>
<table>
    <tr>
        <th>Before</th>
        <th>After</th>
    </tr>
    <tr>
        <td></td>
        <td></td>
    </tr>
</table>


syphyr:
This looks really good. Nice work! No issues.

syphyr:
One small thing I noticed is that when Orbot starts up for the first time, the navbar position starts on the "More" fragment and then quickly switches over to the "Connect" fragment.  It would be better if Orbot started up on the "Connect" fragment and did not have to swtich fragments on initial startup.

meenbeese:
> One small thing I noticed is that when Orbot starts up for the first time, the navbar position starts on the "More" fragment and then quickly switches over to the "Connect" fragment. It would be better if Orbot started up on the "Connect" fragment and did not have to swtich fragments on initial startup.

I tried to address it with my last commit, but the navigation code is messy and should probably be refactored anyhow.

bitmold:
@meenbeese 

Has this been addressed in your most recent commits:

>  One small thing I noticed is that when Orbot starts up for the first time, the navbar position starts on the "More" fragment and then quickly switches over to the "Connect" fragment. It would be better if Orbot started up on the "Connect" fragment and did not have to swtich fragments on initial startup.

meenbeese:
> @meenbeese
> 
> Has this been addressed in your most recent commits:
> 
> > One small thing I noticed is that when Orbot starts up for the first time, the navbar position starts on the "More" fragment and then quickly switches over to the "Connect" fragment. It would be better if Orbot started up on the "Connect" fragment and did not have to swtich fragments on initial startup.

I just checked and that issue has been resolved. Also rebased on latest master.

syphyr:
The navbar position still starts on the More fragment when initially starting Orbot. Nonetheless, I like the design of the new pill navbar.

bitmold:
Hi @meenbeese I'm closing this now since it's been buggy and out of sync with the main branch sometime. Happy to revisit it if you ever end up taking another stab at it !

PR #1523, opened 2025-12-18
Prevent crash by selecting Connect too fast

### LINKED ISSUE #1539: [BUG] orbot wont wont start VPN connection andeietherboth crashes or  any of the buttons dont work
Bug: Pressing "Start VPN" on the Connect page won't enable it. This affects the "Choose how to connect" option, as well as other buttons like "Settings" and misc functions, which do work.

Steps to reproduce: Connecting and disabling frequently causes Orbot to either crash or the connection buttons to become unresponsive. Leaving or closing the app triggers a pop-up that asks, "Orbot is causing issues. Would you like to clear cache or data?" (You might have to do this a couple of times.)

[LOGS] "Orbot is Shutting Down
Orbot is Deactivated
Updating torrc custom configuration…
Orbot is Starting…"

These logs were still on and hadn't been updated or deleted as they used to be.

I downloaded Orbot from the official source on the Play Store, and it is on the latest update. I have a good internet connection, and I have DNS disabled so Orbot won't cause any conflicts with my internet.

syphyr:
Give OrbotService time to check if Tor is already running

Fixes #1539

### DIFF
--- a/app/src/main/java/org/torproject/android/ui/connect/ConnectFragment.kt
+++ b/app/src/main/java/org/torproject/android/ui/connect/ConnectFragment.kt
@@ -12,6 +12,7 @@ import android.view.View
 import android.view.ViewGroup
 import android.view.animation.Animation
 import android.view.animation.AnimationUtils
+import android.widget.CompoundButton
 import androidx.activity.result.contract.ActivityResultContracts
 import androidx.appcompat.app.AppCompatActivity
 import androidx.core.content.ContextCompat
@@ -106,7 +107,7 @@ class ConnectFragment : Fragment(),
             }
         }
 
-        binding.switchConnect.setOnCheckedChangeListener { _, value ->
+        binding.switchConnect.setOnThrottledCheckedChangeListener { _, value ->
             if (value)
                 startTorAndVpn()
             else
@@ -406,5 +407,37 @@ class ConnectFragment : Fragment(),
         )
 
         refreshMenuList(requireContext())
+    }
+}
+
+private var lastChange = 0L
+private const val DEFAULT_THROTTLE_INTERVAL = 5000L
+
+/**
+ * Prevents rapid consecutive checked-change calls while visually indicating cooldown
+ */
+fun CompoundButton.setOnThrottledCheckedChangeListener(
+    intervalMs: Long = DEFAULT_THROTTLE_INTERVAL,
+    onCheckedChange: (button: CompoundButton, isChecked: Boolean) -> Unit
+) {
+    setOnCheckedChangeListener { buttonView, isChecked ->
+        val now = System.currentTimeMillis()
+        if (now - lastChange >= intervalMs) {
+            lastChange = now
+            onCheckedChange(buttonView, isChecked)
+        } else {
+            buttonView.isChecked = !isChecked
+            val timeRemain = intervalMs - (now - lastChange)
+
+            // Visual feedback
+            buttonView.isEnabled = false
+            buttonView.alpha = 0.38f
+
+            // Restore after delay
+            buttonView.postDelayed({
+                buttonView.isEnabled = true
+                buttonView.alpha = 1f
+            }, timeRemain)
+        }
     }
 }


n8fr8:
A good idea, but I am not sure that is the place to put it. Perhaps the OrbotService should throttle/queue calls itself?

syphyr:
The problem was that I was able to send two ACTION_START intent calls to OrbotService.  Maybe `startTorAndVpn()` could check if Tor is already started before sending ACTION_START?  I would rather not use sleep calls if possible. I'm just not sure what the best way to do this is.

syphyr:
Instead of using a sleep here, I think a better way is to use a companion object to keep track of the state. I will test and post an update if it works as expected.

syphyr:
After some stress testing, everything I have tried seems to still not work 100% of the time.  It may have improved the race a little bit, but its not totally fixed.  Maybe someone with better knowledge can come up with an improved solution.  @n8fr8 I think your idea of a command queue seems like the best approach, but I do not think I am able to implement that.  Feel free to close this PR if you can come up with another solution.  

The way I am testing this is just selecting connect and cancel a bunch of times in a row and orbot will eventually crash within a few seconds.

syphyr:
There needs to be something that ignores the "Cancel" button if Tor is already in the process of shutting down.

n8fr8:
Please see the new UX ideas I've just pushed here as a part of a new 17.8 Alpha. Thinking about a toggle vs button.

syphyr:
> Please see the new UX ideas I've just pushed here as a part of a new 17.8 Alpha. Thinking about a toggle vs button.

I like the new toggle, but it still crashes if I switch the toggle back and forth several times within a few seconds.  I'm thinking something needs to be done within OrbotService and checks if Tor is currently shutting down before it tries to shut it down again.. and also the same with Tor starting.  Orbot seems to crash when the service tries to start or stop a process thats currently starting or stopping.

syphyr:
I tried something like this, but it did not work:

```
--- a/app/src/main/java/org/torproject/android/service/OrbotService.java
+++ b/app/src/main/java/org/torproject/android/service/OrbotService.java
@@ -686,6 +686,8 @@ public class OrbotService extends VpnService {
             if (TextUtils.isEmpty(action)) return;
             switch (action) {
                 case ACTION_START -> {
+                    if (mCurrentStatus.equals(STATUS_STARTING))
+                        break;
                     var transport = Prefs.getTransport();
                     transport.start(OrbotService.this);
                     startTor();
@@ -703,6 +705,8 @@ public class OrbotService extends VpnService {
                     }
                 }
                 case ACTION_STOP -> {
+                    if (mCurrentStatus.equals(STATUS_STOPPING))
+                        break;
                     var userIsQuittingOrbot = mIntent.getBooleanExtra(ACTION_STOP_FOREGROUND_TASK, false);
                     // When user cancels connecting, make sure, the SmartConnect timer is also cancelled.
                     SmartConnect.cancel();
```

syphyr:
Another idea might be to rate limit the toggle.

syphyr:
The rate limiting idea seems like it would work, but I do not know how to actually implement it.  Users might not know they are being rate limited, so  perhaps a toast message could  pop-up saying its being rate limited.

syphyr:
I have fixed the crashing by implementing a throttle on the toggle switch.  This seems to be working great so far. Please let me know what you think. @n8fr8 

I have set the rate limit to 5 seconds.  I think it should not take longer than 5 seconds for Tor to finish starting.  The user will get a toast message if they keep switching the toggle more than one time per 5 seconds.

syphyr:
I just need to figure out how to disable the toggle while it is being throttled.

syphyr:
I set the rate-limit down a little to 4 seconds.  Since this currently does prevent crashes, I think this is ready for review.

syphyr:
I finally fixed the toggle so it stays in sync with the correct state while its being throttled.  This works perfectly now. I can select the toggle a bunch of times in a row, but only the first touch is used. A toast message pops up while the toggle is being throttled saying to retry again in a few seconds.  This solves all the crashing issues.

syphyr:
I removed the usage of WeakHashMap because there is only one toggle that is being throttled.  WeakHashMap only needs to be used if there were multiple toggles. This also reduces memory usage.

Also, the toast message will now display the exact time remaining for the rate limit on the toggle.

syphyr:
I changed the throttle time back to one click per 5 seconds.  Sometimes Tor takes just under 5 seconds to start.  This ensures that Tor has the time it needs to start properly before trying to stop it.

syphyr:
I am interested in hearing how this affects different devices that are faster and slower than the one I'm stress testing with.

n8fr8:
Just catching up on all this now

n8fr8:
I wonder if instead of showing the Toast telling them to "retry in X seconds", we could just have a waiting dialog that shows for that # of seconds before starting Tor?

n8fr8:
(Otherwise this is working very well for me)

syphyr:
In regards to using a waiting dialog instead of a toast, I'm wondering how that would work if the user pressed the toggle multiple times.  Currently, the toast messages are queued, depending on how many times the toggle is selected.

syphyr:
I really like your idea to replace the toast with a dialog. It looks much more professional!  I first tried a wait dialog with a spinner, but then updated it to use a progress bar.  I like the dialog with the progress bar best. Please let me know which one you like.

syphyr:
I changed the wait dialog message to say "Waiting X seconds for Tor process to finish"

syphyr:
One last change to fix the wait dialog layout. Now it looks really good to me.

syphyr:
Well, one more change. I decided to use xml for the layout which proved to be more reliable and easier to maintain.  I've also changed the color of the progress bar to match Orbot's color scheme. ;)

I think its ready now.

bitmold:
Hi @syphyr , I'm going to review and test this in the next day. I might make some changes to the UI aspects of your implementation, but appreciate this change and how you implemented this throttling.

syphyr:
I added support for material 3. Thanks for testing!

bitmold:
I just tested this and there are a handful of UI issues so this is not ready for merging. 

- the app isn't responsive. users have a reasonable expectation to be able to immediately cancel something that they started and will get frustrated/impatient with the app if they are told "no, wait 5 seconds". 
- it's messy to see two progress bars moving on the screen at the same time 
- if i'm being told to wait 5 seconds while the progress bar is filling up, I'd expect that my action will occur as soon as the progress bar reaches 100%, almost as if Orbot is "loading" my click. When the progress bar fills up and the dialog disappears the app seems broken almost like it didn't do the thing I asked. It's not immediately clear to me in anyway, that I am now supposed to repeat my click all over again in order to do the thing I had intended to do 5 seconds earlier. 

I appreciate the intention behind this and maybe it can be reworked another way. Putting a throttling extension on a `CompoundButton` button is a cool idea.

If you want to take another stab for it, grab my commit here based off of your work fb7eeac48a152648df6cd3b1e684ffc33f890bcf

Nothing functional has changed, just removed unused imports, redundant qualifiers, simplified expressions, and moved the extension function to the bottom of the file.

bitmold:
Sorry, I'm now realizing that some of the things I had cleaned up were from the master branch and not your PR. I didn't expect to see some of those things in master...

syphyr:
The purpose of a wait dialog is to prevent any additional input.  There is no way to throttle the toggle and allow the user to cancel operations on the toggle at the same time.  Under normal usage, the user will probably not even notice the toggle is being throttled to one change per five seconds.

So, would it be better to remove the progress bar from the wait dialog and just have a spinner? I do not see two progress bars at the same time on my device because the wait dialog covers everything behind it.

If it is confusing as to what the wait dialog is doing, perhaps the message should be changed to better clarify what is happening. Maybe it would be less confusing if the message on the wait dialog said something like, "The connect toggle is being throttled".  That should indicate that the last input is ignored until the toggle is no longer being throttled.

syphyr:
@bitmold It sounds like you would prefer just a toast message instead of a wait dialog, like my original patch https://github.com/guardianproject/orbot-android/pull/1523/changes/f23aeb50cc3f0b88a0eefcda849f60dbeca77e9e used.  Maybe try with just that one commit and see how it feels. A toast message is much less obtrusive than a wait dialog.

n8fr8:
What if we disable the switch instead? Is that the "throttling extension on CompoundButton" in a way?

syphyr:
The toggle is already ignoring any input when it is being throttled, which is pretty much the same thing as being disabled.

Also, queuing the input on the toggle when the toggle is rapidly being switched does not sound like the right thing to do either.

I feel like most of the issue here is related to how the user is notified that the toggle is being throttled and the wait dialog with the progress bar is just a little overkill.

syphyr:
I have made the following changes:

1. Remove toast and wait dialog completely
2. Visually disable button by:
   - Setting `isEnabled = false`
   - Reducing alpha to 50%
3. Automatically restore normal state after cooldown period
4. Cleaner behavior than toasts and wait dialog since it prevents interaction entirely during cooldown

syphyr:
@bitmold I think the latest changes here should fix all the issues you mentioned. I just tested it and it seems much better now and also much more simplified.

syphyr:
Changed toggle to 38% opacity for disabled state. This is what Material recommends.

kasia-de:
Why not do a lock instead of throttling or queueing?

I assume it would take less code to lock immediately after stopping Tor, then unlock after we know Tor has stopped?

syphyr:
> Why not do a lock instead of throttling or queueing?
> 
> I assume it would take less code to lock immediately after stopping Tor, then unlock after we know Tor has stopped?

The real issue is that we need C binding that will unload `libtor` via `dlclose` after Tor stops, but that will not happen until we switch over to Arti.  Ideally, there needs to be some kind of callback to tell the toggle when Tor has finished stopping and starting.

n8fr8:
Thanks for all the work here. Just waiting for @bitmold to sign off and merge, and then we will move to a beta release.

bitmold:
I'm going to make some quick UI adjustments, but I like your new implementation - merging soon - thank you !

PR #1524, opened 2025-12-19
Fix string for direct connection

syphyr:
Transport.NONE is always a direct connection


syphyr:
This string should describe what the type of connection is being used in the connect fragment.

n8fr8:
Yes, but I wanted to the user to know they can choose the connection type there as well. Need to figure out a better than showing the wrong string, so will merge your chance!

PR #1529, opened 2025-12-21
Refactor Tor status handling to use TorStatus enum

meenbeese:
Provides a single SoT by wrapping the JNI primitives in one place and exposing it as a type-safe enum. Also cleaned up some related code in the OrbotService using early returns and cleaned up the top-level variable declarations.

### DIFF
--- a/app/src/main/java/org/torproject/android/service/OrbotConstants.kt
+++ b/app/src/main/java/org/torproject/android/service/OrbotConstants.kt
@@ -117,19 +117,6 @@ object OrbotConstants {
     const val LOCAL_ACTION_PORTS = "ports"
     const val LOCAL_ACTION_V3_NAMES_UPDATED = "V3_NAMES_UPDATED"
 
-
-    /**
-     * All tor-related services and daemons are stopped
-     */
-    const val STATUS_OFF = TorService.STATUS_OFF
-
-    /**
-     * All tor-related services and daemons have completed starting
-     */
-    const val STATUS_ON = TorService.STATUS_ON
-    const val STATUS_STARTING = TorService.STATUS_STARTING
-    const val STATUS_STOPPING = TorService.STATUS_STOPPING
-
     /**
      * The user has disabled the ability for background starts triggered by
      * apps. Fallback to the old Intent action that brings up Orbot:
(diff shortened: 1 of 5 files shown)


bitmold:
This is nice work and definitely a smart way to go about handling Tor Status instead of using the Strings defined in [tor-android](https://github.com/tor-andriod). However, as I said before, for the time being I'm not interested in merging these types of refactors that don't address any bugs - especially in critical parts of the project like `OrbotService`.

meenbeese:
I was mostly following up refactorings similar to what you have done before, to reduce the complexity of OrbotService...

bitmold:
@meenbeese I appreciate your enthusiasm, but don't appreciate that ... at the end of your comment,  it can come off as passive aggressive. As I told you [before:](https://github.com/guardianproject/orbot-android/pull/1462#issuecomment-3508863046)

> Currently not interested in these minor rewrites and refactoring. Its a lot to thoroughly test everything and ensure there are no regressions. If theres something that seems especially worth refactoring open a ticket and we can talk about it.

Most of my refactorings occur when I'm working in an area of the project that I am having to modify for other reasons, because I'm already going to have to spend time thoroughly testing those areas regardless, and even *then*, there is still an inherent cost and risk. But it's in those moments when the trade off between time investment and cost of possible risk that "reducing complexity" seems most worth it to me. `OrbotService` happens to be the most intricate part of the program, so I try to approach it as delicately as possible. 

Stopping what I'm doing to understand these new changes, review them, test them as thoroughly as I can and then merge them - all of this increases complexity and overhead for me while consuming my time. If something becomes broken that the contributor didn't catch I then have to either to fix it myself, go back and forth suggesting fixes on the PR, or just close the PR. With all this in mind, I sometimes find it not worth going down this road on behalf of someone who I don't personally know at all.

Right now, with Orbot, I love bug fixes, security updates, meaningful performance improvements, and again, am *open* to refactorings - but please - if you want to make sure  any small refactorings or UI adjustments get merged - just take a few minutes to open an issue and make a case for them instead of sending me a pull request. We can talk it out and figure out if that's something I find worth pursuing.

Thanks for all of the other great work you've done - I genuinely appreciate it.

Sincerely - the person that spends her very real time + energy reviewing and testing your code.

Issue #1531, opened 2025-12-22
[BUG] Changes in "Outbound network proxy" are not respected.


subsys-R9boq8:
**Describe the Bug**
Changing "Outbound network proxy" doesn't work and Orbot still attempts to reach the old proxy server.

**To Reproduce**
0. Have two devices (A, B) in local network running Socks5 proxy. Orbot was originally configured to use A.
1. Shutdown Socks5 proxy on A.
2. Start Orbot and configure "Outbound network proxy" to Socks5 proxy server running on B.
3. Start connecting to Tor network.
4. The connection progress stagnated at 3%, log reported that Orbot is still trying to connect to the Socks5 proxy server on A but failed (Connection Refused) since proxy server on A has been shutdown.

**Expected Behavior**
New outbound network proxy settings are respected.

**What Custom Configuration Do You Use?**
"Start Orbot on boot" disabled. "Allow background starts" disabled. "Power user mode" enabled. "Screen security" disabled (to capture recording).
"Outbound network proxy" is configured.

**Screen Recording (WebM/VP9)**
https://github.com/user-attachments/assets/a146421c-8765-4d43-be2d-08c2901795b0

**Smartphone (please complete the following information):**
- Device: Oneplus 8T
- OS: Android 15
- Version: 17.8.0-ALPHA-1-tor-0.4.8.21-1-g0e1e3d51

syphyr:
Currently, the proxy settings have been moved here:
https://github.com/guardianproject/orbot-android/blob/58998a386fcd2d476dfbf7aaccc26e62ea1f3153/app/src/main/java/org/torproject/android/service/circumvention/Transport.kt#L131

But, this worked correctly when the proxy settings were located in https://github.com/guardianproject/orbot-android/blob/master/app/src/main/java/org/torproject/android/service/tor/TorConfig.kt

So, any easy fix would be to place the proxy settings back into TorConfig.kt where it was generated each time Tor stopped and started. Example: https://github.com/guardianproject/orbot-android/commit/63dc13154ac295467ae1055decab74745e6cf8be

Perhaps @tladesignz has another solution that keeps these proxy settings within Transport.kt.

tladesignz:
Nice guess, @syphyr, but that's not the problem. The settings are still coming from where they're stored - in Shared Preferences. And they're constructed in TorConfig on restart.

While checking this, I had a different but similar problem: I couldn't switch out of the Snowflake bridge.

[...]

PR #1532, opened 2025-12-23
Fix preferences access through multiple processes

tladesignz:
There's one more interface between the service process and the rest of the app: The preferences file.

There's caching going on in both processes, the change from #1511 broke this interface.

I introduced a `ContentResolver` for preferences, to make it IPC-safe.

Please test, if you can find a loophole.

I *think*, we should be good now, as all preferences file access should now be contained in the app process: The settings activities are in the app, and the content resolver should be started on app start, which is inside the app process.


tladesignz:
Will fix #1531.

bitmold:
- found that the unit tests wouldn't build which you wouldn't see unless you were building the entire project
- fixed a NPE, the apps wants the torified apps string to be an empty string and not a null. it'd crash with these changes when you open the choose apps screen.

bitmold:
looks good, didn't find anything else beyond those two issues

PR #1533, opened 2025-12-25
Bug Fixes Around New Connect Screen Log Printng 

bitmold:
- Noticed that the `ConnectFragment` was checking if a log contained "Connected to control port" and if contained would print "Tor service has started". The former String isn't  hardcoded but actually is `R.string.log_notice_connected_to_tor_control_port` - one of our dynamically localized ones. This means that prior to this fix, "Tor service has started" would only be displayed if Orbot is in Enlgish.
- Moved longwinded if/else/else/else avalanche out of `ConnectFragment` and into a more compact file that maps Strings to we want to print.


bitmold:
actually just merging this, it's harmless and I want to keep making fixes around `ConnectFragment`

PR #1554, opened 2026-01-11
ACTION_STOP is now public in TorService

syphyr:
It is no longer necessary to define ACTION_STOP in OrbotConstants because ACTION_STOP is now made public in TorService.

This PR requires latest version of tor-android which includes the following change: https://github.com/guardianproject/tor-android/commit/7e5b3c5714c560a82144eaaf6d8f93f61185f658

Fixes build error:

app/src/main/java/org/torproject/android/service/OrbotService.java:635: error: reference to ACTION_STOP is ambiguous
both variable ACTION_STOP in TorService and variable ACTION_STOP in OrbotConstants match

### DIFF
--- a/app/src/main/java/org/torproject/android/service/OrbotConstants.kt
+++ b/app/src/main/java/org/torproject/android/service/OrbotConstants.kt
@@ -5,8 +5,6 @@ import androidx.core.net.toUri
 object OrbotConstants {
     const val TAG = "Orbot"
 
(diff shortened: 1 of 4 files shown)


bitmold:
Thanks, but I know this. I made `ACTION_STOP` public in `TorService` for this reason. However, there is no latest-build of tor-android yet - and so I'm not merging this for now.

PR #1577, opened 2026-01-30
Clean up margins/paddings on PT sheet.

tladesignz:


bitmold:
Looks great, though we should also still remove that translatable=false on the DNS Tunnel string too, right. That was the other small feedback from #1572 that never saw the light of the day.

_Ok, praying to the heavens and all the stars above that we can all see this comment after I click the "Comment" button... :crying_cat_face:_ :upside_down_face:

PR #1586, opened 2026-02-04
Additional GP MOAT querying

tladesignz:
- Housekeeping: Renamed MEEK_AZURE to MEEK.
- Improved DNSTT support.
- Added `MoatTunnel` as an abstraction to handle different tunnel transports for MOAT access.
- Added additional querying of new GP MOAT service and merge responses, if any.

### DIFF
--- a/app/src/main/java/org/torproject/android/service/circumvention/AutoConf.kt
+++ b/app/src/main/java/org/torproject/android/service/circumvention/AutoConf.kt
@@ -41,7 +41,6 @@ object AutoConf {
                 return PasswordAuthentication(tunnel.config, "\u0000".toCharArray())
             }
         }
-
         Authenticator.setDefault(authenticator)
 
         val api = MoatApi.getInstance(context, tunnel)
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/torproject/android/service/circumvention/AutoConf.kt
```diff
@@ -27,24 +25,25 @@ object AutoConf {
     ): Pair<Transport, List<String>>? {
         var cannotConnectWithoutPt = cannotConnectWithoutPt
 
+        val tunnel = MoatTunnel.TOR_PROJECT
+
         val done = fun(conf: Pair<Transport, List<String>>?): Pair<Transport, List<String>>? {
```
bitmold: `conf` could default to `null` ie `fun(conf: Pair<Transport, List<String>>? = null)` so as to invoke `done()` instead of `done(null)` on failure

bitmold:
Looks great, I need a little more time to test it but can merge tomororw.

Issue #1583, opened 2026-02-02
[BUG] Screen elements overlap each other


myrightvoice-png:
**Describe the Bug**
Screen elements overlap (see screenshot)

**To Reproduce**
1. Go to main screen

**Expected Behavior**
Difficult to interact with screen elements 

**What Custom Configuration Do You Use?**
Bigger font size

**Screenshots**
Attached 

**Smartphone (please complete the following information):**
- Device: Samsung Galaxy A54 5G 
- OS: Android 16
- Version: v17.8.0 RC 3

**Crash Logs (Advanced)**
N/A

**Additional Context**
N/A

![Image](https://github.com/user-attachments/assets/0af7e4dd-6815-4bec-a1c4-2d3438bc795a)

myrightvoice-png:
Buttons are also affected when font size increases.

![Image](https://github.com/user-attachments/assets/a4b31090-0014-40f8-9775-cf2fb1d9ed01)

PR #1606, opened 2026-02-17
Fix #1595 issue with navbar, sidebyside, floating window

### LINKED ISSUE #1595: In new design, connect button is hidden behind the menu in floating window mode.
Image

bitmold:
The navbar would sit on top of the UI content which broke things on small/floating displays causing undesirable overlapping/scrolling behavior. This fixes that and it looks consistent on any sized screen. Also trimmed down egregious margins for things like Orbie's shadow, which took up so much real estate on smaller displays.

This PR closes #1595 and #1583

On Pixel 2 (running API 28, has a small screen, if necessary the UI will still scroll) 
image
image
image

On a Pixel 7 (API 36, bigger screen, also with modern android edge2edge) it renders fine and looks the same as we're used to:
image

 Small side/by/side correct: 
image

### DIFF
--- a/app/src/main/res/layout/config_connection_bottom_sheet.xml
+++ b/app/src/main/res/layout/config_connection_bottom_sheet.xml
@@ -72,15 +72,15 @@
                 android:layout_marginStart="12dp"
                 android:layout_marginTop="4dp"
                 android:layout_marginBottom="4dp"
-                android:hint="@string/my_country_optional"
-                android:inputType="text"
-                android:textColorHint="@android:color/darker_gray"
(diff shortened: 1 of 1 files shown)


bitmold:
As a free form window it'll try to be the width in the spec, but will make compromises to not break the UI as needed. 
image

image

bitmold:
@n8fr8 after dealing with a merge conflict this is ready for review

bitmold:
Thanks for looking, just wanted a second set of eyes on it / it getting tested on an additional configuration.

PR #1608, opened 2026-02-18
Use internal Kotlin provided by AGP 9

syphyr:
The plugin `org.jetbrains.kotlin.android` is deprecated in AGP 9 and will stop working with AGP 10.  The android kotlin plugin is now built into AGP 9.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -5,7 +5,6 @@ import java.net.URI
 import java.util.*
 
 plugins {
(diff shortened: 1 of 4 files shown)


bitmold:
Oh, I'm very happy about this! It was annoying how complicated it was to rename the APK files with the new API. These are the last two gradle things we have yet to figure out. Testing it right now, **thank you** :star:

PR #1610, opened 2026-02-21
Add initial support for IPtProxy's new SnowflakeProxy callbacks

syphyr:
IPtProxy has just added support for additional callbacks for proxy client events and statistics. These changes will be in the next IPtProxy release.
https://github.com/tladesignz/IPtProxy/commit/922c1796a8253010162319f723af27ffd088d421

This PR fixes the build for Orbot by adding support for those new SnowflakeProxy callbacks.

@tladesignz Please review

### DIFF
--- a/app/src/main/java/org/torproject/android/ui/kindness/SnowflakeProxyWrapper.kt
+++ b/app/src/main/java/org/torproject/android/ui/kindness/SnowflakeProxyWrapper.kt
@@ -1,6 +1,6 @@
 package org.torproject.android.ui.kindness
 
-import IPtProxy.SnowflakeClientConnected
+import IPtProxy.SnowflakeClientEvents
 import IPtProxy.SnowflakeProxy
 import android.content.Context
 import android.os.Handler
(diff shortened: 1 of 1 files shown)


n8fr8:
This would be great to have, so we can provide more interactive feedback to the user

syphyr:
> This would be great to have, so we can provide more interactive feedback to the user

IPtProxy 5.2.0 has just been released, so this can now start to be tested here.

tladesignz:
Hey, @syphyr, thanks for this PR. Since the update to IPtProxy is actually missing, and pulling your PR into my branch which does that just to give you credits started to become complicated, I just replicated your change [here](https://github.com/guardianproject/orbot-android/pull/1628) and I will close this one.

Thanks for you pushing us always. :-)

tladesignz:
> This would be great to have, so we can provide more interactive feedback to the user

In what way?

To be clear, I added this to IPtProxy on request of another user, because they wanted the stats.
I tried it out myself, and they're… underwhelming. The `stats` method is getting called about 2 times per second, mostly with zeros everywhere. And when there's data in there, it's still mostly zeros and then some bytes from some sent or received packet. So putting this into something useful will create some work.

I threw in `connectionFailed` and `disconnected` just for good measure.

Didn't have anything in mind, really.

What are you thinking how we could make use of this info?

tladesignz:
Turns out, the wild stats call interval is a bug: https://github.com/tladesignz/IPtProxy/issues/81#issuecomment-3999949831

tladesignz:
Fix will be available soonish.

PR #1612, opened 2026-02-22
Migrate BottomSheet to WindowMetrics API

meenbeese:
Replace deprecated Display.getMetrics() usage with WindowMetricsCalculator.computeCurrentWindowMetrics() to ensure correct behavior in multi-window mode, split-screen, rotation, and on foldables.

The previous nullable Activity safety check was also removed. The height calculation now uses requireActivity(), which is safe because it is executed inside setOnShowListener after the fragment is attached and the dialog is being shown. At that lifecycle point, the Fragment is guaranteed to have a non-null host Activity.

Tested on Pixel 9a API 36.

<table>
    <tr>
        <th>Before</th>
        <th>After</th>
    </tr>
    <tr>
        <td>
</td>
        <td>
</td>
    </tr>
</table>

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -156,6 +156,7 @@ dependencies {
     implementation(libs.androidx.biometric)
     implementation(libs.androidx.lifecycle.common)
     implementation(libs.androidx.lifecycle.process)
+    implementation(libs.androidx.window)
     implementation(libs.retrofit.converter)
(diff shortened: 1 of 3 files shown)


syphyr:
This works fine with Nougat.

bitmold:
This works great, tested on API 36 and with split screen mode. Thank you

PR #1618, opened 2026-02-25
Always copy the license file to fix ./gradlew assemble

syphyr:
Fix `gradlew assemble` from failing because the LICENSE file is not copied to assets first.
This PR makes gradle always copy the LICENSE file to the assets directory with both release and debug builds.
The task for checking for updated bridges will still only be for release builds.


syphyr:
Task ':app:mergeNightlyDebugAssets' uses this output of task ':app:copyLicenseToAssets' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.

syphyr:
Both `./gradlew assembleRelease` and `./gradlew assembleDebug` already work fine. This PR is just to fix `./gradlew assemble`.

PR #1619, opened 2026-02-25
Don't replace non-existent parent manifest declarations

syphyr:
Eliminates two Android lint warnings about orphaned tools:replace directives that were attempting to override non-existent parent manifest declarations. The actual functionality remains unchanged since the attributes are still declared normally.

Also, suppress some deprecated java calls.  Use `@Suppress` instead of `@SuppresLint`.


bitmold:
Thanks :)

PR #1621, opened 2026-02-25
Relocate import from the old toolbar

meenbeese:
A nice UX improvement that also replaces the legacy toolbar menu. Functionality is the same.

Tested on Pixel 9a API 36.

<table>
    <tr>
        <th>Before</th>
        <th>After</th>
    </tr>
    <tr>
        <td>
</td>
        <td>
</td>
    </tr>
</table>


bitmold:
I think that the menu is fine, and I'm a bit confused by this. The functionality appears to be different because now there is this huge icon on the screen. What happens on the client authorization screen when you have imported client auth configurations for onion services? Would go they go on top of this new icon? 

I'm not seeing the value in this right now, so closing.

meenbeese:
If there are imported configurations, then this icon does not appear at all.

bitmold:
That's incorrect then, it should always appear since you can import as many as you want

meenbeese:
> That's incorrect then, it should appear you can import as many as you want

Good catch, I reworked the UI slightly so that it always appears.

meenbeese:
<table>
    <tr>
        <th>Before</th>
        <th>After</th>
    </tr>
    <tr>
        <td>
</td>
        <td>
</td>
    </tr>
</table>

bitmold:
I'm not interested in merging this. The menu option is fine and this takes up more space on the screen.

PR #1629, opened 2026-03-06
Improved DNSTT support for challenged countries.

tladesignz:
Rework to auto-generate DNSTT bridge lines for a list of challenged countries using UDP DNS servers known to work in each country, using 2 known working DNSTT servers.

Since Tor, as it turns out, uses up to 32 SOCKS connections to PTs, no more than 30 bridge lines are returned. If more than 30 are available, a random selection of available bridge lines is returned.

2 are left as buffer, since Tor seemingly uses these SOCKS connections to bridges, and 32 seems like a total limit. (At least I once saw an error message which indicated something like that.)

This is a copy of what is implemented on iOS and macOS: 
https://github.com/tladesignz/IPtProxyUI-ios/blob/main/IPtProxyUI/Classes/Shared/BuiltInBridges.swift

This should increase the chances of getting a connection in said countries a lot.

Notes:

1. Bootstrapping over UDP DNSTT might take a long time. But when finished, should work quite ok. (For me Afghanistan works, somehow. Try that to test!)
2. For an unknown reason, "Connected" is shown way too early. IDK if this is a general problem since some time, or if it is specific to DNSTT. I would appreciate it very much, if you could investigate, @bitmold!

Thank you!


syphyr:
Is there currently a way to select `global` as the `countryCode`?  I was thinking that DNSTT could be opened up for everyone if `global` was the default for everyone not included in the list of `dnsCountries`. Something like this: https://github.com/syphyr/orbot/commit/1201d6466584ba4f51d0c5b40d2c6ff6c98d8cf4

app/src/main/java/org/torproject/android/service/circumvention/BuiltInBridges.kt
```diff
@@ -85,4 +109,95 @@ data class BuiltInBridges(
...
+     * (Like our base ``BuiltInBridges.dnsttBridges`` list does).
+     *
+     * The UDP servers only make sense to use in a heavily censored environment, where public DoH
+     * and DoT DNS servers are blocked.
+     *
+     * That's what these UDP lists are for. Insofar the `global` list is only there for the sake of
+     * completeness, not so much because it makes sense to use it.
+     *
+     *  Here's a list of publicly available DoH servers, in case you're unhappy with our choice:
+     *  https://github.com/curl/curl/wiki/DNS-over-HTTPS#publicly-available-servers
```
tladesignz: @syphyr It's a trap. I documented it here.

bitmold:
Wow I finally found the Connected bug. I sometimes noticed it but could never reproduce it, but debugging dnstt with configs for Afghanistan and Turkey made things slow enough for me.

The faulty code isn't in Orbot, it's from a commit in `tor-android` in 2019... 

https://github.com/guardianproject/tor-android/blob/master/tor-android-binary/src/main/java/org/torproject/jni/TorService.java#L269:

```java
 /**
     * Announce Tor is available for connections once the first circuit is complete
     */
    private final RawEventListener startedEventListener = (keyword, data) -> {
        if (TorService.STATUS_STARTING.equals(TorService.currentStatus)
                && TorControlCommands.EVENT_CIRCUIT_STATUS.equals(keyword)
                && data != null && !data.isEmpty()) {
            String[] tokenArray = data.split(" ");
            if (tokenArray.length > 1 && TorControlCommands.CIRC_EVENT_BUILT.equals(tokenArray[1])) {
                TorService.broadcastStatus(TorService.this, TorService.STATUS_ON);
            }
        }
    };
```


image

Working on a fix right now...

bitmold:
This works great, and I merged it. 

Though the Connected bug won't be fixed until https://github.com/guardianproject/tor-android/pull/191 gets merged into tor-android (cc @n8fr8 )

PR #1636, opened 2026-03-10
Small code cleanup in AppManager

meenbeese:


bitmold:
Sorry, this looks good, but i'm not going to merge this PR that just refactors some code, since this would require both a good deal of time spent testing on my while also assuming ownership of anything that this PR might've inadvertently changed.

PR #1655, opened 2026-04-02
Make Orbot AGP 10 Compliant 

### LINKED ISSUE #1617: Final Issues Which Will Break Builds in AGP 10
If you compile Orbot with every warning enabled, there are three gradle issues that will prevent Orbot from updating to the soon-to-be-released AGP 10 toolchain:

```bash
./gradlew assembleFullpermDebug --warning-mode-all
```

From the newly generated `build/reports/problems/problems-report.html ` there are three more issues pertaining to custom gradle tasks we invoke in `app/build.gradle.kts`: 
```
- [warn]  Invocation of Task.project at execution time has been deprecated. (3)
    - [warn]  Invocation of Task.project at execution time has been deprecated.
This will fail with an error in Gradle 10.
        - Solutions
[enum]  This API is incompatible with the configuration cache, which will become the only mode supported by Gradle in a future release.
        - Locations
- `:app:getVersionFromGit`
    - [warn]  Invocation of Task.project at execution time has been deprecated.
This will fail with an error in Gradle 10.
        - Solutions
[enum]  This API is incompatible with the configuration cache, which will become the only mode supported by Gradle in a future release.
        - Locations
- `:app:renameApkFiles`
    - [warn]  Invocation of Task.project at execution time has been deprecated.
This will fail with an error in Gradle 10.
        - Solutions
[enum]  This API is incompatible with the configuration cache, which will become the only mode supported by Gradle in a future release.
        - Locations
- `:app:renameApkFiles`
```

1. `getVersionFromGit` needs to become compatible with the soon to be mandated configuration cache 
2. and so does `renameApkFiles` 
3. `renameApkFiles` is invoked in a certain way at execution time, we need to figure out the right way to use this task in `app/build.gradle.kts`

bitmold:
The following gradle tasks in `app/build.gradle.kts` will prevent Orbot from being able to upgrade to AGP 10.0 which should be released in a few months:

- `:app:renameApkFiles`
- `:app:updateBuiltInBridges`

Today, gradle's [configuration cache](https://docs.gradle.org/current/userguide/configuration_cache.html) is an optional feature, but on AGP 10 it's going to be turned on by default and you can't disable it.

With the CC turned on, Orbot's build script has to be deterministic. These changes makes it so that we can keep building with new Android Studio releases, and also has the side effect of **significantly** improving compile speeds after an initial build. 

Renaming the APKs was pretty straightforward with a new API... However, refactoring the bridge update task to work under the CC was a good deal harder. There's now a build script `buildSrc/src/main/kotlin/UpdateBridgeConfig.kt` which is where we now hit tor's MOAT API and get dnstt config when assembling release builds of Orbot. The class is an abstract gradle `DefaultTask` and properties that we use to determine if the config needs to be regenerated (based around `git status`...) are now injected in the correct, deterministic, stateful way when gradle executes. 

I made sure that:
- all debug builds do not trigger the bridge updating logic ie (`./gradlew assembleDebug`, `./gradlew assembleFullpermDebug` and `./gradlew assembleNightlyDebug`)
- all release builds do trigger the bridge logic ie (`./gradlew assembleRelease`, `./gradlew assembleFullpermRelease` and `./gradlew asssembleNightlyRelease`)
- that running `./gradlew assemble` (which builds everything `debug` and `release) behaves as you'd expect (this is achieved by _always_ doing the debug builds _after_ the script evaluation for the bridge config is finished for the release builds...) 
- and that the resultant APKs all have the correct filename pattern that GP has been using for releases: 

## APK filename patterns:
filenames

## When you try to make a release build without updating the bridge config after 24hrs
Screenshot From 2026-04-02 05-41-45

## Once things are updated in a release build:
Screenshot From 2026-04-02 05-41-11

Really can't stress how much faster everything's building now. Wish this had been done sooner...

This closes #1617

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -85,7 +85,7 @@ configure<ApplicationExtension> {
             proguardFiles(
                 getDefaultProguardFile("proguard-android-optimize.txt"), "proguard-rules.txt"
             )
-//            signingConfig = signingConfigs.getByName("release")
+            signingConfig = signingConfigs.getByName("release")
         }
         getByName("debug") {
             isDebuggable = true
@@ -146,10 +146,8 @@ val updateBuiltinBridges = tasks.register<UpdateBridgeConfig>("updateBuiltinBrid
 
 }
 
-
 androidComponents {
     onVariants { variant ->
-// Rename APKs (Replaces removed archivesBaseName)
         base {
             archivesName.set("Orbot-${android.defaultConfig.versionName}")
         }
@@ -159,7 +157,7 @@ androidComponents {
             }
             variant.sources.assets?.addGeneratedSourceDirectory(
                 updateBuiltinBridges,
-                UpdateBridgeConfig::assetsDir // Points to the @get:OutputDirectory in your class
+                UpdateBridgeConfig::assetsDir
             )
         }
     }
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -1,6 +1,5 @@
 import com.android.build.api.dsl.ApplicationExtension
 import java.io.FileInputStream
-import java.net.URI
 import java.util.Date
 import java.util.Properties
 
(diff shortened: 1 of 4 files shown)


tladesignz (approved):
Looks good to me. Smoke testing looks good, too.

tladesignz:
Please merge yourself, whenever you see fit.

PR #1657, opened 2026-04-02
Add preference to hide battery optimization dialog

syphyr:
Once the user selects "Connect and Turn Off This Warning", there is no way to enable it back again without reinstalling Orbot.  This preference allows the user  enable the battery optimization check again without having to reinstall Orbot.


bitmold:
Power users can still enable battery optimization from the settings even if they've opted to dismiss the dialog

(this is available whether or not you are in power user mode).

So, I don't think it's worthwhile to introduce an additional setting (that has to be localized into other languages) for *all* users that exists to dismiss a dialog.

PR #1658, opened 2026-04-04
Make an exit button

syphyr:
Create an actual exit button with animations.  Closing the application from the title bar is not typical UI behavior.

 
Screenshot_20260404-233004


bitmold:
We're not interested in changing this UI, the exit from the main connect screen works fine

PR #1666, opened 2026-04-19
Run OrbotService and TorService in single process

syphyr:
When using power user mode and not setting battery optimization exemption (may drain the battery), this helps Orbot from being killed by Android for the following reasons:

1. Reduces memory footprint (one process instead of two)
2. Makes the service appear as a single, larger component to the OS
3. Eliminates the overhead of inter-process communication

Android's process killer considers:
- Total memory usage
- Number of active processes
- Process importance (foreground vs background)

By combining:
- You halve the process count
- The unified service can more easily maintain foreground priority
- The OS sees one "important" service rather than two separate ones

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -127,18 +127,12 @@
             android:exported="true"
             android:foregroundServiceType="systemExempted"
             android:permission="android.permission.BIND_VPN_SERVICE"
-            android:process=":tor"
             android:stopWithTask="false"
             tools:ignore="VpnServicePolicy">
             <intent-filter>
                 <action android:name="android.net.VpnService" />
             </intent-filter>
         </service>
-
-        <service
-            android:name="org.torproject.jni.TorService"
-            android:process=":tor"
-            tools:node="merge" />
 
         <service
             android:name=".ui.quicksettings.LaunchOrbotTileService"


syphyr:
Consider the following test case:

1) Start Orbot in power user mode.
2) Configure Tor Browser to use Orbot as an external Tor source in about:config with socks address and port. Note: Regular Firefox can also be used to test this, but Tor Browser is probably better because it is setup not to leak connections.
    a) chrome://geckoview/content/config.xhtml
    b) network.proxy.socks  127.0.0.1
    c) network.proxy.socks_port  9050
    d) extensions.torlauncher.start_tor  false
    e) extensions.torlauncher.prompt_at_startup  false
3) Browse to a website that opens many connections, like cnn.com or imdb.com (make sure connections are not blocked with noscript).
4) Orbot will get killed because of too many simultaneous connections.

Orbot no longer gets killed by Android when OrbotService and TorService run in a single process in this test case. Maybe because of the change in inter-process communication?

bitmold:
@n8fr8 @tladesignz and I discussed this a while ago. 

Unless I'm mistaken, this PR is to return the two processes into one when Orbot is in power user mode. 

But we have had, and publicly discussed, some very good reasons to isolate the two processes. Mainly being that a crash in tor, which often is beyond our control, doesn't take down the entire app. 

I'll reopen this if I'm incorrect about something, but closing because:

- it is generally beneficial for the vast majority of users to have these two processes as separate

- power user mode is a niche feature, we primarily focus on the VPNService side of things and the vast majority of our users never use power user mode. On top of this, an extremely significant number of Orbot users are non-technical, use the app because they live under censorship - they *need* to, and do not speak English so they instead use partially translated versions of the app without any smooth access to the majority English language orbot/tor resources online

- we *still* gve users in power user mode a warning dialog describing the exact problem you outlined above and make it easy for them to disable battery optimizations. that's the compromise for power user mode users. people in your situation basically need to do this...




I think you either need to disable battery optimizations or just apply this change yourself since you already custom build the app with your own patches. I see how this is valuable to you but it just doesn't belong in the main codebase. 

If you ever make other discoveries in this area though, I would be very open to entertaining: 
- rigorous + serious evidence that shows having two separate processes is no longer beneficial to the broader user base. I highly doubt this, but we could revisit things if this could be proven 

- a different change to fix this problem you're having with power user mode and enabled battery optimizations, but one that doesn't negatively impact other users

Issue #1509, opened 2025-12-09
Kindness Mode UX Rework


cstiens:
I've documented my initial thoughts for the UX flow below. At the moment, I'm not settled yet on any of the UX copy or language we use to present this concept to users, but we'll continue to work on that.

Basic UX Flow to Nudge Users to Upgrade:
- From the main Kindness Mode view, indicate if their proxy is level 1 (restricted) or 2 (unrestricted). 
- If it's restricted, we give the user the option to upgrade their proxy.
- When the user taps 'upgrade' we show a modal that gives basic information about upgrading. Something along the lines of: "Your proxy can be more powerful when kindness mode is running on wifi. To upgrade, enable UPnP on your wifi router." The goal is to tell them why it matters and what it takes to do it. From this modal, the user can tap to 'Learn More' which would take them to a help article (hosted on Orbot's website) with steps and risks/considerations. This help article can also contain more advanced information about setting specific ports, etc.
- The status of a user's proxy would change from restricted to unrestricted once the app knows which one it is. I imagine we would need to write logic for when to check for that. From a UX POV, we want to give the user that feedback. So, if they have taken the steps to enable UPNP, they see that it's worked.

tladesignz:
https://gitlab.com/guardianproject/tormobile/orbot-design/-/work_items/1

https://www.figma.com/design/rM9ELHipCf0zdsu8NunQ5o/Orbot-UX-Improvements--Q3-Q4-2025-?node-id=2208-3351

bitmold:
Closing this, as we finished this work. However there are some new UI considerations I'll open on another ticket pertaining to getting this feature working on Android 17....

Issue #1226, opened 2024-12-19
[Question] Does Using Kindness in Proxy Mode while having a different VPN always on create risks to users in anti-VPN countries?


amd1890:
**Describe Your Question**
I am connecting to a VPN always and I am using Orbot in proxy mode for different Apps. I would like to be able to have kindness mode turned on.

I am concerned about this combination. The App doesn't stop me from doing this, but if I am connecting to a data center IP, it may be a known VPN IP address and it may have a high fraud score.

I am not sure if Orbot connects to Tor on top of my VPN. Since my VPN is set to always and deny any non-VPN connections, it should be connecting through my VPN.

I am just worried about kindness somehow leading to some user having a VPN address as the attempted Snowflake IP and somehow this causing problems. This was probably already taken into account during programming, but I would not want someone to use Snowflake, connect to a data center IP through me prior to getting to Tor, and then wind up in trouble.

**Is Your Question Related?**
I don't think so.

**Additional Context**
Many of the data center IPs I connect to are sometimes blacklisted with high fraud scores. If someone is in a hostile country that doesn't have a firewall (but people can get in trouble later), this seems like a risk unless Orbot is programmed to avoid the VPN. If Orbot can avoid the VPN in proxy mode, this seem like a bug in the OS, because the OS is designed to specifically not allow this in certain modes, which I have on.

n8fr8:
It is a somewhat unique combination of configuration that you are attempting.

If Orbot is the VPN, when Kindness mode is on, that happens outside of the Tor network exit node VPN IP space. We can control it in that case.

With a separate VPN running, we don't have any control. It may be possible for Orbot to detect that, and explain why it may not be a good idea to use Kindness mode in that case.

Otherwise, the Snowflake Broker is constantly evaluating the pool of Snowflake Proxy IPs it has for quality and throughput. I don't think there is any harm for you offering up your IP in that way, through the VPN, but it may be of less quality and use than a Snowflake Proxy that is NOT on a VPN.

amd1890:
It may be a decent idea to either display a message or disable kindness mode for VPN users when Orbot is in proxy mode.

[...]

PR #1670, opened 2026-04-22
Kindness Mode UX Update

tladesignz:
Fixes issue #1509.


tladesignz:
@bitmold, it's about time you have a look at this.

Since you know the Tor-controlling part way better than me: What would you consider the best way to implement the test?

Additionally, I would be happy for any testing against non-standard phones and tablets.

Thank you!

syphyr:
please rebase

bitmold:
Awesome, I'll take a look at it tonight + tomorrow morning.

syphyr:
Something is strange with the latest merge. This commit https://github.com/tladesignz/orbot/commit/b070bf1bb81db191332b38cfa9eac4bc152e8e7c changes `String` to `String?`, but then the following change to `override fun natTypeUpdated` in `SnowflakeProxyWrapper.kt` is 6bfcb7e6e98d9ff36ec76a940d7bd52f8c57f5f7 which shows  `String` not `String?`, but this does not match HEAD.

Also, `kindness_mode_disclaimer_bridge` should be removed from app/src/main/res/values-tr/strings.xml

```
diff --git a/app/src/main/java/org/torproject/android/ui/kindness/SnowflakeProxyWrapper.kt b/app/src/main/java/org/torproject/android/ui/kindness/SnowflakeProxyWrapper.kt
index eefd62d14..6ae9cc305 100644
--- a/app/src/main/java/org/torproject/android/ui/kindness/SnowflakeProxyWrapper.kt
+++ b/app/src/main/java/org/torproject/android/ui/kindness/SnowflakeProxyWrapper.kt
@@ -89,8 +89,8 @@ class SnowflakeProxyWrapper(private val service: SnowflakeProxyService) {
                         // Ignored.
                     }
 
-                    override fun natTypeUpdated(natType: String?) {
-                        // TODO feature added in IPtProxy 5.4.1
+                    override fun natTypeUpdated(natType: String) {
+                        service.updateNatType(natType)
                     }
                 }
 
diff --git a/app/src/main/res/values-tr/strings.xml b/app/src/main/res/values-tr/strings.xml
index 695dc397c..bd94cd87a 100644
--- a/app/src/main/res/values-tr/strings.xml
+++ b/app/src/main/res/values-tr/strings.xml
@@ -277,7 +277,6 @@
     <string name="kindness_never_had_a_direct_connection">Paylaşım Modunu kullanabilmek için, en az bir kez Tor ağına doğrudan (herhangi bir köprü kullanmadan) bağlanmanız gerekir. Bağlantı kurduktan sonra, lütfen buraya geri dönerek Paylaşım Modunu etkinleştirin. Bu doğrudan bağlantı testini tamamladıktan sonra, Tor ağına bağlı olsanız da olmasanız da Paylaşım Modunu dilediğiniz gibi kullanabilirsiniz.</string>
     <string name="kindness_mode_cant_run_with_bridge">Tor ağına bağlanmak için bir köprü kullanırken Paylaşım Modu çalıştırılamaz.</string>
     <string name="kindness_mode_cant_run_in_your_country">Sansür nedeniyle, ülkenizde diğer kişilerin Tor ağına bağlanmasına yardımcı olmak için kullanılan Paylaşım Modu şu anda devre dışıdır.</string>
-    <string name="kindness_mode_disclaimer_bridge">Sansür uygulanan bir bölgedeyseniz veya Tor ağına bağlanmak için bir köprü kullanmanız gerekiyorsa, Paylaşım Modunu kullanmayın.</string>
     <string name="battery_optimization_title">Pil İyileştirmelerini Devre Dışı Bırak</string>
     <string name="battery_optimization_pref_msg">Aygıtlarının Orbot uygulamasını arka planda çalışırken erken bir şekilde kapattığından şüphelenen kullanıcılar için ayar.</string>
     <string name="battery_optimization_dialog_msg">Bazı Android aygıtlar, Orbot arka planda çalışırken uygulamayı kapatabilir. Orbot\'un bağlantısının kesilmemesini ve çalışmaya devam etmesini isteyen çoğu kullanıcı için, sistem ayarlarından Orbot\'u \"Her Zaman Açık\" VPN olarak ayarlamak yeterlidir. Bununla birlikte, gerekli görürseniz, aygıtınızın Orbot üzerinde herhangi bir pil iyileştirmesi yapmasını engellemek için gerekli adımları atabilirsiniz.</string>
```

app/src/main/java/org/torproject/android/ui/kindness/SnowflakeProxyWrapper.kt
```diff
@@ -89,8 +89,8 @@ class SnowflakeProxyWrapper(private val service: SnowflakeProxyService) {
                         // Ignored.
                     }
 
-                    override fun natTypeUpdated(natType: String) {
-                        service.updateNatType(natType)
+                    override fun natTypeUpdated(natType: String?) {
+                        // TODO feature added in IPtProxy 5.4.1
```
tladesignz: Seems, you broke my changes.
I *deliberately* didn't merge main just yet, because this is all still work in progress.

bitmold:
Noting this regression here, though I understand that this dialog that checks if a direct tor connection has been established is being replaced in some way by this testing dialog.

When you click activate on a cold install, the dialog to ensure that you have at least historically have had a direct tor connection appears. At this moment snowflake proxy is running and there's no way to turn it off.


If I go ahead and connect to tor, then go to activate kindness mode, the bug still exists. The testing dialog appears, but kindness mode is already running. 

image

image

bitmold:
On small screens, the headings get cramped 
image

bitmold:
The connected UI looks great, but unfortunately it breaks in landscape 

image

And this is a bit nitpicky, but the labels aren't lining up with these grey panels 
Screenshot From 2026-04-29 18-30-45

Also @cstiens it just dawned on me that on the kindness screen rectangular grey panels are used, whereas on the "more" screen to the right rounded panels are used. Personally, I think the slightly rounded ones look much nicer.

tladesignz:
> Noting this regression here, though I understand that this dialog that checks if a direct tor connection has been established is being replaced in some way by this testing dialog.
> 
> When you click activate on a cold install, the dialog to ensure that you have at least historically have had a direct tor connection appears. At this moment snowflake proxy is running and there's no way to turn it off.
> 
> If I go ahead and connect to tor, then go to activate kindness mode, the bug still exists. The testing dialog appears, but kindness mode is already running.

Good catch! I removed the legacy test.

tladesignz:
- Fixed the header indent.
- Added rounded cards as suggested.
- Added first test code and a mock test to be able to continue development.
- Added final behaviour to show initial scene only, when no valid test result. (older than 24 hours.)

Found another problem:

We need to bind to the `SnowflakeProxyService` to receive the evaluation of the proxy quality.

However, doing so unintentionally starts a SnowflakeProxy. I'm afraid, `SnowflakeProxyService` is in for a redesign. @bitmold, what do you think?

bitmold:
> We need to bind to the SnowflakeProxyService to receive the evaluation of the proxy quality.
> However, doing so unintentionally starts a SnowflakeProxy. I'm afraid, SnowflakeProxyService is in for a redesign. @bitmold, what do you think?

This is true, it needs to be modified, however I don't think a thorough redesign is needed.

`SnowflakeProxyService` is capable of running as a foreground service without actually running the snowflake proxy in IPtProxy. For instance, if the Service is configured to run on WiFi only and/or if the device is connected to a power source, the `Service` is capable of calling `stopSnowflakeProxy(String?)` to turn off the proxying. 

When the Service is created in `onCreate()` it calls `initNetworkCallbacks()` which immediately starts proxying if the network conditions are met. In this function you could add preconditions to ensure that the correct proxy evaluations are met before invoking `startSnowflakeProxy(String?)`

bitmold:
I'll provide more feedback as I can the next few days, but something to keep in mind, I know you're removing a lot of Strings here. I just merged more translations into the `master` branch, could you please merge master into here and remove whatever strings need to be removed for this PR. thanks. this is coming along really nice :+)

bitmold:
It's hard to see in the first screenshot, but the labels on the configuration bottom sheet were misaligned. The view was also broken in landscape, which I fixed in this tiny xml commit I pushed right now.

image
image

Is there a reason why this tiny bottom sheet needs to be so big? Perhaps it could be made smaller. 

```kotlin
    override fun getHeightRatio(): Float = 1/2f
```

tladesignz:
> Is there a reason why this tiny bottom sheet needs to be so big? Perhaps it could be made smaller.
> 
> ```kotlin
>     override fun getHeightRatio(): Float = 1/2f
> ```

It's too big on **typical** screens in portrait mode, and too small in landscape. Defining the size as a ratio of the main screen is an approach which will not work.

Is there a way to define the height of the sheet so it contains all content, but not higher than the 4/5th of the screen?

[OrbotBottomSheetDialogFragment](https://github.com/guardianproject/orbot-android/blob/master/app/src/main/java/org/torproject/android/ui/OrbotBottomSheetDialogFragment.kt) needs improvement.

tladesignz:
Oh, well. Gemini to the rescue!

bitmold:
With this commit https://github.com/guardianproject/orbot-android/pull/1670/changes/13c350f3f08ed831ad4ea9a2c9644edb70e1540a I obtained a reference to `OrbotActivity`'s view model. All fragments, including the new testing dialog, run within that Activity. This object lets us know tor's current state and is updated from `OrbotService` on connect/disconnect. 

I replaced the check of `Prefs.useVpn()` with a check to see if this object is currently set to `ConnectUiState.On`. This is becuase that check would fail if a user was directly connected to tor but was using "power user mode" (using the SOCKS proxy instead of the system VPN)

bitmold:
Now that we can read the Activity's UI state, my strategy for checking if a direct tor connection is possible will be the following:

1. if we currently have `ConnectUiState.On` or `ConnectUiState.Starting` then turn off `OrbotService` and wait until `ConnectUiState.Off`. Make sure that the user is aware their current tor connection will be temporarily shut off and obtain their consent before proceeding. 
2. Spin up and start `tor-android`'s `TorService` with the default configuration. Wait 90 seconds for a `STATUS_ON` event. Essentially, using the most minimal logic in the fragment that this [sample app](https://github.com/guardianproject/tor-android/blob/master/sampletorapp/src/main/kotlin/org/torproject/android/sample/MainActivity.kt) uses. 
3. Proceed with that success or failure with regards to the kindness mode flow.
4. If you had to stop `OrbotService` earlier, start it back up and indicate that this is happening again to the user.

bitmold:
Ok, here's a rundown if the testing dialog determines it needs to perform a connection test 

1. If `OrbotService` is running, stop it. (if the user is using at tor this point, they have configured bridges/proxies)   
2. Start the new `TestTorForSnowflakeProxyService` which ignores every setting the user may have configured, it:
  - Runs with the simplest torrc possible 
  - Doesn't do anything with the VPN service 
  - isn't a foreground service, it runs within context of the app being opened 
3. Set a timer for 90 Seconds (that's running in parallel to this new tor connection service)   
4. Either:
  - a. you get a valid tor connection, set the correct flag, and unbind from the service 
  - b. the timer runs out, do what's needed to clean things up 

I've tested this and have had positive results with:
- a fast timer of 10 seconds, sometimes tor would connect in time, other times it wouldn't 
- an active obfs4 connection to tor, the flow handles connection and reconnection fine 
- tor being off, same results as above  
- dismissing the dialog before the test completes 

I disabled device rotation between the `onCreate()` and the `onDestroy()` of `TestingDialogFragment`. The fragment opens in whatever orientation the app was in when the dialog was launched. Supporting fragment rotation for this dialog is a lot of additional work, since we have to safely keep tabs on a rather ephemeral `ServiceConnection`.

I think it's important to show on the screen before the user opens the dialog, that in order to do this test a few things will happen: 
- they're active tor connection, if they have one, will be temporarily turned off ⬅️ users relying on a tor connection/having an active VPN connection, running an onion service, etc should know that it's going to be shutoff and to avoid the test if they don't want this to happen 
- that we will actually connect directly to tor ⬅️ users in areas with censorship should be informed that they are doing something their censor may not like

bitmold:
@cstiens @tladesignz just pasting all the links I shared on the call earlier here regarding not using snowflake proxy on a device connected to the Internet via a (non-Orbot) VPN:

- https://support.torproject.org/anti-censorship/running-snowflake/

> "f your internet access is not censored, you should consider installing the [Snowflake](https://snowflake.torproject.org/) extension to help users in censored networks. When you run Snowflake on your regular browser, you will proxy traffic between censored users and an entry node in the Tor network, and that's all.
> Due to censorship of VPN servers in some countries, we kindly ask you to not run a snowflake proxy while connected to a VPN.

- https://forum.torproject.org/t/running-a-snowflake-proxy-behind-a-vpn-consequences-for-tor-users/2047/4 From Cecylia 
> If all of your outgoing and incoming traffic is sent through a VPN then it is true that connecting clients will be connecting to your proxy through an IP address in that VPN’s address space. Do you know whether your routing setup makes an exception for P2P traffic?

> In general, we would recommend not running a proxy in this kind of network. The IP address space is more likely to be blocked, and the extra hop may adversely affect the connection quality.

- https://github.com/guardianproject/orbot-android/issues/1226
> I am just worried about kindness somehow leading to some user having a VPN address as the attempted Snowflake IP and somehow this causing problems. This was probably already taken into account during programming, but I would not want someone to use Snowflake, connect to a data center IP through me prior to getting to Tor, and then wind up in trouble.

> Many of the data center IPs I connect to are sometimes blacklisted with high fraud scores. If someone is in a hostile country that doesn't have a firewall (but people can get in trouble later), this seems like a risk unless Orbot is programmed to avoid the VPN. If Orbot can avoid the VPN in proxy mode, this seem like a bug in the OS, because the OS is designed to specifically not allow this in certain modes, which I have on.

- https://forum.torproject.org/t/discourage-vpn-usage-when-running-a-snowflake-proxy/2408
- https://forum.torproject.org/t/suggestion-amendment-to-browser-snowflake-proxy-documentation/3450

bitmold:
⬆️ Regarding the above ⬆️

@cstiens @tladesignz @n8fr8 
In the app today, if you are using kindness mode and then set another app as the active android VPN app (say ProtonVPN), the Snowflake Proxy Service will stay running but stop running the actual proxying code. From a user's perspective, it should say "Kindness Mode is running" but it's just not doing anything and won't until the user disconnects the VPN 

We do have constraints on Kindness Mode, for if the user goes offline the notification will say "Connect to the Internet..." (kindness mode will still be on, but snowflake proxy isn't doing anything). Similarly, we display in the notification instructions to the user to get on Wi-Fi if they configured that but get off Wi-Fi, and for disconnecting from Power. In all these cases Kindness Mode is still **on**, it's just that internally snowflake proxy is turned off.

```xml
    <string name="kindness_mode_disabled_internet">Connect to the Internet to help people circumvent censorship</string>
    <string name="kindness_mode_disabled_wifi">Connect to Wi-Fi to help people circumvent censorship</string>
    <string name="kindness_mode_disabled_power">Connect your device to power to help people circumvent censorship</string>
```

At the least, we have the capabilities to do something like this for using a non Orbot-VPN. So I think we should display this to the user as well. Right now if a user runs MullvadVPN and turns on kindness mode they'll probably just be like "Idk why this thing isn't working, maybe it's broken". So in addition to some copy on the actual kindness mode screen, we should add a notification text like the one above. 

There is perhaps one more thing that I should investigate now too. As you all know, on Android a VPN app can be for the entire device's traffic, or on a per-app basis. I've tested kindness mode in the past and it definitely detects a full device VPN and starts+stops itself accordingly. 

**What I have never investigated** is what Orbot sees if ProtonVPN is the current VPN app set, but ProtonVPN is configured on a per-app basis, and Orbot _is not_ one of the apps going through  ProtonVPN. I'll test this soon, but **if in this scenario Orbot "sees" that it's not going through a VPN conneciton, then the copy should read something like "If you're using another VPN app you can use kindness mode if Orbot isn't one of the apps going through the VPN" _instead of_ "you can't use kindness mode if you are using another VPN app"**. Will report back here on this test....

tladesignz:
Ok, this looks fine to me so far.

## Code Improvements

One thing, though, @bitmold : Can you please tune down your logging? `Log.wtf()` (highest level) is certainly inappropriate for most of the newly added log lines. How about `Log.d()`?


## Concerns

Regarding your concerns about surprise Tor disconnections, VPN and *actual* status of the Snowflake Proxy code:

I suppose, we need the following:

- Stronger wording about *not* using this in censored environments and also *not* behind VPNs on the initial scene.
- Also clear wording that a test might happen which might shortly disconnect them a direct Tor connection is attempted. (Either on the initial scene or probably better in the testing scene, where users can return before it happens.)
- An additional status line on the active kindness mode scene, where the *actual* status of the Snowflake Proxy is shown:
  - Active
  - Waiting for power
  - Waiting for Wi-Fi
  - Stopped

Please add/improve the list of different states!

## Last missing piece

To communicate the quality of the Snowflake Proxy, [`KindnessFragment` needs to bind to `SnowflakeProxyService`](https://github.com/guardianproject/orbot-android/pull/1670/changes#diff-b1646e6ec1ee49b9e5dabf9395909e4b22614aace0041e124f2e216fc91ba3daR111).

However, the way `SnowflakeProxyService` is currently built, binding will automatically trigger a start, which is not good.

@bitmold, can you rework `SnowflakeProxyService` such, that it behaves properly with that changed requirement. Or should I take care of that?

bitmold:
> One thing, though, @bitmold : Can you please tune down your logging? Log.wtf() (highest level) is certainly inappropriate for most of the newly added log lines. How about Log.d()?

Re the logging, I had mentioned on the call that for the time being it's excessively logged and that we'd want to tune things down. Perhaps I should've made things clearer on the PR though. 

When viewing both sets of logs (the main app's and the logs from the tor process side-by-side with ./scripts/view_logs_tmux.sh` the intensity of the error logs makes it easy to filter out the ongoing work on the noise. Personal preference for me. Piping `adb logcat` into `grep` can also be helpful, but this quickly gets complicated if the logs you're interested in following have different tag arguments. Either way, feedback received, it'll be tuned way down before merging.

bitmold:
One quick thing: in this commit https://github.com/guardianproject/orbot-android/pull/1670/commits/e9bb460282968c7da9364e59d47f1c1be8876b85 I added small horizontal margins (`2dp` @ start/end) to the kindness mode header, the one on the screen where you have the option to activate kindness mode.

If you are using a small screen, it feels very cramped spanning the maximum possible space

Screenshot From 2026-05-20 14-07-04

bitmold:
@cstiens would appreciate your feedback on how to go about visually defining some of these [additional states Benjamin and](https://github.com/guardianproject/orbot-android/pull/1670#issuecomment-4496973281) [I were discussing above](https://github.com/guardianproject/orbot-android/pull/1670#issuecomment-4482612978)

> Stronger wording about not using this in censored environments and also not behind VPNs on the initial scene.

> Also clear wording that a test might happen which might shortly disconnect them a direct Tor connection is attempted. (Either on the initial scene or probably better in the testing scene, where users can return before it happens.)

> An additional status line on the active kindness mode scene, where the actual status of the Snowflake Proxy is shown:
> - Active
> - Waiting for power
> - Waiting for Wi-Fi
> - Stopped

@tladesignz Happy to start implementing this though

bitmold:
@tladesignz I have some ideas and will work on this. Will report back here in the coming days with something for you to review... 

> To communicate the quality of the Snowflake Proxy, [KindnessFragment needs to bind to SnowflakeProxyService](https://github.com/guardianproject/orbot-android/pull/1670/changes#diff-b1646e6ec1ee49b9e5dabf9395909e4b22614aace0041e124f2e216fc91ba3daR111).
However, the way SnowflakeProxyService is currently built, binding will automatically trigger a start, which is not good.
@bitmold, can you rework SnowflakeProxyService such, that it behaves properly with that changed requirement. Or should I take care of that?

bitmold:
@tladesignz I have some questions about NAT type quality, which as far as I'm aware only applies to Wi-Fi

```
        // TODO: We need this, to receive the proxy quality, but on the other hand, this
        //  will unintentionally start the SnowflakeProxyService, which we don't want, because
        //  it will automatically start SnowflakeProxy if there are no Wi-Fi/battery limits.
        //  -> hence a redesign of the SnowflakeProxyService seems to await us.
```

I understand that In order to get the NAT type, we need to start snowflake proxy, and that's why you are saying it's problematic that we can start `SnowflakeProxyService` since it will just start being a proxy when all we want is to know the NAT type (since in Orbot kindness mode is disabled). 

My understanding from reading `snowflake.go` is that in [Snowlfake `v2.313.1`
](https://gitlab.torproject.org/tpo/anti-censorship/pluggable-transports/snowflake/-/tree/v2.13.1?ref_type=tags) this code runs in `proxy/lib/snowflake.go` when you start snowflake from within IPtProxy:

```go
// some earlier initial config 
	err = sf.checkNATType(config, sf.NATProbeURL)
	if err != nil {
		// non-fatal error. Log it and continue
		log.Printf("Error checking NAT type: %s", err.Error())
		setCurrentNATType(NATUnknown)
	}
	sf.EventDispatcher.OnNewSnowflakeEvent(event.EventOnCurrentNATTypeDetermined{CurNATType: getCurrentNATType()})

	NatRetestTask := task.Periodic{
		Interval: sf.NATTypeMeasurementInterval,
		Execute: func() error {
			return sf.checkNATType(config, sf.NATProbeURL)
		},
		// Not setting OnError would shut down the periodic task on error by default.
		OnError: func(err error) {
			log.Printf("Periodic probetest failed: %s, retaining current NAT type: %s", err.Error(), getCurrentNATType())
		},
	}
	if sf.NATTypeMeasurementInterval != 0 {
		NatRetestTask.WaitThenStart()
		defer NatRetestTask.Close()
	}

// more config and eventually we get to the main loop...
```

So essentially, you get the NAT type once Snowflake Proxy starts, and then every `NATTypeMeasurementInterval` you get that information updated again (so long as snowflake proxy is running) 

Pls, correct me if I'm wrong in my understnading of how any of the above works...

But with this understanding a few questions and concerns come to mind: 

- `NATTypeMeasurementInterval` isn't set in Orbot. Is this intentional? Since this is 0, you never find out what the NAT type is again if snowflake is running for a long time (even throughout the not-too-long timespan of a phone traveling in a users pocket throughout a day...). So, is the design here with the deactivated screen to get the NAT type by starting snowflake proxy in the fragment's `onStart()` & then stopping it once we learn the NAT type, and thus preenting it from doing any actual proxying? And then this initial value that you just set from when you opened the deactivated screen for kindness mode is _then_ updated exactly one more time once the user actually enables kindness mode (thus starting starts snowflake proxy a second time, but of course keeping it running this time)

- Let's say somehow we are able to obtain the NAT type before the user toggles on kindness mode. It's restricted and i'm at my house and configure my router to be unrestricted. I then turn on kindness mode for the next three weeks, as far as I can tell I wouldn't get any more NAT type updates from snowflake until I turn off/on kindness mode again, correct? This seems like a bug/something tht's not accounted for in the design. If I'm unrestricted, and I go over to my friends house and they're restricted, I'd incorrectly think that I'm still on an unrestricted network even though I'm not. This can get even more confusing if my friend turned on kindness mode on their device, and their Orbot said "Restricted " while mine still said "Unrestricted" even though we're still on the same network. 

- If Orbot _should_ be setting `NATTypeMeasurementInterval` when starting snowflake proxy, what's a reasonable amount of time? If a user sets kindness mode to run when they're disconnected from power and whether or not they're on Wi-Fi, they really won't get at a new NAT type until they manually restart snowflake. Oudated NAT type info will follow them around everywhere... 

- in `SnowflakeProxyWrapper.kt`'s `stopProxy()` we release the mapped ports and then tell `SnowflakeProxyService` it's new NAT type is `IPtProxy.IPtProxy.NATUnknown`. This can happen when a user manually turns off kindness mode in the UI. But also, while kindness mode is running (ie KM's configured to be  Wi-Fi only and the user loses Wi-Fi...) In this case The Android `Service` stays running, but it turns off Snowflake Proxy until the restrictions are complied with. From a user's perspective, kindness mode stays on the entire time, but when the new restrictions are met we of course get the new NAT type again.

- Exactly how are we even supposed to get the NAT type when kindness mode is disabled? Would it go like this:
 1. the user opens the fragment
 2. snowflake proxy is started  until `NATTypeUpdated` is fires off in some async way 
 3. we learn the NAT type, update the disabled UI to show it the NAT type
 4. turn off snowflake proxy before it starts proxying stuff, all while making sure to shutdown this ephemeral invocation of snowflake proxy if the user moves away from the fragment? Or if somehow the user turns on kindness mode (which actually starts snowflake proxy, all while the NAT type is being determined).

That :arrow_up:  seems like a lot, and for what? All of these questions/uncertainties lead me to think that we **shouldn't display the "Proxy Quality: unknown/restricted/unrestricted" label when Kindness Mode is disabled.** It suggests that you _actualy_ have a proxy running when you don't, and may very well change once you start the proxy, and as far as I can tell you do actually have to start the proxy to learn this. _Why not hide it when it's disabled_, and then re-enable its visibility when kindness mode is turned on? 

Unless I'm missing something, it seems like `SnowflakeProxyService` shouldn't be redesigned. it runs **if and only if**  kindness mode is turned on, and while it runs it displays a notification that tells the user if snowflake proxy is running or not (depending on Wi-Fi/Power settings). Starting `SnowflakeProxyService`, while kindness mode is disabled means you give the user a notification bar icon and a persistent notification - for a short period of time - just to determine a NAT type. Even if you kill the service immediately right after learning the NAT type, the user sees a quickly appearing notification/top icon appear and quickly disappear, just from opening the kindness mode screen... Running foreground services is a big deal on Android, they should be launched only by intentional direct user action (such as clicking Start) instead of launching automatically launching without consent after a user opens a screen.

bitmold:
If, in `onStart()` we jiust want to show the Fragment's NAT type value when kindness mode is turned On, that's easy. `OrbotActivity` can have a `BroadcastReceiver` that we use when binding to `SnowflakeProxyService`. Whenever we get a new NAT type from IPtProxy, it sends an update to the Activity which stores it in its view model. The Fragment, can query this value whenever it's opened up without needing to directly communicate to the Service. 

Another, i had another idea that came to mind just now... _Every_ time we get a new network event, (gain Wi-Fi, revert to cellular, connect to a new Wi-Fi, etc) `SnowflakeProxyWrapper` can be modified to stop and then restart snowflake proxy. This causes the initial NAT Type evaluation logic that happens when snowflake proxy first starts to fire off on every Android network event. You'd get accurate + timely NAT type updates for whatever connection you're currently on as soon as you connect to it (provided that kindness mode is turned on, of course). To me it seems better than finding an optimal repeat interval for `NATTypeMeasurementInterval`. As stated above, we already do something like this if the user has a Wi-Fi/power restriction. The difference is that rather than disconnecting when you're on Wi-Fi only and drop to a cellular network, you instead _always disconnect on a new network_, and then either immediately restart snowflake proxy if conditions are met/absent, or otherwise do what it already does today and wait to start snowflake proxy back up @ some future point after the device rejoins a Wi-Fi network.

Alternatively, is there anyway `snowflake.go` could be patched/modified for to accept a signal that on-demand fires off its NAT evaluation instead of us starting/stopping it in Orbot. This seems  like the most elegant approach instead of my above idea ofkilling/relaunching `SnowflakeProxyService` on new network events, or using a polling interval.

tladesignz (commented):
Well, well, well. TBH, I'm not really fond of that approach to deny Kindness Mode when the user selected a certain country in an unrelated feature.

You're reusing information which was meant for a completely different feature. In the context of that feature, the country list makes sense, as for the countries there are specific DNS servers to use.

But said country list is not a definitive list of countries where censorship happens. These two things correlate, but are not necessarily the same.

Think of special connections: white SIM cards; Starlink; university research Wi-Fi

That's why I always pushed for the ultimate test: Try to run a straight Tor connection. That's the actual always true fact upon which we can decide if running a Snowflake Proxy makes sense or not.
How about this:
There is certainly the case, that for some people, it could theoretically be detrimental to get their Tor connection closed and a straight Tor connection established.
In the current implementation, this is somewhat surprising.
We certainly need some interstitial alert before actually starting the test.
That seems to me the right place to use the country information.
"We need to run a straight Tor connection as a test to see, if your current location is a good place for running a Snowflake Proxy. We will now disconnect your Tor VPN and run that test.

DO NOT PROCEED, if running an unobfuscated Tor connection right now could get you in trouble."
And, additionally or alternatively, when we see a country on our list:
"You selected XXX as your country. Running a Snowflake Proxy in that country most probably does not make sense. You cannot help others here. STOP RIGHT NOW."
I leave it to @cstiens  to improve my wording.

tladesignz:
@bitmold re NAT quality evaluation:

Wow, you quite overthought this a lot.

No, I don't want to start Snowflake Proxy *just* to get the NAT quality info.

I just want to get that info at all, *when* it is available. My first approach was to bind to the service and get the value communicated that way, but that turned out to be a bad idea, because that would actually start the Snowflake Proxy as the service is currently set up.

Your idea of rather using a local broadcast of course also works. AFAIR, that is discouraged though since recently and my Android Studio complains about it.

But, if you think to stay on local broadcasts is better than to redesign the service so it doesn't automatically start the Snowflake Proxy, I'm fine.

I like the idea of restarting the Snowflake Proxy on network changes.

And thanks for the hin with the `NATTypeMeasurementInterval`. I totally forgot about it and was actually waiting that something like that gets implemented. 🤣 I'll use that on iOS, as there it's more complicated to listen for network changes.

I still think that showing `NATUnknown` when the Snowflake Proxy connection stops is fine. I wouldn't want to show/hide whole elements all the time. That seems to bring a sort of restlessness into the UI.

bitmold:
> No, I don't want to start Snowflake Proxy just to get the NAT quality info.

>I just want to get that info at all, when it is available. My first approach was to bind to the service and get the value communicated that way, but that turned out to be a bad idea, because that would actually start the Snowflake Proxy as the service is currently set up.

@tladesignz my confusion is there's currently a view displaying the NAT type, introduced by this PR, when kindness mode is turned off. Why even show it? Since the _only_ way for the NAT type to be available is after starting snowflake proxy, correct? 

Or to put it another way, the only way to get the NAT type in `snowflake.go` is to start proxying, it tells you the NAT type when the proxy first starts, and from there, it optionally can do the test at fixed intervals.

app/src/main/java/org/torproject/android/ui/kindness/SnowflakeProxyWrapper.kt
bitmold: This isn't broken. Maybe on one commit over the course of the rebase it was, but if you look at the actual file at the end of this branch, your change is in tact.

We needed to merge the main branch in so I could test the tor connection test with obfs4, as previously on master there was a bug that broke obfs4

bitmold:
Curious on your thoughts for a reasonable `NATTypeMeasurementInterval` value. Is 5 minutes too much, too little?

app/src/main/java/org/torproject/android/ui/kindness/SnowflakeProxyWrapper.kt
tladesignz: Yeah, I know. Sorry for the confusion. Github reviews are just weird to do right.

tladesignz:
> @tladesignz my confusion is there's currently a view displaying the NAT type, introduced by this PR, when kindness mode is turned off. Why even show it? Since the _only_ way for the NAT type to be available is after starting snowflake proxy, correct?
> 
> Or to put it another way, the only way to get the NAT type in `snowflake.go` is to start proxying, it tells you the NAT type when the proxy first starts, and from there, it optionally can do the test at fixed intervals.

As I already stated above:

>I still think that showing NATUnknown when the Snowflake Proxy connection stops is fine. I wouldn't want to show/hide whole elements all the time. That seems to bring a sort of restlessness into the UI.

tladesignz:
> Curious on your thoughts for a reasonable `NATTypeMeasurementInterval` value. Is 5 minutes too much, too little?

For iOS/macOS I decided upon 1 hour. Since it's a foreground thing on iOS, I suppose user's are not going to move much while running the Snowflake Proxy.

Of course, that can be very different on Android. But 5 minutes still seems a little much.

app/src/main/java/org/torproject/android/ui/kindness/SnowflakeProxyWrapper.kt
bitmold: No worries, this one is definitely one of the more chaotic ones I've been involved in for a minute - lots of moving parts. i feel like we can both be a bit pedantic at times, but i know it's just because we want this done excellently. excited to wrap this one up ! 

Also realizing that your comment was from 3 weeks ago, but was just seeing it for the first time today. Perhaps a user error on my part, but I wouldn't be surprised if that annoying GH Start Review "feature" wasted our time yet again....

bitmold:
Since there's no way to simply ask `snowflake` to do the test for us on demand, I'm going to take 20 min and implement restarting it on network change, and then test it by biking to around to several nearby friends' front porches (I'm on their Wi-Fi networks). if this works, we'll go with it. If not, I might try something like `NATTypeMeasurementInterval` = 30 minutes 🤷‍♀️

For mac an hour seems perfect though

bitmold:
>That's why I always pushed for the ultimate test: Try to run a straight Tor connection. That's the actual always true fact upon which we can decide if running a Snowflake Proxy makes sense or not.
How about this:
There is certainly the case, that for some people, it could theoretically be detrimental to get their Tor connection closed and a straight Tor connection established.
In the current implementation, this is somewhat surprising.
We certainly need some interstitial alert before actually starting the test.
That seems to me the right place to use the country information.
"We need to run a straight Tor connection as a test to see, if your current location is a good place for running a Snowflake Proxy. We will now disconnect your Tor VPN and run that test.

I didn't mean to suggest that those countries were an exhaustive list of regions where you can't run snowflake. I made the change after it seemed like a user of the iOS app based in Turkey complained after failing the test. Of course, in all the countries not in that list, if you try to start kindness mode you have the opportunity to pass or fail the definitive test. 

The idea was to save users who want to just get online to talk to their families or read the news time by blocking off part of this app because even though there's been all this amazing UX work put into Orbot, the app still is still inevitably filled with a lot of confusing terminology and technical concepts that most users do not have a solid understanding of. 

I'd rather just block it in some places where it's almost always not going to work anytime soon, rather than say something to the effect of 

> "You selected XXX as your country. Running a Snowflake Proxy in that country most probably does not make sense. You cannot help others here. STOP RIGHT NOW."

I think many users would find this warning confusing and **scary**. To me, keeping things streamlined and uncomplicated for users who have chosen to specify that they're in a sensitive area **vastly outweighs** being able to on ramp the rare starlink user in say Afghanistan, for example. And even if If they're clandestinely using starlink against the violent morality laws of the Taliban, then it's safe to assume that they probably have bigger concerns than helping users connect to the tor network via Orbot's kindness mode. If I had to guess, the lion's share of people in that region who are able to pass the test and sustainably run kindness mode would probably be some less than tech-savvy people in proximity to the Supreme Leader that have Orbot installed and then randomly turned on kindness mode and are somewhat unknowingly dedicating  bandwidth to help people over the world connect to Tor. 

But say you _do_ have Starlink in a violently censored place and do use kindness mode - well for those, what 2 dozen users, having kindness mode on is only useful while they're on that starlink network, the minute they go 6-15ft (1.8-4.5m) away from the router and possibly join another network then they're running kindness mode uselessly, while perhaps also jeopardizing their personal safety. 

As for someone on a privileged university Wi-Fi the same drawback is the case. _Maybe_, this very negligible amount of users _could_ be snowflakes, but they're still going to become immediately unhelpful proxies the minute they leave their campus network. I would imagine this hypothetical researcher would probably be more drawn to the [standalone proxy](https://community.torproject.org/relay/setup/snowflake/standalone/) anyway.

As for white SIM card holders in Iran, my understanding was those [SIMs were issued to those aligned with IRGC](https://en.wikipedia.org/wiki/White_SIM_Card#Emergence_in_Iran). Since it seems like you have to be aligned with the state's project of censorship to get one of those SIM cards, I don't think those Orbot users would be interested in helping to run a  technology that's historically been used for circumvention in Iran. 

How about this:
- we remove some countries from the list I borrowed from DNSTT to only include ones where it's been and probably will continue to be a statistical improbability that someone could effectively use kindness mode in a sustained and helpful way.  Off the top of my head, Afghanistan and Iran come to mind as ones to keep, but I'm sure there are a few more.
- if a user selected that they are in a country on this list, instead of saying my proposed ` "Thank you so much for your interest in helping others connect to Tor. Unfortunately, Kindness Mode isn't available in your country at this time. We appreciate your generosity."` we take your suggestion `"You selected XXX as your country. Running a Snowflake Proxy in that country most probably does not make sense. You cannot help others here. STOP RIGHT NOW."` and modify it to ➡️ `"You selected XXX as your country. Running a Snowflake Proxy in that country most probably does not make sense. If you still want to try to see if your device is capable of being a Snowflake Proxy, please clear your selected country."` It's less scary that way, and it gives users who really want to do this the option to try the test just like everyone else, even though they'd likely fail it.

bitmold:
I 100% agree with you though that the testing dialog should have some kind of interstitial warning like you mentioned in that comment I just quoted

tladesignz:
> Since there's no way to simply ask `snowflake` to do the test for us on demand, I'm going to take 20 min and implement restarting it on network change, and then test it by biking to around to several nearby friends' front porches (I'm on their Wi-Fi networks). if this works, we'll go with it. If not, I might try something like `NATTypeMeasurementInterval` = 30 minutes 🤷‍♀️
> 
> For mac an hour seems perfect though

Love your physical dedication! 😁

tladesignz (commented):
You're going a long way so you don't have to rework the SnowflakeProxyService…

app/src/main/java/org/torproject/android/ui/kindness/TestingDialogFragment.kt
```diff
@@ -294,7 +296,7 @@ class TestingDialogFragment : DialogFragment() {
     companion object {
         const val KEY_RESULT = "kindness_test_result"
         const val TAG = "TestingFragment"
-        const val CONNECTION_TEST_TIMEOUT_MS = 90 * 1000L
+        const val CONNECTION_TEST_TIMEOUT = 90
```
tladesignz: Great! Now the unit of the value is absolutely unclear. Can you please either add a suffix or a comment?

bitmold:
@tladesignz this is a WIP. I will ping you when it's ready for review. There are a lot of tricky corner cases on this. It'll be done before Tuesdays meeting

app/src/main/java/org/torproject/android/ui/kindness/TestingDialogFragment.kt
bitmold: Resolving, this is no longer the case anymore

bitmold:
@tladesignz This is now ready for review, but in the process of writing this, I had to fix a few bugs on `master`, which is why there were so many merge commits...
- if the app kills `OrbotService` it might close before `TorService` (which is bound to `OrbotService`, *not* the app) might not unbind before `OrbotService` dies. This leak was handled in the UI of the app, but broke the connection test. 
- Annoyingly, deploying code in Android Studio triggered a boot event. Which made debugging this (and many other past things, brutal)

Some notes:
- `TestingDialogFragment` uses `CoroutineUtils.waitUntilStateFlowEquals` to turn off an active Tor connection. It blocks until Tor's off or there is a 90 second timeout. Once tor's off, it starts the new testing service. If the service connects, the test passes. If 90 seconds - `however long turning tor off took` elapses we fail if the quality check preference wasn't set. 
- I did not want to rework `SnowflakeProxyService` as you had suggested, It's designed to be run if and only if kindness mode is activated. It registers receivers for power and network events. It's already rather complex. It's much simpler for `KindnessFragment` to just subscribe to changes in the nat type value if and when the service updates it. 

Small UI observations:
- Should the green snowflake icon when kindness mode is deactivated be the same size as Orbi on the main connect screen? It looks odd having the two graphics be different sizes when switching back and forth between the two screens 
- When you pass the test, there's just a green check mark and a Success header. I think a small sentence explaining that the user is good to go ahead and use kindness mode would be very nice here.
- When the connection test is being performed, there is a spinning progress circle, followed by the dialog header beneath it. I think we should put the dialog header horizontlaly after the spinning progress circle (to the right in English, to the left in RTL languages)

app/src/main/java/org/torproject/android/service/circumvention/BuiltInBridges.kt
syphyr: This can be removed since dnsCountries no longer exists

tladesignz: Good catch, thanks!

tladesignz:
Looks good to me now. @bitmold, please merge when you feel fine about it!

syphyr:
I don't see how this can be even be tested without fixing this first https://github.com/guardianproject/orbot-android/commit/d6a85cdf2cec4674362681b195adbf4f5076fc58

syphyr:
There are also some strings that still need to be removed. 

```
diff --git a/app/src/main/res/values-bg/strings.xml b/app/src/main/res/values-bg/strings.xml
index 38f4d0d20..a7048cb0e 100644
--- a/app/src/main/res/values-bg/strings.xml
+++ b/app/src/main/res/values-bg/strings.xml
@@ -284,10 +284,6 @@
     <string name="country_clear">Премахване на държавата</string>
     <string name="battery_optimization_title">Изключване оптимизиране на батерията</string>
     <string name="battery_optimization_pref_msg">Настройка за потребителите, които подозират, че устройството им преждевременно изключва Orbot, докато приложението работи във фонов режим.</string>
-    <string name="kindness_mode_cant_start">Режим на доброта не може да бъде включен</string>
-    <string name="kindness_never_had_a_direct_connection">За да използвате Режим на доброта, трябва поне веднъж да сте се свързвали директно с Tor (без използване на мостове). След свързване се върнете тук, за да включите този режим. След като преминете този тест за директна връзка, можете да използвате Режим на доброта, независимо дали сте свързани с Tor или не.</string>
-    <string name="kindness_mode_cant_run_with_bridge">Не можете да използвате Режим на доброта докато сте свързани с Tor посредством мост.</string>
-    <string name="kindness_mode_cant_run_in_your_country">Режимът на доброта с цел подпомагане на други хора да се свържат с Tor в момента е изключен във вашата държава поради цензура.</string>
     <string name="kindness_mode_disclaimer_bridge">Не използвайте Режим на доброта в райони с цензура или когато трябва да използвате мост за свързване с Tor.</string>
     <string name="battery_optimizations_dialog_btn_disable">Настройки за оптимизиране на батерията</string>
     <string name="btn_connect_anyway">Всързване въпреки това</string>
diff --git a/app/src/main/res/values-ca/strings.xml b/app/src/main/res/values-ca/strings.xml
index 2f08eb2a4..6570a2117 100644
--- a/app/src/main/res/values-ca/strings.xml
+++ b/app/src/main/res/values-ca/strings.xml
@@ -282,13 +282,9 @@
     <string name="unable_to_start_other_vpn_app_error_msg">L\'aplicació %1$s s\'ha configurat per a ser una VPN «sempre activa». Desactiveu-la abans d\'iniciar l\'Orbot.</string>
     <string name="unable_to_start_unknown_reason_error_msg">El permís de la VPN s\'ha denegat, o altra aplicació VPN té activada la funció «sempre activa». Ho podeu descartar i reconnectar si heu tancat el diàleg de sol·licitud de permisos de VPN anterior per error. Els usuaris més tècnics poden fer servir una altra aplicació VPN i utilitzar el Tor com a servidor SOCKS activant el mode Power User de l\'Orbot.</string>
     <string name="open_vpn_settings">Obre la configuració VPN d\'Android</string>
-    <string name="kindness_mode_cant_start">No s\'ha pogut iniciar el mode altruista</string>
     <string name="pref_proxy_no_proxy">Cap servidor</string>
     <string name="pref_proxy_ss_title">URL de la configuració del ShadowSocks</string>
     <string name="pref_proxy_ss_summary">URL complet de la configuració del ShadowSocks que comenci per ss://</string>
-    <string name="kindness_never_had_a_direct_connection">Per a poder utilitzar el Mode altruista, cal que us hàgiu connectat al Tor directament (sense fer servir cap pont) almenys una vegada. En haver-vos connectat, torneu aquí per a activar el Mode altruista. En completar aquesta prova de connexió directa, podreu usar el Mode altruista tant si esteu connectat al Tor com si no.</string>
-    <string name="kindness_mode_cant_run_with_bridge">El Mode altruista no es pot executar mentre feu servir un pont per a connectar-vos a Tor.</string>
-    <string name="kindness_mode_cant_run_in_your_country">Actualment, no podeu utilitzar al vostre país el Mode altruista per a ajudar altres persones a connectar-se al Tor a causa de la censura.</string>
     <string name="kindness_mode_disclaimer_bridge">No feu servir el Mode altruista si esteu en una regió on hi hagi censura, o si us cal utilitzar un pont per a connectar-vos al Tor.</string>
     <string name="battery_optimization_title">Desactiva les optimitzacions de la bateria</string>
     <string name="battery_optimization_pref_msg">Opció per a aquells usuaris qui sospitin que el seu dispositiu estigui tancant l\'Orbot de manera prematura mentre s\'executa al rerefons.</string>
diff --git a/app/src/main/res/values-sq/strings.xml b/app/src/main/res/values-sq/strings.xml
index 97842d3da..9ed597b46 100644
--- a/app/src/main/res/values-sq/strings.xml
+++ b/app/src/main/res/values-sq/strings.xml
@@ -285,10 +285,6 @@
     <string name="unable_to_start_other_vpn_app_error_msg">Aplikacioni %1$s është formësuar të jetë një VPN “Përherë aktiv”. Ju lutemi, çaktivizojeni, para se të nisni Orbot-in.</string>
     <string name="unable_to_start_unknown_reason_error_msg">S’u dha leje për VPN, ose një tjetër aplikacion VPN ka të aktivizuar veçorinë “VPN përherë aktive”. Mund ta hidhni tej këtë sinjalizim dhe të rilidheni, nëse mbyllët gabimisht dialogun e mëparshëm Kërkesë për VPN. Përdoruesit e thelluar mund të përdorin një tjetër aplikacion VPN dhe të përdorin Tor-in si një Ndërmjetës SOCKS, duke aktivizuar Mënyrën “Përdorues i Thelluar” të Orbot-it.</string>
     <string name="open_vpn_settings">Hap Rregullime VPN-je Android</string>
-    <string name="kindness_mode_cant_start">S’arrihet të Vihet Në Punë Mënyra “I sjellshëm”</string>
-    <string name="kindness_never_had_a_direct_connection">Që të mund të përdorni Mënyrën “I sjellshëm”, duhet të jeni lidhur drejtpërdrejt me Tor-in (pa përdorur ndonjë urë) të paktën një herë. Pasi të jeni lidhur, rikthehuni këtu, që të aktivizoni Mënyrën “I sjellshëm”. Pas plotësimit të kësaj prove për lidhje të drejtpërdrejtë, jeni të lirë të përdorni Mënyrën “I sjellshëm”, pavarësisht në jeni lidhur a jo me Tor-in.</string>
-    <string name="kindness_mode_cant_run_with_bridge">Mënyra “I sjellshëm” s’mund të përdoret teksa jeni duke përdorur një urë për t’u lidhur me Tor-in.</string>
-    <string name="kindness_mode_cant_run_in_your_country">Përdorimi i Mënyrës “I sjellshëm”, për të ndihmuar persona të tjerë të lidhen me Tor-in, aktualisht është i çaktivizuar për vendin tuaj, për shkak censurimi.</string>
     <string name="kindness_mode_disclaimer_bridge">Mos e përdorni Mënyrën “I sjellshëm”, nëse gjendeni në një rajon ku censuroheni, ose nëse keni nevojë të përdorni një urë për t’u lidhur me Tor-in.</string>
     <string name="battery_optimization_title">Çaktivizo Optimizime Baterie</string>
     <string name="battery_optimization_pref_msg">Rregullim për disa përdorues që dyshojnë se pajisja e tyre po asgjëson para kohe Orbot-in teksa xhiron në prapaskenë.</string>
```

tladesignz:
```
-    <string name="kindness_mode_cant_start">S’arrihet të Vihet Në Punë Mënyra “I sjellshëm”</string>
-    <string name="kindness_never_had_a_direct_connection">Që të mund të përdorni Mënyrën “I sjellshëm”, duhet të jeni lidhur drejtpërdrejt me Tor-in (pa përdorur ndonjë urë) të paktën një herë. Pasi të jeni lidhur, rikthehuni këtu, që të aktivizoni Mënyrën “I sjellshëm”. Pas plotësimit të kësaj prove për lidhje të drejtpërdrejtë, jeni të lirë të përdorni Mënyrën “I sjellshëm”, pavarësisht në jeni lidhur a jo me Tor-in.</string>
-    <string name="kindness_mode_cant_run_with_bridge">Mënyra “I sjellshëm” s’mund të përdoret teksa jeni duke përdorur një urë për t’u lidhur me Tor-in.</string>
-    <string name="kindness_mode_cant_run_in_your_country">Përdorimi i Mënyrës “I sjellshëm”, për të ndihmuar persona të tjerë të lidhen me Tor-in, aktualisht është i çaktivizuar për vendin tuaj, për shkak censurimi.</string>
```

Sorry, but I cannot find these in the current version. Or do you mean others? The `Unused resources` check also doesn't show strings related to Kindness Mode.

syphyr:
I saw these unused strings when merging to master branch. I just assumed it was also on this one.

bitmold:
> I don't see how this can be even be tested without fixing this first [d6a85cd](https://github.com/guardianproject/orbot-android/commit/d6a85cdf2cec4674362681b195adbf4f5076fc58)

This doesn't block this feature from being tested at all.

syphyr:
Currently, it only works if you are using an emulator

bitmold:
Can you please explain specifically *what* is breaking on your physical device? I've tested this extensively on real android devices

And, for that matter, why reverting that commit fixes this.

syphyr:
> Can you please explain specifically *what* is breaking on your physical device? I've tested this extensively on real android devices And, for that matter, why reverting that commit fixes this.

When an emulator is used, the system uptime is reset each time the emulator is started, so if you try to connect within one minute, it will succeed.  But, on a real device that has been running for a while, this will always prevent Tor from connecting because the uptime is greater than one minute.

I've made a  revision to my proposed solution so that it keeps the uptime comparison for debugging within the emulator.  Please take a look at the latest version https://github.com/guardianproject/orbot-android/pull/1716

When I mention "connecting to Tor", I mean using the toggle in the main connect fragment.

bitmold:
When you press the deploy button in Android Studio, it triggers the on boot receiver. This is a known issue. Not just with emulators. 

The on boot receiver is separate from this work. It prevents tor from starting in the on boot receiver when the device has been booted for over 10 minutes. Unless your device takes 10 minutes to boot I am failing to see the problem.

syphyr:
> When you press the deploy button in Android Studio, it triggers the on boot receiver. This is a known issue. Not just with emulators.

By definition, any event that only happens in Android Studio is most likely testing with an emulator. Can you give an example when the on boot receiver is triggered unexpectedly without an emulator?

 >Unless your device takes 10 minutes to boot 

This is not measuring the time it takes to boot your device.  It is measuring current uptime, which can be much more than the time it takes to boot the device.  So, if you update Orbot or restart Orbot after the device has been running for more than 10 mins, then `sReceivedBoot` always stays false. When testing with an emulator, the emulator's uptime is constantly being reset to zero each time the emulator gets restarted.

 This code was originally put in there for pixel emulators only and a normal device should never trigger that return.  So, the test should be to only check for emulators before returning.  Checking for the goldfish hardware should cover all  cases of emulation, not just for emulated pixels with a fingerprint containing "sdk_gphone".

syphyr:
Perhaps power user mode and start Orbot on boot options must also be enabled to trigger this bug. I have not verified that yet.

PR #1675, opened 2026-04-26
Allow ipv6 addrs to be used in a bridge line's arguments 

### LINKED ISSUE #1672: [BUG] Custom WebTunnel bridges with IPv6 `addr` attributes are not recognized
**Describe the Bug**

The WebTunnel bridge supports the `addr` attribute. [The description](https://gitlab.torproject.org/tpo/anti-censorship/pluggable-transports/webtunnel#addr-string) and [the implementation](https://gitlab.torproject.org/tpo/anti-censorship/pluggable-transports/lyrebird/-/blob/fc105a03c0e0acc2479301c361c012ffed359c43/transports/webtunnel/client.go#L76-80). However, when one assigns an IPv6 address with a port to the attribute, orbot reports "Invalid bridge format" and one cannot continue to use the bridge.

**To Reproduce**
1. "Set Transport"
2. "Custom Bridges"
3. input this line: `webtunnel [2001:db8:6116:8c99:9999:9999:deed:beef]:443 183376C431B132375452BBA77777890123456789 url=https://addr.invalid/path addr=[2400:8d8d:1::1:dead:beef]:443` (this bridge is unreal but the format should be correct)

**Expected Behavior**
Orbot accepts this bridge line, and users can use it.

**Smartphone:**
- OS: Android 13
- Version: v17.9.2-RC-1

bitmold:
Closes #1672, previously webtunnel bridges with the `addr` field would incorrectly validate because the last line of this regex:

```kt
private val validBridgeRegex = Regex(
            """^"""
                    + """((meek_lite|obfs4|webtunnel|snowflake|dnstt)\s+)?""" // Optional bridge type: Currently supported PTs + vanilla bridges
                    + """((\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}|\[[\da-fA-F:]+])""" // Cheap IPv4 and IPv6 address test
                    + """(:\d{1,5})?)\s+""" // Optional port number, space after address
                    + """([\da-fA-F]{40})?""" // Optional bridge fingerprint
                    + """[ \t\f\w\-/+:=.,]*$""" // Optional bridge arguments (different per bridge type, subject to change)
        )
```

did not take into account the square bracket `[` chars `]` needed to write an ipv6 address. 

This is a cheap fix on top of the existing logic, and corrects the problem blocking  people from using webtunnel as reported in #1672. A nicer implementation would be to actually validate each bridge type to ensure has the correct optional bridge arguments.

### DIFF
--- a/app/src/main/java/org/torproject/android/ui/connect/CustomBridgeBottomSheet.kt
+++ b/app/src/main/java/org/torproject/android/ui/connect/CustomBridgeBottomSheet.kt
@@ -36,7 +36,7 @@ class CustomBridgeBottomSheet :
                     + """((\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}|\[[\da-fA-F:]+])""" // Cheap IPv4 and IPv6 address test
                     + """(:\d{1,5})?)\s+""" // Optional port number, space after address
                     + """([\da-fA-F]{40})?""" // Optional bridge fingerprint
-                    + """[ \t\f\w\-/+:=.,]*$""" // Optional bridge arguments (different per bridge type, subject to change)
+                    + """[ \t\f\w\-/+:=.,\[\]]*$""" // Optional bridge arguments (different per bridge type, subject to change)
         )
 
         fun isValidBridge(input: String): Boolean {


tladesignz:
In my point of view, this regex is completely superfluous. I don't use something like that on iOS/macOS. Because, what for? There is Tor and the respective PTs, which are the ultimate validator. They don't need protection.

The benefit this regex brings (a better error message), doesn't outweigh the many problems it already has introduced. (Namely, the many times, it turned out that the regex was incorrect.)

But, well. This was here before me, so I'm not taking this upon me to remove it.


Looks good so far. Merge when you see fit.

tladesignz (approved):

bitmold:
I can see it being helpful in preventing a usre who isn't technically inclined from pasting nonsense they think is correct in there, then connecting to tor, having it fail, then thinking the app is broken. But yeah it's been a PITA, hopefully this is the last we see from this regex for a long time

Issue #1683, opened 2026-05-06
[BUG] Invalid connection happening with built in bridges


n8fr8:
**Describe the Bug**
Choose snowflake or built-in obfs4 and try to "CONNECT"
but get invalid configuration

**To Reproduce**
1. Open Orbot
2. Choose a connection type - snowflake, built-in obfs4
3. Connect!

**Expected Behavior**
Connect works!

**What Custom Configuration Do You Use?**
None

**Smartphone (please complete the following information):**
- Device: Pixel 9
- OS: Latest Android 16?

**Crash Logs (Advanced)**
org.torproject.android.debug:787edfb2: onCancelled at PHASE_CLIENT_ALREADY_HIDDEN
setTopOnBackInvokedCallback (unwrapped): null
hide(ime())
org.torproject.android.debug:c9219e31: onCancelled at PHASE_CLIENT_ALREADY_HIDDEN
obfs4 connected
Updating torrc custom configuration…
torrc.custom=
RunAsDaemon 1
AvoidDiskWrites 1
SOCKSPort 9050 IPv6Traffic PreferIPv6 
SafeSocks 0
TestSocks 0
HTTPTunnelPort 8118 
ReducedConnectionPadding 1
CircuitPadding 1
ReducedCircuitPadding 1
TransPort auto 
DNSPort auto 
VirtualAddrNetwork 10.192.0.0/10
AutomapHostsOnResolve 1
DormantClientTimeout 10 minutes
DormantCanceledByStartup 1
DisableNetwork 0
Log debug syslog
SafeLogging 0
UseBridges 1
ClientTransportPlugin obfs4 socks5 127.0.0.1:39249
Bridge obfs4 209.148.46.65:443 74FAD13168806246602538555B5521A0383A1875 cert=ssH+9rP8dG2NLDN2XuFw63hIO/9MNNinLmxQDpVa+7kTOa9/m+tGWT1SmSYpQ9uTBGa6Hw iat-mode=0
Bridge obfs4 212.83.43.74:443 39562501228A4D5E27FCA4C0C81A01EE23AE3EE4 cert=PBwr+S8JTVZo6MPdHnkTwXJPILWADLqfMGoVvhZClMq/Urndyd42BwX9YFJHZnBB3H0XCw iat-mode=1
[...]

PR #1685, opened 2026-05-06
Ignore set Entry Nodes when a PT is also set

tladesignz:
This fixes issue #1683.


bitmold:
I wasn't familiar with this limitation, but this is a good fix. Will add something to the UI too...

PR #1700, opened 2026-05-20
disable shadowsocks on x86 devices

bitmold:
As discussed, this removes the shadowsocks binary on x86 (32 bit intel CPUs) which are effectively just android emulators with 64 bit Intel hosts emulating very old versions of Android. 

It removes the value from the UI, which should invalidate all ShadowSocks flows, but still, the wrapper and code around it aren't called if you're on a 32 bit Intel device.

I also changed all the String literals for `"ss"` into a constnat, in case anyone accidentally mistypes a character unknowingly, the program compiles, but something breaks.

### DIFF
--- a/app/src/main/java/org/torproject/android/service/tor/ShadowSocks.kt
+++ b/app/src/main/java/org/torproject/android/service/tor/ShadowSocks.kt
@@ -95,6 +95,6 @@ object ShadowSocks {
 
     @JvmStatic
     fun isShadowSocksSupported(): Boolean {
-        return Build.SUPPORTED_ABIS.firstOrNull()?.equals("x86") == true
+        return Build.SUPPORTED_ABIS.firstOrNull()?.equals("x86") != true
     }
 }
--- a/app/src/main/java/org/torproject/android/service/tor/ShadowSocks.kt
+++ b/app/src/main/java/org/torproject/android/service/tor/ShadowSocks.kt
@@ -46,8 +46,9 @@ object ShadowSocks {
      * as a library.
      */
     @JvmStatic
-    fun start(context: Context, serverUrl: String): String {
-        if (!isShadowSocksSupported()) return ""
+    fun start(context: Context, serverUrl: String): String? {
+        if (!isShadowSocksSupported()) return null
+
         stop()
 
         val file = File(context.applicationInfo.nativeLibraryDir, "libsslocal.so")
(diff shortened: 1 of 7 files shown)


tladesignz:
Well, I had some issues with the code, so I took the liberty to improve them myself. :-)

@bitmold, please review and merge if you're ok.

tladesignz (approved):

bitmold:
The edge case seems a bit redundant, no one has shadowsocks support since we haven't shipped it, and even if someone made a build of orbot from code this week, they were not using a 32 bit Intel android device

tladesignz:
> The edge case seems a bit redundant, no one has shadowsocks support since we haven't shipped it, and even if someone made a build of orbot from code this week, they were not using a 32 bit Intel android device

I personally was that edge case.

bitmold:
My bad, I thought you were that you were on one of those slick Apple ARM chips. Though in that case, you could just reinstall Orbot in your emulated device. Either way, I'm fine with the code being there - just seemed redundant when I did the implementation. 

Better safe than sorry.

app/src/main/java/org/torproject/android/service/tor/ShadowSocks.kt
```diff
@@ -95,6 +95,6 @@ object ShadowSocks {
 
     @JvmStatic
     fun isShadowSocksSupported(): Boolean {
-        return Build.SUPPORTED_ABIS.firstOrNull()?.equals("x86") == true
+        return Build.SUPPORTED_ABIS.firstOrNull()?.equals("x86") != true
```
bitmold, line 98: goodness, this was a stupid test.

bitmold:
Thanks for catching my mistake. This is ready to go and we get a smaller APK as a result. I can't wait until we can drop x86 completely. 

Perhaps you running into that issue of not setting a default was because of that UI testing bug I had committed by mistake, you needed to reset from shadowsocks `"ss"` to no outbound proxy `""` perhaps because you weren't using an x86 build of Android. 

Anyway it's ready to merge. In the future, I'd prefer to see more String constants for state variables such as `"ss"` just to prevent if someone mistakenly edits the String while debugging and wastes a lot of time figuring out why the project builds but something isn't working quite right. Thanks for looking this back over.

tladesignz:
Yeah. I'm absolutely for string constants. Don't know why I didn't do it properly this time. It was good that you caught that!

PR #1702, opened 2026-05-22
report to snowflake broker as "orbot-android" instead of "iptproxy"

bitmold:
just setting this here for now, so someone can login and merge it whenever a decision is reached...


bitmold:
don't merge until cecylia gives the green light

tladesignz (approved):

PR #1707, opened 2026-06-02
Strict Private DNS Leaks 

bitmold:
Private DNS on android has three states we can kind of vaguely reason about:

- Off
- `"automatic"` - the system will use DoT at its own discretion. seemingly prioritizes the user getting their websites loaded over adhering to universal DoT 
- `"hostname"` - the system says it will force DoT, even if this leads to a broken/poor UX. in this state, we can then obtain the user-specified TLS hostname 

## Querying This Stuff
We can basically do this on demand whenever we want. IE on the connect screen like @tladesignz had suggested, but could also ask for it anywhere, even in Notifications hypothetically. 

I'm 99.9% sure this is mostly what we want:
- we can learn if it's set
- possibly display some info about what the user set 
- explain to the user about say leaking a tor onion service URL 
- would be easy to ask for this information when you open a screen, or even when you are building a notification to show the user.  

```kt
    sealed class PrivateDns {
        object Off : PrivateDns()
        object Opportunistic : PrivateDns()
        data class Strict(val hostname: String) : PrivateDns()
        companion object {
            const val KEY_MODE = "private_dns_mode"
            const val KEY_HOSTNAME = "private_dns_specifier"
            const val HOSTNAME_UNKNOWN = ""
            const val MODE_OFF = "off"
            const val MODE_HOSTNAME = "hostname"
            const val MODE_AUTOMATIC = "automatic"
            fun isPrivateDnsSupported(): Boolean = Build.VERSION.SDK_INT >= Build.VERSION_CODES.P
        }
    }

    fun getPrivateDnsConfiguration(context: Context): PrivateDns {
        if (!PrivateDns.isPrivateDnsSupported()) return PrivateDns.Off
        val dnsMode =
            Settings.Secure.getString(context.contentResolver, PrivateDns.KEY_MODE)
                ?: PrivateDns.MODE_OFF
        when (dnsMode) {
            PrivateDns.MODE_OFF -> return PrivateDns.Off
            PrivateDns.MODE_AUTOMATIC -> return PrivateDns.Opportunistic
            PrivateDns.MODE_HOSTNAME -> {
                val hostname = Settings.Secure.getString(
                    context.contentResolver,
                    PrivateDns.KEY_HOSTNAME
                )
                    ?: PrivateDns.HOSTNAME_UNKNOWN
                return PrivateDns.Strict(hostname)
            }
        }
        return PrivateDns.Off
```

## `ConnectivityManager` and `ContentObserver`

But you can _also_ subscribe for network events in the app, which `SnowflakeProxyService` does. The connectivity lets us see when the user moves between networks. Whenever one of these network event fires off, we get the set of DNS resolvers the system has curated for the user at that moment. You don't get any info on if DNS is being resolved or plain text of if it's DoT style though. You also don't get notified which resolver is actually handling your query...

You can also subscribe with `ContentObsrever`s to get a callback fired whenever those values from above change. Seems like you can masochistically subscribe to both the variable change events, and the network change events, and perhaps make some guesses about what's going on.  

This seems complex, also battery draining. and it doesn't give us much value than the code I already wrote above did.


bitmold:
i unfortunately need to get on the original GitLab thread and won't be able to do that until later today :upside_down_face: 

I'm very curious why OnionMasq/TorVPN are working differently... 

I'm with cyBerta and concerned about how this rubs against Tor's anonymity promise. 

And at the bare minimum I agree with @tladesignz in the immediate term there's UX work to do to convey **_exactly_ what is happening to the user.** I bet there are a lot of users who have security fatigue, see this thing as "private" and would be upset to learn there's a leak.

I think some users presumably understand exactly what's going on and desire this. But I bet there are many other users who are in the dark, it's not especially clear...

But it gets even more complicated!! 
- When you have a `VPNService` on Android you can set it to be "[Always on](https://developer.android.com/reference/android/net/VpnService#isAlwaysOn())". And if you do that, you may _also_ set it t[o Block all connections outside of the VPN](https://developer.android.com/reference/android/net/VpnService#isLockdownEnabled()) 

- [Google says that on Android API 28, when Private DNS was released, they made a mistake and VPN's would rule over the Private DNS Setting ](https://developers.google.com/speed/public-dns/docs/using) They fixed this in API 29+ 

>Caution: Android 9 only! These Private DNS settings have no effect when you use a VPN like Nexus/Pixel [Wi-Fi Assistant](https://support.google.com/nexus/answer/6327199) or Google Fi [Enhanced Network](https://www.blog.google/products/project-fi/project-fis-enhanced-network-brings-faster-more-secure-connections/) VPNs, or third-party VPN or [DNS changer](https://play.google.com/store/search?q=DNS&c=apps) apps. Those features and apps override Private DNS and do not send [DNS-over-TLS](https://developers.google.com/speed/public-dns/docs/dns-over-tls) queries to Google Public DNS. Most DNS changers send cleartext queries (a few like [Intra](https://play.google.com/store/apps/details?id=app.intra) use other secure DNS protocols) and VPN apps may not secure queries beyond the VPN server. This is "fixed" in Android 10.

- Based on my cursory understanding I think Google actually has this extremely unclear system that would confuse just about everyone,wherein
  - If you have a `VPNService` like Orbot, with Private DNS in Strict Mode, then the DNS resolution is leaked outside of the VPN. You also have know way to verify this reliably.
  - But if you have a `VPNService` in Lockdown Mode and set Private DNS to strict mode, than the system seemingly quietly ignores the Private DNS Strict Mode. Incidentally the method to see if a `VPNService` is locked down is added in Android 29, the same version of Android where Private DNS started leaking...

FWIW Android apps can open the system setting screen where they can enable lockdown mode by starting this `Intent`

```kt
startActivity(Intent("android.net.vpn.SETTINGS")
                .addFlags(Intent.FLAG_ACTIVITY_NEW_TASK))
```

tladesignz:
> - explain to the user about say leaking a tor onion service URL

This typically **IS NOT TRUE**, as networking stacks are typically implemented to drop DNS queries for onion services early, thanks to Tor Project's lobbying and [RFC writing](https://datatracker.ietf.org/doc/html/rfc7686#section-2).

app/src/main/java/org/torproject/android/util/NetworkUtils.kt
```diff
@@ -21,6 +23,42 @@ object NetworkUtils {
         }
     }
 
+
+    sealed class PrivateDns {
```
tladesignz, line 68: This looks pretty wild. Isn't it possible to use enum parameters instead of static constants?

Also, a few empty lines would enhance readability.

tladesignz:
As already mentioned in the Gitlab issue:

I suggest adding a warning near the connect button about this. Nothing more.
However, that warning should go away automatically, when users change configuration. Hence, I think we need to hook up to the network change notifications.


To be clear, I think what Google is doing here is correct:

With normal ISPs as well as with normal VPN providers, it's a good idea to hide DNS queries from them. With such a "Secure VPN" setup, users split information about them between two (hopefully) unrelated parties which improves privacy.

This is an edge case in conjunction with Tor as "VPN", where Tor takes care of anonymization already, and does it better than said distribution of information.
We might also want to take steps (Alert box when tapping warning? Or when tapping connect button? ^^) to explain exactly that to the user. 

But @cstiens might have better advice here. After all, she is dedicated to reduction and de-complication.

syphyr:
I am currently using dnscrypt-proxy with Orbot to get a system wide private DNS. Dnscrypt has an option to convert all UDP DNS requests to TCP and query Tor over a socks proxy with a user/pass for isolation between apps. For my setup on cm-14.1, this also requires root in order to use iptables to forward all port 53 requests over to dnscrypt, but apparently newer versions of android do not require that?

Here is my source:
https://github.com/syphyr/dnscrypt_proxy_prebuilt

One benefit of using dnscrypt is that it allows the usage of a domain block list so that I can also use Tor DNS with ad blocking and malware blocking.

app/src/main/java/org/torproject/android/util/NetworkUtils.kt
bitmold: [This type of `sealed class` approach is an idiomatic thing to do in Kotlin and isn't wild.](https://stackoverflow.com/a/65226315). It's used elsewhere in Orbot and is very common in other projects.

It's akin to [using `trait`s in Rust](https://doc.rust-lang.org/book/ch10-02-traits.html#returning-types-that-implement-traits) and functions like an `enum`, but better. For an `enum` to have a parameter, _every_ type of the enum must _all_ have that parameter. Whereas here, we have the type `PrivateDns`, which can be `Off`, `Opportunistic` or `Strict`; and if it's `Strict`, it is accompanied with the property `hostname` which we know is a non-null `String` . If there wasn't this `hostname` property, or if the other possible types of `PrivateDns` also meaningfully had some notion of `hostname`, then`enum` would be the correct + simpler choice. 

So, if we used `enum`, we'd have to define `Off` and `Opportunistic` as _also_ having the property `hostname`, which would now be of type `String?`. We'd have to pass `null` into those two of the three possible `enum` values, and _every_ time where using `hostname` is relevant with `PrivateDns.Strict` we'd now also have to add `null` checking around `hostname`, even though we know it's non-null.

bitmold: It's explained further here in the official Kotlin docs: https://kotlinlang.org/docs/sealed-classes.html#use-case-scenarios

bitmold: I did cleanup the class though, and provided some comments

bitmold:
> I suggest adding a warning near the connect button about this. Nothing more. However, that warning should go away automatically, when users change configuration. Hence, I think we need to hook up to the network change notifications.

I don't think we need to hook into network configurations for the warning to appear/disappear when the user changes the settings. simply putting the logic to show/hide it in `onResume()` should suffice. Since if the user closes the app to change the settings, then reopens the app, `onResume()` is invoked and the text can be updated/removed if necessary

bitmold:
Whenever the user changes the setting, this updates `ConnectFragment` without the overhead/drawbacks of hooking into network change callbacks. (Which is typically done from an `Activity`/`Service` (which are `Context`s) 

```kotlin
override fun onResume() {
        super.onResume()
        updatePrivateDnsLabel()
}

private fun updatePrivateDnsLabel() {
        var labelVisibility = View.VISIBLE
        var labelText = ""
        var dialogMsg = ""
        var onLabelClick: () -> Unit = {
            AlertDialog.Builder(requireContext())
                .setTitle("Private DNS") // TODO
                .setMessage(dialogMsg)
                .show()
        }

        val privateDns = NetworkUtils.PrivateDns.getPrivateDnsConfiguration(requireContext())
        when (privateDns) {
            is NetworkUtils.PrivateDns.Off -> {
                labelVisibility = View.GONE
                onLabelClick = {}
            }

            is NetworkUtils.PrivateDns.Opportunistic -> {
                labelText = "Opportunistic Message"
                dialogMsg = "opportunistic explanation"
            }

            is NetworkUtils.PrivateDns.Strict -> {
                labelText = "Strict msg ${privateDns.hostname}"
                dialogMsg = "struct explanation ${privateDns.hostname}"
            }
        }

        binding.tvPrivateDnsStatus.apply {
            visibility = labelVisibility
            text = labelText
            setOnClickListener { onLabelClick() }
        }
    }
```

I'm going to wait until we can all talk to figure out specifically _what text_ to display to the user on the connect screen + the possible explain more dialog box. But the logic is complete, and can probably be tidied further once the exact implementation details are made clear.

app/src/main/java/org/torproject/android/util/NetworkUtils.kt
tladesignz: Guess I'm too spoiled from Swift… Thanks for the deep explanation!

n8fr8:
@abeluck let's talk about setting up our own DoH server for auditing/testing of what comes through

bitmold:
> @abeluck let's talk about setting up our own DoH server for auditing/testing of what comes through

Specifically using this DoH server to test an app using Orbot to resolve .onion domains.

app/src/main/java/org/torproject/android/util/NetworkUtils.kt
bitmold: AFAIK it's effectively doing this in Swift

```swift
enum PrivateDns { case Off() case Opportunistic() case Strict(String)
```

tladesignz: The casing is a bit off for Swift, but yes. I was hoping to get near such clearness, but it seems Kotlin doesn't let me.

bitmold: Yeah, i have never really written swift for anything more than a code tweak or a one off project 🤷‍♀️

SkewedZeppelin:
> This typically IS NOT TRUE

quick search via cs.android.com says only some external libraries handle this, the regular android stack doesn't.

- https://cs.android.com/android/platform/superproject/+/android-latest-release:prebuilts/go/linux-x86/src/net/dnsclient_unix.go;l=484
- https://cs.android.com/android/platform/superproject/+/android-latest-release:external/cronet/stable/net/base/registry_controlled_domains/effective_tld_names.dat;l=5070
- https://cs.android.com/android/platform/superproject/+/android-latest-release:external/curl/lib/hostip.c;l=830

> I suggest adding a warning near the connect button about this. Nothing more.

I agree with this. maybe a one-time in your face pop-up too.

but users should be able to use private dns if they want so they can have more trustworthy dns responses and dnssec support

abeluck:
@n8fr8 @bitmold I'm just catching up here.. reading the thread it seems like you need a DOH server setup for testing purposes with access to logs.... but I have some questions:

- Is this only for internal testing or does this need to receive production traffic from in-the-wild orbot users?
- What exactly do you need to see on the server side? Do you need to look at the logs?
- Who resolves the .onion domain? The user's orbot's tor or a tor running on the DOH server?

abeluck:
Ok I have a DOH server online that just forwards to cloudflare and has logging enabled. I can share it in a private channel and dump logs on request. This is assuming this is for small-scale private testing scenarios.

tladesignz:
> Ok I have a DOH server online that just forwards to cloudflare and has logging enabled. I can share it in a private channel and dump logs on request. This is assuming this is for small-scale private testing scenarios.

It's only about checking, if a typical Android network stack follows [RFC 7686](https://datatracker.ietf.org/doc/html/rfc7686#section-2), meaning, it doesn't try to resolve `.onion` domains over networks, which do not actively advertise, that they can handle exactly that.

The Apple network stack follows this, and I deducted Apple -> FreeBSD -> Unix -> Linux -> Android, should probably do the same. But of course, we want to make absolutely sure. :-)

I'll talk to you about this.

tladesignz:
Ok, just tested this with the help of Able's DoH forwarder. I'm pretty surprised (and also pissed): the Android stack doesn't seem to honour said RFC, and *doesn't* shortcut requests to `.onion` domains (like Apple systems do).

@bitmold, Abel will leave the server up for a little while longer, so you can test yourself. (I assume you want to check more devices…)

abeluck:
I can see how that RFC is very difficult to implement. Software authors often don't t know the context in which it runs, so they default to fewer restrictions.

Maybe I want to use some off the shelf dns software in my tor onion server setup, but if that software refused to work with onion urls that would seem like a bug in the software preventing me from using it. 

Likewise maybe I have a wifi hotspot whose WAN goes through Tor, if my Android device refused to resolve .onion's through it that would seem an awful lot like a bug/problem in Android.

SkewedZeppelin:
it was mentioned above that there are cases where private dns bypasses the vpn, which is different from what I've understood
in testing, did all requests route through the vpn?

tladesignz:
> it was mentioned above that there are cases where private dns bypasses the vpn, which is different from what I've understood in testing, did all requests route through the vpn?

It's unclear, if Android makes the DoH requests bypass the VPN or not. I suppose, not, though. (Well, I was wrong with another assumption here, already, but anyway…)

It would be even worse, if they would. On the other hand, while using Tor, using your own DNS resolver instead of letting the exit nodes resolve, is already reducing your privacy. *Especially*, when requests for `onion` domains leak outside, which would otherwise never leave the Tor system at all. (And on top, it breaks the functionality of actually reaching them.)

So we need to warn users about it, anyway.

Insofar, this information would not affect our decision. Hence it is work, which I would leave to interested third parties.

syphyr:
scaleway-fr dns servers now support post quantum key exchange:
[scaleway-fr] using the post-quantum X-Wing key exchange

n8fr8:
Working on the language tuning with the UX team... thinking we only show this when Full Device VPN is enabled, or if known web browsers are in the app list. 

If a user is just proxying Signal, Telegram, etc this isn't that consequential perhaps?

cstiens:
Draft language:

 **Private DNS Enabled**
Private DNS is turned on in your Android settings. This setting may interfere with your ability to use **Orbot VPN** or **Onion Sites**. If you experience connection issues, try turning Private DNS off.

**Privacy note**
When Orbot is running, **Tor handles your DNS requests** to help protect your privacy. If you turn off Private DNS **and Orbot is not running**, your internet provider (ISP) may be able to see more information about the websites and services you access.

[button - Change Settings]
[button - Ignore]

SkewedZeppelin:
I don't understand this chosen merged message. I made a diagram for the possibilities:
pdns_routing

In my understanding there are two primary leaks of Tor+Private DNS:
- the first is that the [request] circuit doesn't handle DNS, and particularly if circuit isolation is used: that means all DNS traffic for a given time period (eg. 10 mins) is over a single circuit, which enables/allows correlation by the Private DNS provider
  - Tor could actually improve this situation if the opposite of LongLivedPorts was implemented for 853, although DoH seems to have gained more traction due to being harder to block, but at the cost of increased load on the network
- the second is that resolving an onion _will_ go to the Private DNS provider and break their usage

The reasons to use Private DNS are:
- filtering DNS (malware/nrd/ads/etc.)
  - although both Firefox and Chromium based browsers have their own built in resolver, but they have no special treatment
- DNSSEC support
- ECH support
  - while the primary benefit is anti-censorship, some Tor exits have been observed blocking access
- shifting trust towards a potentially more trustworthy resolver than the exit node
  - although afaiu the exit could still just reroute the IP as it pleases anyway

edit: I did indeed verify (under GrapheneOS/Android 17) that in the private dns+vpn slot case that the system only briefly sends private dns traffic over the upstream interface in order for the vpn app to establish its connections, afterwards it goes radio silent implying all traffic is correctly routed over the tunnel.
`tcpdump -i [] tcp port 853`

edit 2: re: RFC 7686: the Chromium stance on this was WONTFIX but it probably wouldn't hurt to open an issue for Android: https://issuetracker.google.com/issues/40083293#comment13

tladesignz:
@bitmold, this was @cstiens recommendation for the texts to be displayed.

Seems very reasonable to me.

> Draft language:
> 
> **Private DNS Enabled** Private DNS is turned on in your Android settings. This setting may interfere with your ability to use **Orbot VPN** or **Onion Sites**. If you experience connection issues, try turning Private DNS off.
> 
> **Privacy note** When Orbot is running, **Tor handles your DNS requests** to help protect your privacy. If you turn off Private DNS **and Orbot is not running**, your internet provider (ISP) may be able to see more information about the websites and services you access.
> 
> [button - Change Settings] [button - Ignore]


But, AFAIU, what you merged now is [this](https://github.com/guardianproject/orbot-android/pull/1707/changes#diff-5e01f7d37a66e4ca03deefc205d8e7008661cdd0284a05aaba1858e6b7bf9103R368)

> Using a strict Private DNS with a specific resolver will leak your traffic and onion services outside Orbot\'s VPN. Please set your Private DNS to be Automatic.


Can you explain your decision?

bitmold:
A DNS leak :skull_and_crossbones:  **will always happen** :skull_and_crossbones:  whenever the user is using Strict Private DNS ( the one where they explicitly provide a hostname). So I changed the behavior to only show this warning with Strict Private DNS. Automatic Private DNS, which is the default setting for most every android device, **does not** have this leak outside of tor. 

[As you said yourself here, private DNS doesn't leak when it's in the default automatic mode](https://gitlab.com/guardianproject/tormobile/orbot/-/work_items/20#note_3406810744):
  ```
   Ok, I had a look at this with Orbot using https://browserleaks.com/ip
   1.  Orbot running, "Private DNS" switched off or auto:
   -   Clearnet site resolves, Browserleaks identifies Tor exit node correctly,
       DNS leak test shows DNS requests are originating from servers related
       to exit node.
   -   DuckDockGo .onion page also resolves and loads.
   2.  Orbot running, "Private DNS" switched to one.one.one.one:
   -   Clearnet site resolves, Browserleaks identifies Tor exit node correctly,
       DNS leak test shows DNS requests are originating from Cloudflare servers.
   -   DuckDuckGo .onion page does not resolve and does not load.
 ```

So there were these problems with the original text:

> Private DNS Enabled
> Private DNS is turned on in your Android settings. This setting may interfere with your ability to use Orbot VPN or Onion Sites. If you experience connection issues, try turning Private DNS off.
> Privacy note When Orbot is running, Tor handles your DNS requests to help protect your privacy. If you turn off Private DNS and Orbot is not running, your internet provider (ISP) may be able to see more information about the websites and services you access.

- The text makes **no mention** of the very serious situation that when you use Orbot with Strict Private DNS you **are** leaking traffic. At the absolute bare minimum, users need to know that with Strict Private DNS there is a 100% chance of a DNS leak and that this is a bad thing.

- Instead, the text says that private DNS _may interfere_ with your ability to use Orbot VPN and access onion services. This is true, you can't access onion services with it. Otherwise though, a user who isn't accessing onion services with Orbot will not experience any friction with OrbotVPN and would continue to use it - all the while being completely unaware that they are leaking their DNS requests! In the original draft, we merely suggest users disable this setting if it's getting in the way of their browsing, which for many, it will not.

- This statement is incorrect:
 > Privacy note When Orbot is running, Tor handles your DNS requests to help protect your privacy. If you turn off Private DNS and Orbot is not running, your internet provider (ISP) may be able to see more information about the websites and services you access.
 - Orbot _does_ handle your DNS requests when users have Private DNS enabled in Automatic mode, which is the default we can expect almost all of our users to have.
 - But Orbot doesn't handle anything and leaks DNS with Strict Private DNS. It's incorrect to say this about Tor and DNS here. With Strict Private DNS enabled, we show the user a message saying Tor is handling their DNS traffic when it just very much isn't.

- > If you experience connection issues, try turning Private DNS off.
  - we absolutely do not want to tell our users, especially ones who aren't tech savvy, to disable private DNS on their device. **that's a harmful thing to do**. We _only_ want to tell them to stop using Strict Private DNS with Orbot. 
  - If a user followed these misleading instructions and disabled Private DNS, it will remain disabled when not using Orbot, after Orbot is uninstalled, when using another privacy app, etc etc.
       
In light of all this, I chose to tell users to go back to using automatic private DNS - the default setting that doesn't leak. By doing so, we can rest assured that we _never_ sent our users on a pointless + disorienting odyssey through the system settings which resulted in both fatigue & a permanent reduction to their device's security. We also made it clear that there is a guaranteed leak,  _**if and only if **_  Strict Private DNS is active.  

- Regarding the suggestion @cstiens had for a "Change Settings" button, I had mentioned at some point to her + perhaps others on the team, that it's impossible! Android provides no way for apps to progrmatically open their device's Private DNS settings. So that's why I didn't include that.

<span style="color: red;">TorVPN is outright disabling their app until the user stops using Strict Private DNS, FWIW. I think that ultimately is the correct thing to do - it prevents non-technical users into leaking traffic they think is going through Tor. Alternatively, we could disable Orbot + Strict Private DNS by default, while providing a way for power users to go very out of their way to  _opt into enabling_ Strict Private DNS, if for whatever reason they have a legitimate use case for it.</span>

See also these TorVPN discussions that informed my decision: 
- https://gitlab.torproject.org/tpo/ux/design/-/work_items/342
- https://gitlab.torproject.org/tpo/applications/vpn/-/work_items/147
- https://gitlab.torproject.org/tpo/core/onionmasq/-/merge_requests/694
- https://forum.torproject.org/t/tor-vpn-doesnt-work-with-private-dns-enabled/21198/7
> It’s recommended to disable Private DNS feature (this feature name really should be changed to “Encrypted DNS” or just “DNS-over-TLS” to not give users a false sense of privacy) of Android if you need to use TorVPN (or Orbot). Private DNS settings is enforced for both connections outside VPN tunnel and inside VPN tunnel, so even if you’re able to get TorVPN working with Private DNS, you would let this one DNS resolver know every website you visited via Tor, which largely defeats the purpose of Tor.

SkewedZeppelin:
The merged message implies that the DNS request goes outside of the tunnel though.
Yes the request leaks to a third party, but it is still in tunnel and routed over Tor.

My primary concern is that people may end up thinking years of their DNS traffic was observable by their ISP on the next update, when it was infact routed over Tor.

re TorVPN: afaiu they're disabling use with it primarily due to how onionmasq handles it currently, not on purpose


edit: perhaps a wording like this:
> Use of Private DNS with a custom resolver has numerous privacy issues. In all cases the request is effectively leaked to the resolver and not handled by the request's matching Tor circuit. As a result: a) the requests will be routed over Tor, but the resolver can correlate your requests across any Tor circuits; b) .onion requests will not work.

bitmold:
I appreciate your perspective and bringing up the ambiguity with how i went explaiIning that there is a leak. I agree that important to not scare existing users, and that clarity is a good way to do that. 

But to me your suggestion seems too technical, In the first place, many users don't know what DNS even means - especially all kinds of nontechnical people who find themselves coming to Orbot while under duress in uprisings. periods of censorship, etc. 
 
I really like the approach theimpulson takes with TorVPN here:
https://gitlab.torproject.org/tpo/applications/vpn/-/merge_requests/564/diffs

> <string name="error_private_dns_title">Change your Private DNS settings to connect</string>
> 
>    <string name="error_private_dns_message">Tor VPN does not support using a specific hostname for Private DNS.This bypasses Tor and can reveal the sites you visit to your DNS provider.\n\nIn Settings, set Private DNS to Automatic or Off.</string>

They use this text in a dialog that comes up whenever there's a network change event. It seems other parts of their design include disconnecting the VPN until strict DNS is turned off and a "Learn more" button that opens a website outside of TorVPN.

bitmold:
I really appreciate how TorVPN clearly states that using Strict Private DNS **"bypasses Tor"**. It gets right to the point.

The strength in their approach also comes from how TorVPN is positioned as an authority on what is and isn't fundamentally safe for users of all backgrounds. 

So even though the current Orbot implementation unambiguously states that there is a leak, Orbot still delivers that information in such a way where the user now has to decide for themselves about what this means, if this is important, etc.... 
- Orbot's passive stance could suggest that this kind of  leak might not be that serious, since it's just a dismissable dialog box they quickly click through before they use Orbot to get on their favorite censored websites. 
- Other users might be worried about a leak, but think that things are fine since they're still able to get to websites that might normally be censroed
- Users who are genuinely invested in knowing if this "really matters" suddenly have to find answers to all kinds of new questions, such as "Is it more or less secure to use Tor with Private DNS, even though there's a warning?' "What is DNS?" etc. if you don't have a technical background, this rabbit hole of questions involves reading a bunch of conflicting/unrelated/incorrect info online, possibly written in a language you don't speak.

Instead, TorVPN's approach gently conveys to the user
- this is a real risk that needs to be taken seriously 
- that all is well, nothing's leaking
- clear instructions on what to do to make the problem go away
- that there are not multiple ways the user can connect to Tor, some of which are evidently leakier than others.  
- that some innocuous Android setting ( perhaps from the user's perspective one of many ) does not somehow have a special power that can partially undermine TorVPN's anonymity 
- that when a (nontechnical) user makes a change to their device settings that they may not fully understand, TorVPN will step in and make sure they don't inadvertantly shoot themselves in the foot

I think it'd be wise for Orbot to adopt a similar strategy here. Perhaps while also providing a setting the user has to go out of their way to access, where they  explicitly opt into being able to using Strict Private DNS + Orbot ...

tladesignz:
*sigh*, from my point of view, you're a little too emotionally invested in this, @bitmold.

Yeah, sure, we're only talking about the *strict* private DNS mode here. That fine distinction is not something, @cstiens isn't aware of, probably. That should be taken into account when working with her wording, and it should probably be slightly adapted.

On the other hand, you're not weighing your hip-shot of a text with the same gold scale.

> Using a strict Private DNS with a specific resolver will leak your traffic and onion services outside Orbot's VPN. Please set your Private DNS to be Automatic.

"leak your traffic"? There is no traffic leaked. A list of visited domains/onion services is leaked. That's slightly less than the complete traffic.

On the other hand, @cstiens text doesn't mention the leaking of visited domains at all. That should probably be improved.

> This bypasses Tor and can reveal the sites you visit to your DNS provider.

I do like this piece of TorVPN's text, I have to admit. It's crisp and clear.

*But*, what she mentions, and what your text (and also TorVPN's) does let completely fall under the table, is the fact, that, as soon as users *switch off* Orbot, they *worsened* their situation.

I very much assume, "automatic" means, Android tries to route DNS queries through Google's 8.8.8.8 server and probably falls back to the DHCP-provided one, if that fails, since you cannot set your own DNS provider with "automatic". And, obviously, when a VPN is used which configured a DNS resolver, Android will instead forward it to that one.

So, if users choose "off", they're now leaking their DNS queries to their ISP, which, esp. in the Americas, that is a serious privacy problem.

And if users choose "automatic", they're now leaking their DNS queries to Google. Which is also a serious privacy problem.

So, the optimal strategy for users, is to switch strict private DNS off, while using Orbot, and then immediately turn it back on, when they stop Orbot.

That's a tough ask.

And that's why I think, we shouldn't let that aspect unmentioned. Which is exactly what @cstiens text does.


@SkewedZeppelin, thank you for your input on this. While, as a programmer, I do love the technical detail in your suggested text, I did at one point understand (or better: give in to the fact), that, at most people, you cannot throw such level of detail. They'll just phase out and ignore warnings all together.

That's why I currently would prefer a slightly revised version of @cstiens text.

- Only shown when "strict" mode, and adapted to that fact.
- Adding that piece from TorVPN about the leakage.

Here's an attempt:

> Strict Private DNS Enabled
> 
> Strict Private DNS is turned on in your Android settings. This bypasses Tor and reveals the sites you visit to your DNS provider. It also interferes with your ability to use Onion sites. While using Orbot, better switch it to "automatic". 
> 
> Privacy note
>
> When Orbot is running, Tor handles your DNS requests to help protect your privacy. If you set Private DNS to "off" or "automatic" and Orbot is not running, your internet provider (ISP) or Google is able to see more information about the websites and services you access.

SkewedZeppelin:
> I very much assume, "automatic" means, Android tries to route DNS queries through Google's 8.8.8.8 server and probably falls back to the DHCP-provided one

no, it tries to upgrade the network/DHCP provided one to DoT (Comcast was a big supporter iirc):
https://developer.android.com/reference/android/app/admin/DevicePolicyManager#setGlobalPrivateDnsModeOpportunistic(android.content.ComponentName)
> In this mode, the DNS subsystem will attempt a TLS handshake to the network-supplied resolver prior to attempting name resolution in cleartext. 

there are very few cases where it would actually fall back to using 8.8.8.8, the network provided values are always preferred

it also has a hardcoded list which it tries to upgrade from DoT to DoH (Google and CF):
https://cs.android.com/android/platform/superproject/+/android-latest-release:packages/modules/DnsResolver/PrivateDnsConfiguration.h;drc=9991b91a16174826f2a5d071c4fa9da83e99b8a6;l=276

tladesignz:
> > I very much assume, "automatic" means, Android tries to route DNS queries through Google's 8.8.8.8 server and probably falls back to the DHCP-provided one
> 
> no, it tries to upgrade the network/DHCP provided one to DoT (Comcast was a big supporter iirc): https://developer.android.com/reference/android/app/admin/DevicePolicyManager#setGlobalPrivateDnsModeOpportunistic(android.content.ComponentName)
> 
> > In this mode, the DNS subsystem will attempt a TLS handshake to the network-supplied resolver prior to attempting name resolution in cleartext.
> 
> there are very few cases where it would actually fall back to using 8.8.8.8, the network provided values are always preferred
> 
> it also has a hardcoded list which it tries to upgrade from DoT to DoH (Google and CF): https://cs.android.com/android/platform/superproject/+/android-latest-release:packages/modules/DnsResolver/PrivateDnsConfiguration.h;drc=9991b91a16174826f2a5d071c4fa9da83e99b8a6;l=276

Oh, wow. Very interesting. Thank you!

It seems rather useless to upgrade an ISP-provided DNS server to DoT, except, when they actually provide you with a public one which doesn't terminate within their network. Which typically means, they're not interested in your data anyway, and rather let others bear the cost of DNS resolution? 

So, in the end, that protects you from passive listeners on the route behind the ISP?

Pretty weird.

bitmold:
> _sigh_, from my point of view, you're a little too emotionally invested in this, @bitmold.

I'm definitely emotionally invested in preserving the anonymity assumptions of Tor and keeping Orbot as simple as possible for every user. Talking to me this way is dismissive and inappropriate, please don't do this again. 

> But, what she mentions, and what your text (and also TorVPN's) does let completely fall under the table, is the fact, that, as soon as users switch off Orbot, they worsened their situation.
> I very much assume, "automatic" means, Android tries to route DNS queries through Google's 8.8.8.8 server and probably falls back to the DHCP-provided one, if that fails, since you cannot set your own DNS provider with "automatic". And, obviously, when a VPN is used which configured a DNS resolver, Android will instead forward it to that one.

- as said by @SkewedZeppelin and is widely understood across many of the Google/AOSP/TorVPN/GrapheneOS docs+discussions posted throughout this issue - this is **_not_** how Automatic Private DNS works. Instead, Automatic Private DNS optimistically tries to use DNS w/ TLS over your phone's active connection. In rare fallback cases it falls back to 8.8.8.8 (and even then, this can vary depending on your OS vendor -[ GrapheneOS uses 1.1.1.1 for instance...](https://grapheneos.org/faq#default-dns)) So, **I still think it's both correct and sufficient  to tell the user to use  Automatic Private DNS** instead of: 
  - telling them to turn if Off instead of Automatic 
  - your earlier idea of telling the user that they really ought to go deep into the system settings and toggle between Private DNS Strict and Automatic _every_ time they en/disable the VPN 

- And, like I said earlier,  **we should disable the VPN when Strict Private DNS is enabled** [like how TorVPN does.](https://gitlab.torproject.org/tpo/core/onionmasq/-/merge_requests/694). Optionally we can have an out of the way setting to allow the user to proceed... 

  - We have this same problem as TorVPN:
  > [`@cve:`](https://gitlab.torproject.org/tpo/core/onionmasq/-/merge_requests/694#note_3418283) This would leak DNS, I am not sure if this is something we want.
  > [`@cyberta:`](https://gitlab.torproject.org/tpo/core/onionmasq/-/merge_requests/694#note_3418763) I agree with 2., it clearly contradicts the anonymity promises of Tor.

- @tladesignz I hear your point when I was too vague in saying "leak your traffic" instead of something more specific like "leak your DNS traffic"/etc. 
  
  Since it's now very clear that  Orbot users would _not_ have worsened their situation after turning of the VPN, then this actually wasn't something that TorVPN let "completely fall under the table" with their succinct dialog messages. And so, **I still think we should use the text TorVPN uses:**

  > `Dialog Title:` Change your Private DNS settings to connect
  > `Dialog Message:` Tor VPN does not support using a specific hostname for Private DNS.This bypasses Tor and can reveal the sites you visit to your DNS provider.\n\nIn Settings, set Private DNS to Automatic or Off.
  - These _simple_ strings directly serve to:
     1. State that Strict Private DNS doesn't work with Tor. Whatever your technical knowledge is, it's disabled because it would violate something essential about _why_ you are using Tor. 
     2. The reason: that it would leak/reveal the sites you visit.
     3. While still conveying the concept of a "DNS Provider" - this gives context to technical users about the nature of the problem without overwhelming non-technical users. 
 
  The vast majority of people in the world fundamentally  `have` `no` `understanding` `whatsoever` of what DNS is. If the dialog (which a user may not even have localized in their language) is something the user can click through, then it's safe to assume that (many) users will. This is harmful. 

Let's not build a footgun into the app. Given what we now know about Private DNS I don't see any issues with:
- turning off the VPN  
- using TorVPN's succinct error stings, perhaps with minor modifications 
- providing power users a setting in the menu to allow themselves the option to use Orbot with Strict Private DNS 

do you?

bitmold:
As an aside, [another quirk is that **_only_** the Owner Profile (essentially the "Admin" user) can change the Private DNS settings, and that Owner's Private DNS setting applies to _every other_ profile.](https://discuss.graphenheos.org/d/21S85-grapheneoss-recommendation-on-dns-servers/16)  This is different from VPNs, where every profile has its own distinct VPN. So if you happened to be stuck using Orbot on a profile that's now Owner, you wouldn't be able to change & rechange the Private DNS setting and would perhaps be inclined to ignore whatever a dismissable warning dialog states. 

> [Our recommendation is to use the DNS provided [automatically] by the network and to use a VPN](https://discuss.grapheneos.org/d/2185-grapheneoss-recommendation-on-dns-servers/15). Most people should avoid using [Strict] Private DNS. If you want to filter traffic, then use something like RethinkDNS which provides local filtering alongside being able to use a WireGuard VPN. Filtering traffic adds an easy way to fingerprint for websites and other services so you shouldn't filter traffic if you want to avoid standing out.

And semi-relatedly, GrapheneOS' FAQ has [this](https://grapheneos.org/faq#default-dns) to say about private DNS:

> Using the network-provided DNS servers is the best way to blend in with other users. Network and web sites can fingerprint and track users based on a non-default DNS configuration. Our recommendation for general purpose usage is to use the network-provided DNS servers.

> In some broken or unusual network environments, the network could fail to provide DNS servers as part of dynamic IP configuration. The OS has high availability fallback DNS servers to handle this case. A network can fail to provide DNS servers in order to fingerprint clients based on what they use as the fallback so it's important for it to be consistent across each install. GrapheneOS replaces Google Public DNS with [Cloudflare DNS](https://developers.cloudflare.com/1.1.1.1/faq) for the fallback DNS servers due to the superior privacy policy and widespread usage including as the fallback DNS servers in other Android-based operating systems. We're considering hosting our own servers and offering a toggle for using the standard (Google) servers to blend in with other devices similarly to how we handle the internet connectivity checks

SkewedZeppelin:
I still believe the user having their own choice of DNS and those benefits are important, but after more discussion, I do agree the simple "just don't allow it" TorVPN approach is better, at least for now.

aside
> Using the network-provided DNS servers is the best way to blend in with other users

this is kind of arguable since there'd be other GOS related connections which the network could still observe and be fingerprintable. and if you want privacy from websites the solution is Tor Browser _on desktop_ **not android** and not custom vs default dns. (yes the security of both is awful (I'm pretty sure there are dozens of known issues currently fixed that they haven't backported to esr[1]), but tb for android is also awful fingerprint-wise)


[1] eg. firefox 151 had 153 security fixes, its corresponding esr update 140.11 only had 79 backports.
and firefox 150 had 354 security fixes, esr update 140.10 only had 284 backports. some of those are obviously becuase the code wasn't there, but I would love to see numbers from mozilla on how many they just didn't bother to backport.

bitmold:
Fair enough @SkewedZeppelin 

We can provide this setting for people with use cases like yours / differing analyses / varying threat models such as yourself. But it will be  somewhere in the app settings instead of being actionable on the dialog. 
 
> this is kind of arguable since there'd be other GOS related connections which the network could still observe and be fingerprintable. 

I think this is getting into the weeds a bit, but i see your point here. From the perspective of the DNS resolver, there very well could be some things distinguishing you from other Android devices on the network, but afaiu client-side DNS resolution on Graphene is very similar to AOSP's, if not identical. I could be in the wrong though... You would at, the very least blend, in with other GOS devices on your network, but yeah, let's table this. I'm down to discuss further if you think it fundamentally undermines my solution idea that I outlined above, but at the moment it doesn't seem like you think so.

SkewedZeppelin:
thank you @bitmold

tladesignz:
> - turning off the VPN

Against that. There's a lot of people out there, who just want to bypass censorship. And some of them just try every toggle they can find. I think it's wrong to exclude them on principle.

> - using TorVPN's succinct error stings, perhaps with minor modifications

You had copytext from @cstiens, you had copytext from Tor Project, but you still decided to roll your own. Well.

It is a well-known fact, that both, the big US ISPs *and* Google data-mine DNS requests. Which **is** a serious privacy invasion. This is not something to drop under the table lightly for the sake of simplicity.

It is also a well-known fact, that the user-group of Tor/Orbot is divided between a (smaller) group of IT-sec aware folks, who mostly seek the anonymity aspects of Tor. And then there's a huge group of people, who mostly seek the censorship-circumvention properties of Tor. And I want to point out, that **this** is the group, we're supposed to serve with the current grant.

I do recognise, that privacy concerns might not be a thing for the latter. But I still think, it's something we should not completely ignore.

> - providing power users a setting in the menu to allow themselves the option to use Orbot with Strict Private DNS

Sé non e complicato, non e divertente.

bitmold:
> You had copytext from @cstiens, you had copytext from Tor Project, but you still decided to roll your own. Well.

I don't understand why you seem fixated on my earlier draft of the text. We _all_ had incomplete understandings of this complicated problem. @cstiens when she first wrote the text without mentioning the fundamental problem of bypassing Tor, you when you thought that draft was sufficient or with other things you've been corrected on here, and of course being mistaken or incorrect here myself, such as when my attempt to fix that draft was too broad.

 I found out about TorVPNs text _after_ attempting my own. Since then I've been advocating for changing my own (most recently, in my last message to you...)

Until its changed though, no user is left with my insufficient draft. Its just on github. 

> It is a well-known fact, that both, the big US ISPs and Google data-mine DNS requests. Which is a serious privacy invasion. This is not something to drop under the table lightly for the sake of simplicity.

> it is also a well-known fact, that the user-group of Tor/Orbot is divided between a (smaller) group of IT-sec aware folks, who mostly seek the anonymity aspects of Tor. And then there's a huge group of people, who mostly seek the censorship-circumvention properties of Tor. And I want to point out, that this is the group, we're supposed to serve with the current grant.

Did you read my last response to you? When users have automatic DNS set they hide from their censor/ISP the addresses they're trying to reach while using a VPN app. 

Its reasonable to assume that for most (censored ) users their mental model is roughly  "when I use a privacy app like Orbot im anonymous and the censor is unable to see what I'm doing, and when I turn it off I lose this privacy and am under the censors gaze". 

No VPN app can help users when they turn it off. 

On the other hand, _alowing_ users to reveal the sites that they visit when they use Strict Private DNS is harmful to all users. Censored ones especially. 

How are any users excluded?  Its a lot easier to tell a user "change this setting because it will harm you and defeats the point of why you're here", than it is to explain some stuff about DNS resolution before letting users leak the sites they're visiting while they bypass their censor. 

Hell, what if part of a censorship campaign involved the State encouraging the use of a specific DNS resolver to be paired with things like strict private DNS? Now those censored users are very much at risk deanonimizing what they're doing over Tor to their censor...

Besides, in censored areas what use does a user get out of Strict Private DNS without a VPN anyway? If they can reach a custom DNS provider and successfully use it to resolve hostnames, ultimately the censor still:
 - sees the resolved IPs they're trying to access 
 - has the ability to prevent the user from accessing them 
 - this is true in uncensored areas too. If you use a specific DNS resolver in the US the american ISPs still see every site you visit after resolving those hostnames. The only difference is that the ISP just aren't handling dns resolution anymore. On android you only get away from the evil corporation, crooked ISP , censor when DNS resolution happens inside a VPN, which strict private DNS violates. 

Orbot should not bypass Tor. preventing this from happening is in line with, not opposed to, the goals Orbot has with censorship circumvention.

Why do you think thatbinstead we should let users deanonymize themselves so easily? Or that by preventing this from easily happening we are throwing users under the table?

tladesignz:
> Did you read my last response to you? When users have automatic DNS set they hide from their censor/ISP the addresses they're trying to reach while using a VPN app.

Yes, I read your messages, and I fully understand that, but you invoke the feeling in me, that you didn't read *my* messages.

Once again: AS SOON AS THEY DISCONNECT the VPN, the users now have *less* privacy than before. I'm of the opinion that that is not something to ignore lightly. (And my learning about  the details of what "automatic" exactly means didn't change that at all.) That is a conundrum, we cannot resolve. Hence, it is something we need to inform the user about.

To be totally clear: My problem with the way you're acting here is these two things:

1. You still seem to not wanting to acknowledge that fact, hence seem to argue to remove that addition.
2. I tried to work with cstiens to come up with the best explanation text for an ugly situation, but you just ignored our efforts and instead of engaging in it just did your own thing. I had to actually force you to engage in the discussion. I hate doing that, but I'm certainly not the one to avoid a conflict.

bitmold:
> Once again: AS SOON AS THEY DISCONNECT the VPN, the users now have less privacy than before. I'm of the opinion that that is not something to ignore lightly. (And my learning about the details of what "automatic" exactly means didn't change that at all.) That is a conundrum, we cannot resolve. Hence, it is something we need to inform the user about.

1. I addressed this multiple times throughout the issue, most recently, last night:

[I addressed this less than 16 hours ago by saying this:](https://github.com/guardianproject/orbot-android/pull/1707#issuecomment-5028949682)

> Its reasonable to assume that for most (censored ) users their mental model is roughly "when I use a privacy app like Orbot im anonymous and the censor is unable to see what I'm doing, and when I turn it off I lose this privacy and am under the censors gaze".
> No VPN app can help users when they turn it off.

[And this](https://github.com/guardianproject/orbot-android/pull/1707#issuecomment-5028949682): 

> Besides, in censored areas what use does a user get out of Strict Private DNS without a VPN anyway? If they can reach a custom DNS provider and successfully use it to resolve hostnames, ultimately the censor still:
> 
> sees the resolved IPs they're trying to access
> has the ability to prevent the user from accessing them
> this is true in uncensored areas too. If you use a specific DNS resolver in the US the american ISPs still see every site you visit after resolving those hostnames. The only difference is that the ISP just aren't handling dns resolution anymore. On android you only get away from the evil corporation, crooked ISP , censor when DNS resolution happens inside a VPN, which strict private DNS violates.

[I said something similar 3 days ago](https://github.com/guardianproject/orbot-android/pull/1707#issuecomment-5013005030)
> Since it's now very clear that Orbot users would not have worsened their situation after turning of the VPN, then this actually wasn't something that TorVPN let "completely fall under the table" with their succinct dialog messages. And so, I still think we should use the text TorVPN uses...

[Or about a week ago when I quoted `Lind` from a discussion on TorVPN:](https://forum.torproject.org/t/tor-vpn-doesnt-work-with-private-dns-enabled/21198/7)
> this feature name really should be changed to “Encrypted DNS” or just “DNS-over-TLS” to not give users a **false** sense of privacy

Care to respond to these criticisms/questions I raised with this scenario, instead of continuing to make insinuations about my character? To reiterate:

- **Using Strict Private DNS and no VPN - offers something of so little value to users - whether you are under active censorship or not.** You're not hiding _anything_ from your ISP/censor by using your own DNS resolver, since once the DNS resolution is complete and you get an IP address, you _**still**_ have to use your censorious ISP to access that address. What am i missing? Obviously a user is worse off when they turn off their privacy VPN....
- Even if Strict Private DNS somehow provided users something of significant value when they were not using a VPN like Orbot - which, it does not - there's _still_ not much we can do for users when they turn Orbot off. Like I said, we should optimize for users who are using the privacy-preserving, anti-censorhip tool. Users approach apps like Orbot with an understanding that they are better off with it on, and that they are susceptible to surveillance and censorship with it off. 

As far as I understand, and have been trying to advocate for here:

- Strict "Private" DNS + no VPN - offers `next to nothing` to users who are concerned about thier ISP tracking and/or censoring them, since you still rely on a direct connection with that untrusted ISP to do every other aspect of your online activity that isn't resolving hostnames 
- Strict "Private DNS + a VPN":
  - is an incredibly confusing "feature" of Android, it defeats the point of DNS inside the VPN Tunnel and users are not given enough information about how this overrides their VPN to make an informed decision. 
  - With Orbot, it fundamentally belies the anonymity promises of Tor 
  - Users not in the Owner profile have no way of knowing if Strict Private DNS is enabled or not, let alone of changing the value 

Or, to put it another way:

- Automatic Private DNS + No VPN - ISP provider resolves hostname w/ TLS if possible, ISP handles connection once resolved 
- Strict Private DNS + No VPN - 3rd party resolves hostname, ISP handles connection once resolved. 

### **In either case, your ISP/Censor still sees what you are doing whether or not they are handling your DNS resolution - so a user would `NOT` be significantly worse off when they turn Orbot off, they'd be back where they started. When the VPN is turned off, Strict Private DNS does not give you "more" privacy in any meaningful sense of word than Automatic Private DNS would.** 

|         | Automatic Private DNS                                                                                                        | Strict Private DNS                                                                                                              |
|---------|------------------------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------|
| VPN Off | ISP/Censor likely handling your DNS Resolution. ❌ <br><br>ISP/Censor, then sees all addresses you connect to after DNS resolution. ❌ | 3rd Party handles your DNS resolution. ❓ <br><br>ISP/Censor **_still_** sees each and every address you connect to after DNS resolution ❌  |
| VPN On  | VPN resolves your DNS queries, and handles your connection<br><br>ISP/Censor is out of the picture ✅                                                         | 3rd Party handles your DNS resolution ❓  <br><br>DNS Leaks from the VPN to this 3rd Party ❌                                            |

 
>  but you just ignored our efforts and instead of engaging in it just did your own thing. I had to actually force you to engage in the discussion

2. This is an incredibly bad faith way of looking at me here. Trust me, you couldn't force me to do _anything_. I've been engaging here voluntarily because I want to make sure that the implementation is done correctly. Like I said before, I attempted to address some shortcomings I perceived in the implementation that was proposed here, and I've  been active on this issue ever since I realized my approach was naive and had shortcomings of its own. This entire time I've been eagerly responding to you and anyone else here who has engaged, while I've been researching and sharing things I've learned about how related apps in the space handle this problem.

bitmold:
@tladesignz, after reading the above,

- do you still disagree with me and think that a user with no VPN that has Automatic Private DNS enabled somehow has "less" privacy than if that same user had Strict Private DNS 
  - If so, given the issues I raised above, _why_ do you disagree with me and think that the former situation is somehow worse than the latter?

If you agree with me now, then it should be abundantly clear that the dialog text that TorVPN came up with is actually really quite good, since we, nor them, would be throwing users under the table when they stop using our VPNs...

PR #1716, opened 2026-06-07
Properly check for emulator to fix Tor starting up

syphyr:
This fixes Tor being able to startup on production devices.  Properly check for emulator by value of Build.HARDWARE.  All Google emulators use goldfish hardware.


syphyr:
Added logging just in case.

syphyr:
simplified check

Check hardware is goldfish instead of a fingerprint check.

syphyr:
All emulators show `Build.HARDWARE == "goldfish"` so just use that.

Issue #1713, opened 2026-06-05
[BUG] Orbot does not detect it is connected when automatically starting up on boot


syphyr:
**Describe the Bug**
Set Orbot to automatically start on boot.  After the device is booted up, open Orbot and the toggle shows that it is not connected, when it actually is connected in the notification bar.

**To Reproduce**
1. Set Orbot to automatically startup on boot
2. Open Orbot after boot and it shows it is not connected, when it actually is

**Expected Behavior**
Orbot UI should show it is connected within the application when it automatically starts up on boot.

**What Custom Configuration Do You Use?**
If applicable, how did you set up Orbot? Did you configure some settings?

**Screenshots**

**Smartphone (please complete the following information):**
- Device: T813
- OS: Nougat
- Version: 5c383d5caad9506dbdefdfb2fa3a02c406cfb6de

**Crash Logs (Advanced)**
If applicable, add crash logs collected using ADB Logcat.

**Additional Context**

syphyr:
@bitmold I think this started to happen when Orbot changed the way it detects if Tor is connected (which included changes to the tor-android library).

bitmold:
Will look into this, in the future if you are not using a specific version of Orbot, instead of putting latest could you please put the commit hash when reporting these bugs. Thanks

syphyr:
> Will look into this, in the future if you are not using a specific version of Orbot, instead of putting latest could you please put the commit hash when reporting these bugs. Thanks

Added a hash to bug report. Thanks.

PR #1726, opened 2026-06-22
Orbot UI boot fix 1713

bitmold:
When Orbot started on boot, its UI wouldn't get updated since a few releases 

From #1713 @syphyr says:
> @bitmold I think this started to happen when Orbot changed the way it detects if Tor is connected (which included changes to the tor-android library).

I could reproduce this behavior starting Orbot on boot and noticing that even though Tor bootstrapped fine in the notifications, & that I could use the VPN to reach tor, vut the UI was locked out of sync it's Off state...

This PR does the following: 
- Totally removes the parts `OrbotActivity` from tracking `TorStatus` status in the variable `previousReceivedTorStatus: String?`. From many years ago, before `OrbotActivity`was using a view model and was rewritten in Kotlin, this variable would make sure we weren't over-updating the UI due to some buggy thing in OrbotService. This stuff makes no sense now that `OrbotActivity` *actually* uses a view model, not to mention that `OrbotService` has changed a lot in those many years.
- 
- In `OrbotActivity`'s `onResume()` **_always send CMD_ACTIVE to OrbotService_** instead of just doing it when OrbotActivity's view model is set to `On`. `OrbotService` replies with Tor's status, causing the UI to update - fixing the bug. Theres no risk in sending `CMD_ACTIVE` here. If Tor isn't running, `OrbotService` will just gently reply with its status, and not do anything dangerous. 

- _also accidentally commited a lint thing that I thought was on another branch :woman_shrugging:_

This seems to be working great. But would especially appreciate a few eys on this @n8fr8 @tladesignz @syphyr since it deals with the foundational relationship between `OrbotActivity` and `OrbotService` the whole app's built on.

### DIFF
--- a/app/src/main/java/org/torproject/android/OrbotActivity.kt
+++ b/app/src/main/java/org/torproject/android/OrbotActivity.kt
@@ -26,7 +26,6 @@ import androidx.navigation.ui.setupWithNavController
 import com.google.android.material.bottomnavigation.BottomNavigationView
 import com.scottyab.rootbeer.RootBeer
 import org.torproject.android.service.OrbotConstants
-import org.torproject.android.ui.connect.ConnectUiState
 import org.torproject.android.ui.connect.ConnectViewModel
 import org.torproject.android.ui.connect.RequestPostNotificationPermission
 import org.torproject.android.ui.core.BaseActivity
@@ -46,8 +45,6 @@ class OrbotActivity : BaseActivity() {
 
     var portSocks: Int = -1
     var portHttp: Int = -1
-
-    var previousReceivedTorStatus: String? = null
 
     // used to hide UI while password isn't obtained
     private var rootLayout: View? = null
@@ -69,8 +66,6 @@ class OrbotActivity : BaseActivity() {
                 window.decorView.systemUiVisibility and View.SYSTEM_UI_FLAG_LIGHT_STATUS_BAR.inv()
         }
 
-        previousReceivedTorStatus = savedInstanceState?.getString(KEY_TOR_STATUS)
-
         // programmatically set title to "Orbot" since camo mode will overwrite it here from manifest
         title = getString(R.string.app_name)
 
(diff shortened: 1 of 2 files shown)


syphyr:
This works for me. Thanks for fixing this!

app/src/main/java/org/torproject/android/ui/connect/ConnectFragment.kt
```diff
@@ -200,9 +200,7 @@ class ConnectFragment : Fragment(),
 
 
     // starts Tor, plus often the VPNService. For power users it runs tor as a proxy
-    fun startTorConnection() {
-        setState(TorService.ACTION_START)
-    }
+    fun startTorConnection() = setState(TorService.ACTION_START)
```
tladesignz, line 203: 🤨 I'm not convinced about this idiom. And why change it here in a commit which has nothing to do with it. And then only this function.

app/src/main/java/org/torproject/android/OrbotActivity.kt
```diff
@@ -233,9 +218,14 @@ class OrbotActivity : BaseActivity() {
     override fun onResume() {
         super.onResume()
 
-        if (connectViewModel.uiState.value == ConnectUiState.On) {
-            sendIntentToService(OrbotConstants.CMD_ACTIVE)
-        }
+        /**
+         * When OrbotService gets CMD_ACTIVE it:
+         * 1. Checks if the control part is open & tor is connected:
```
tladesignz, line 223: "control part" looks like a typo. "control port" probably?

tladesignz (changes requested):
Otherwise, this looked good to me, when I did a quick smoke test.

n8fr8:
Great, code looks good. Will try to smoke test on device today.

app/src/main/java/org/torproject/android/ui/connect/ConnectFragment.kt
bitmold: When I was debugging the problem I stumbled over the ConnectFragment. I'm generally in favor of this idiom in the context of bugs like this. It creates friction when a developer wants to shoehorn a kludge into a one liner.

app/src/main/java/org/torproject/android/OrbotActivity.kt
bitmold: Ha ya thanks will fix. Too many hours spent stating at screens this week

app/src/main/java/org/torproject/android/ui/connect/ConnectFragment.kt
tladesignz: Still not convinced. There's a benefit in having a visually distinctive appearance of functions versus a variable assignment: ease of readability.

Additionally: "If it ain't broke, don't try to fix it."

If we'd agree on using this style, this should be a concerted effort on doing it everywhere. Or at least everywhere we can find in a given time frame and then create one or more commits which contain only these changes.

Sneaking this in in totally unrelated PRs is, well, sneaky.

Issue #1112, opened 2024-03-16
[BUG] No automatic network timeout (in german: Kein automatischer Netzwerk-Ruhezustand) setting does not change anything


Uneccessary:
**Describe the Bug**
The setting "Kein automatischer Netzwerk-Ruhezustand" does not change anything, no matter if you enable or disable it, Orbot will always turn into "sleep-mode" or whatever you call it, ingnoring this setting completly.

**To Reproduce**
1. Disable / Enable the "Kein automatischer Netzwerk-Ruhezustand" setting
2. Turn Wifi off

**Expected Behavior**
When you enable the Setting, Orbot will not go automatically into "sleep-mode" if no internet is avaible

**What Custom Configuration Do You Use?**
If applicable, how did you set up Orbot? Did you configure some settings?

Nope, everything is the default, except Snowflake is enabled.

**Screenshots**

**Smartphone (please complete the following information):**
- Samsung Galaxy A33 5G
- Android 14
- Version: v17.1 RC-1

**Additional Context**
This bug didn't appeard in the prior stable version of Tor, avaible in the Gurdian's Project F-Droid Repo

bitmold:
Yeah, this actually does nothing... @n8fr8 😮‍💨

PR #1727, opened 2026-06-22
for #1112 removed unused no network autosleep pref

bitmold:
@n8fr8 what did this setting once do? Was trying to imagine how it modified the `torrc`. 

It seems like it's been removed for some time. So I just propose getting rid of the setting here all together. But I figure you have more context on what this once was, so maybe it'd be beneficial and trivial to it back too?

### DIFF
--- a/app/src/main/java/org/torproject/android/ui/v3onionservice/OnionServiceActivity.java
+++ b/app/src/main/java/org/torproject/android/ui/v3onionservice/OnionServiceActivity.java
@@ -51,7 +51,7 @@ public class OnionServiceActivity extends BaseActivity {
 
         mLayoutRoot = findViewById(R.id.hostedServiceCoordinatorLayout);
         fab = findViewById(R.id.fab);
-        fab.setOnClickListener(v -> new OnionServiceCreateDialogFragment().show(getSupportFragmentManager(), OnionServiceCreateDialogFragment.class.getSimpleName()));
+        fab.setOnClickListener(_ -> new OnionServiceCreateDialogFragment().show(getSupportFragmentManager(), OnionServiceCreateDialogFragment.class.getSimpleName()));
 
         mContentResolver = getContentResolver();
         mAdapter = new OnionV3ListAdapter(this, mContentResolver.query(OnionServiceContentProvider.CONTENT_URI, OnionServiceColumns.getV3_ONION_SERVICE_PROJECTION(), BASE_WHERE_SELECTION_CLAUSE + '1', null, null));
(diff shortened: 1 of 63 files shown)


tladesignz:
That looks like it used https://2019.www.torproject.org/docs/tor-manual.html.en#DisableNetwork once.

n8fr8:
Yes, it used DisableNetwork. The idea was that by default, when we sense there is no network available, that Tor should stop trying to build circuits. This setting allowed people to override that behavior if people wanted.

Not sure why/when it was removed.

bitmold:
I think it's been gone for a very very long time:

- That issue was from over 2 years ago, and it seemingly used somekind of network observer (just basing this off reading changelogs) that I think predates me becoming intimate with this codebase
- Seems like it was added in Orbot 1.0.11:
```
1.0.11
- Tor updated to 0.2.3.23-RC
- ActionBarSherlock updated to latest release
- libevent2 on 2.2.1-alpha
- added data up/down speed and totals display
- added new "Log" SlideDrawer log display 
- added ability to copy logs by touching log (for emailing, debugging)
- fixed iptables Android 4/Jellybean compatible binary detection
- iptables add/delete now on specific rules; no more global FLUSHing
- transproxy ensures that UDP is blocked properly for all or selected apps
- wizard display is centered
- added network connectivity monitor to set Tor's DisableNetwork value
- fixed wizard tips apps recommendations; added DuckGo, Twitter, Firefox Proxymob
- insure that iptables rules are only added once, and that all are cleared
```
- And perhaps removed in `16.1.1-BETA-2-tor-0.4.0.4-rc` with commit d02fb5826d0a2911b97bb80765d45072f97a6085:
```
commit d02fb5826d0a2911b97bb80765d45072f97a6085
Author: n8fr8 <nathan@guardianproject.info>
Date:   Tue Aug 6 16:51:46 2019 -0400

    instead of DisableNetwork, let's try NEWNYM when the network returns
    - we've had issues of losing connectivity setting DisableNetwork, so we are trying a new approach
    - now using DormantClientTimeout settings, to have client go to sleep when it is not used
    - when we receive a network connectivity notification, we call NEWNYM to refresh circuits
```

[Today, in `TorConfig.build` we do this when launching Tor](https://github.com/guardianproject/orbot-android/blob/master/app/src/main/java/org/torproject/android/service/tor/TorConfig.kt#L59). Notably, this isn't behind *any* configuration flags, used or unused. 
```kt
        conf.add("DormantClientTimeout 10 minutes")
        conf.add("DormantCanceledByStartup 1")
        conf.add("DisableNetwork 0")
```

So, unless I'm missing something, it seems like we should drop this setting.

n8fr8:
Sounds good. Glad I wrote that commit message!

PR #1731, opened 2026-06-24
Orbot Targets Android 17 

### LINKED ISSUE #1607: Ensure Orbot Works with Android 17 (API 37 aka Cinnamon Bun)
It's scheduled to be released this spring

https://developer.android.com/about/versions/17/migration

bitmold:
- Target SDK is set to API 37
- App declares `ACCESS_LOCAL_NETWORK` permission 

When the permission is not declared and granted at runtime, using `UPnP.openPortUDP` in Kindness Mode fails at a lower level with `EPERM`:
```
06-24 17:10:42.450  3084  5833 D GatewayFinder: Caused by: android.system.ErrnoException: sendto failed: EPERM (Operation not permitted)
06-24 17:10:42.450  3084  5833 D GatewayFinder:         at libcore.io.Linux.sendtoBytes(Native Method)
06-24 17:10:42.450  3084  5833 D GatewayFinder:         at libcore.io.Linux.sendto(Linux.java:280)
06-24 17:10:42.450  3084  5833 D GatewayFinder:         at libcore.io.ForwardingOs.sendto(ForwardingOs.java:745)
06-24 17:10:42.450  3084  5833 D GatewayFinder:         at libcore.io.BlockGuardOs.sendto(BlockGuardOs.java:413)
06-24 17:10:42.450  3084  5833 D GatewayFinder:         at libcore.io.ForwardingOs.sendto(ForwardingOs.java:745)
06-24 17:10:42.450  3084  5833 D GatewayFinder:         at libcore.io.IoBridge.sendto(IoBridge.java:686)
06-24 17:10:42.450  3084  5833 D GatewayFinder:         ... 3 more
```

1. We don't make that call unless the user has explicitly granted the permission 
2. They grant the permission when they click on the "Restricted" NAT type thing. On API 37+ the dialog shows that the permission was granted, or it explains that it needs to be. 
Screenshot_20260624-184208

The system prompt looks like this:
Screenshot_20260428-212632

After granting it, kindness mode behaves as it did before. The user has the possibility of becoming an unrestricted snowflake. 

If the user continues to deny the permission, the dialog also let's them open the system settings instead. 

Relatedly, there's this PR https://github.com/tladesignz/WaifUPnP-Android/pull/4 
Strips all dependencies out of UPnP that we don't need, compiles it against Android 17, and makes it helpful for debugging in the future by omitting/disabling a lot of the exception strack traces being logged. 

Closes #1607


bitmold:
@n8fr8 @tladesignz @cstiens if the new strings we need for supporting Android 17 look fine as they are, I propose merging just those into master so that translators can get a head start on them while the rest of this is code reviewed. 


```xml

    <string name="kindness_quality_upgrade_need_local_network">On Android 17 and higher, you must also grant the Allow Access to Nearby Devices Permission for Orbot to access your router.</string>
    <string name="kindness_quality_upgrade_perm_granted">The Access Nearby Devices Permission has been granted on this device.</string>
    <string name="grant_permission">Grant Permission</string>
    <string name="set_permission_in_settings">Set Permission in Settings</string>
```

tladesignz (approved):
Ok. Code looks good to me. Quick smoke testing on Android 15 looks just fine. Too late in the day to download Android 17 emulators. I trust you with this, @bitmold!

bitmold:
Thanks <3

PR #1736, opened 2026-06-30
Playstore release tweaks

bitmold:
Small changes/fixes made while trying to get around the play store rejection problem:

- Remove unused `android:hasFragileUserData="false"` from manifest. It defaults to false, so this does nothing. This let's us remove `"tools:suppressApi="29"` from the manifest too, giving us warnings when we do something on older APIs. If set to `true`, this prompts the user to backup their data on app uninstall 
- Remove `android:taskAffinity="'` from manifest. This causes legacy activities in the app (for hosting v3 onion services + v3 client authorization) to create duplicate Orbot's in the app switcher. Very longstanding bug. 
- In french, change "Plus d'information" on the bottom navigation menu to just "Plus". In English, this is the label at the bottom that reads "More". The text gets chopped off even on high resolution phones. "Plus" is legit French for this, and the french string is by far the longest out of _**all**_ the translations we have for this string. 

image
image
image

- **Actually** fix defualt value for screen shot prevention for browser stack issues you were having. I had set the default to `false` in the prefs XML, but the actual Kotlin that retrieved the value still used `true` if the user had never set that preference.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -40,9 +40,16 @@
     </queries>
 
     <!--
-    usesCleartextTraffic is needed for SSDP multicast local router discovery (which is an unencrypted
-    HTTP-style protocol) and for UPnP IGD requests to map ports for Snowflake Proxy, (which are also
-    unencrypted HTTP-style requests).
+        android:allowBackup="false"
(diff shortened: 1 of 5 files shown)


bitmold:
@n8fr8 please get these in before re-releasing to google play

elugo2137-dotcom (approved):

elugo2137-dotcom (approved):

PR #1737, opened 2026-06-30
Fix kindness mode screenshot after UI was reworked for this new release

bitmold:
Rewrote test to integrate with new kindness mode UI @tladesignz and I implemented. 

This all works, but leaving this PR as a draft until the next release goes out...

I am grabbing the new screenshots here and putting them on master, just none of the code changes though.

PR #1742, opened 2026-07-01
Improvement: Kindness Mode UX

tladesignz:
@cstiens created improved designs to explain the little details and problems with Kindness Mode better:

https://github.com/guardianproject/orbot-apple/issues/114#issuecomment-4772751678

Here are the Figma designs:
https://www.figma.com/design/9Jajft5Lbotl6nmPoFAx3v/Orbot-Kindness-Mode?node-id=1-2

I reworked the respective scenes accordingly.

@bitmold, can you do your magic of testing with all the weird devices you have and fix accordingly?


bitmold:
Yeah I'll take a look today/tmrw

bitmold:
@cstiens @tladesignz  I like the new design, but think there are a few considerations/questions I have for Android 
- Regarding the warning "if you are using Tor bridges, you will be temporarily disconnected " 
   - This is true, but I think it's important to emphasize that the users **_entire_** connection to Tor will be disconnected. This means that whatever browsing they're doing in the background will be exposed to their ISP. I think this needs to be phrased clearer so at risk users understand the possible risk they're taking by doing the kindness mode evaluation 
 
- "Kindness mode runs when the device is connected to WiFi"
    - are we altogether disabling using kindness mode with cellular data? Or just defaulting to just using Wi-Fi for Kindness Mode. If it's Wi-Fi as a default but you can use cellular, we should maybe say something like "By default, Kindness Mode runs when the device is connected to Wi-Fi".

- I think there are some misunderstandings with the screen at the bottom.
Screenshot From 2026-07-01 13-15-43
- Kindness Mode **can** run on first use if the VPN is already running on Android. But **if and only if** there is a direct connection to Tor. 
- We **don't** disable Kindness Mode if Orbot is connected to a bridge, provided the user has taken the test and has connected to Tor in the past 24 hours without using a bridge. Users in uncensored countries can prove that they can use Tor without a bridge, and then go back to using bridges if they want to. 

- On the approved/not approved screen, the bottom button shouldn't say "Stop Test" as the test is already finished at this point.
image

- Some nitpicky things. I know this is just a mockup, but in the mockup and throughout the app as it exists today, there are a lot of inconsistencies with terms and spellings that I'd like to see us straighten out over the next releases:
  - "Wi-Fi' is the correct spelling of Wifi/WiFi
  - I think we should consistently use "Kindness Mode", in some places it's "Kindness mode"/"kindness mode" 

I haven't looked at your code yet @tladesignz , just the figma mockups. So not sure if what you're doing here is different than what's in the design spec here. But I'd like to get some of these things cleared up before code reviewing it further

app/src/main/java/org/torproject/android/OrbotApp.kt
```diff
@@ -58,6 +58,8 @@ class OrbotApp : Application() {
             // tell OrbotService it needs to reinstall geoip
             Prefs.isGeoIpReinstallNeeded = true
         }
+
+        Prefs.snowflakeNeedsQualityCheck = true
```
bitmold: Was this a left over from debugging that mistakenly got committed? I did something similar when i was working on this feature. 

Or Do we actually want to reset the quality flag _every_ the app cold starts? I think this is a bad idea, if so and am curious about the rationale.

app/src/main/res/layout/fragment_testing.xml
```diff
@@ -1,175 +1,447 @@
 <?xml version="1.0" encoding="utf-8"?>
-<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
bitmold: I still haven't ran the code yet. But just looking at this from a birds eye view. I think it might be time to break up the fragment into three fragments. I'd rather have smaller fragments with simple layouts for each screen. Than have a huge layout tree, having to deal with hiding / showing screens. And having to have variables like `buttonContinue1`, `buttonContinue2`, `buttonContinue3`. 

I realize the existing `TestingDialogFragment` was already written this way and you're just adapting what's there for the new features. I think a non-trivial amount of time was wasted getting the dialog released by finding/fixing little bugs/edge cases around there constantly being three separate views. It also was a time waste needing to make UI adjustments and hiding/showing `tools:visibility` to see what I was working on. 

I will probably break this up into 3 small fragments. It's much easier that way to immediately understand the  the meaning of a diff to these files (whereas on this PR github hides the changes to this file since it's so big), and also keeps things simpler. 

I'm tempted even to rewrite the simpler two screens in Jetpack Compose (simpler as in the ones that are just basic layouts with text and a button, not the one that actually does the intricate test to establish a tor connection. Probably after getting this merged. It'd be nice to stop introducing more XML code into this project. AFAIK, it's equivalent to that nice SwiftUI you enjoy on iOS

bitmold:
Just making a note here. When all is said and done, I want to make sure that `app/src/androidTest/java/org/torproject/android/ScreenshotKindnessModeFragment.kt`, which was just updated to support an earlier Kindness Mode redesign in #1737 functions as desired with `fastlane screengrab` before merging this PR.

app/src/main/java/org/torproject/android/OrbotApp.kt
tladesignz: Uuups. Yeah, good catch. Of course this is a remnant from development.

app/src/main/res/layout/fragment_testing.xml
tladesignz: I like the idea of splitting up this scene, definitely.

What I don't like is both SwiftUI and Jetpack Compose. Please don't introduce even more dependencies.

tladesignz:
> @cstiens @tladesignz I like the new design, but think there are a few considerations/questions I have for Android
> 
> * Regarding the warning "if you are using Tor bridges, you will be temporarily disconnected "
>   
>   * This is true, but I think it's important to emphasize that the users **_entire_** connection to Tor will be disconnected. This means that whatever browsing they're doing in the background will be exposed to their ISP. I think this needs to be phrased clearer so at risk users understand the possible risk they're taking by doing the kindness mode evaluation

I trust Carries judgement here. We want to stay as simple as possible with the language, without concealing the important details, alright. But trying to "improve" this sentence will ultimately end up in confusing language again.

> * "Kindness mode runs when the device is connected to WiFi"
>   
>   * are we altogether disabling using kindness mode with cellular data? Or just defaulting to just using Wi-Fi for Kindness Mode. If it's Wi-Fi as a default but you can use cellular, we should maybe say something like "By default, Kindness Mode runs when the device is connected to Wi-Fi".

Good point, we should certainly get more precise *here*.

> * I think there are some misunderstandings with the screen at the bottom.
> 
> Screenshot From 2026-07-01 13-15-43 - Kindness Mode **can** run on first use if the VPN is already running on Android. But **if and only if** there is a direct connection to Tor. - We **don't** disable Kindness Mode if Orbot is connected to a bridge, provided the user has taken the test and has connected to Tor in the past 24 hours without using a bridge. Users in uncensored countries can prove that they can use Tor without a bridge, and then go back to using bridges if they want to.

IDK what your explanations and the image shown above have to do with each other. Nevertheless:

We *do* want to stay as simple as possible. We can sacrifice a little precision here. This benefits both the language *and* the code. Please refrain from overcomplicating the code again by injecting all sorts of extra texts like you did before.

> * On the approved/not approved screen, the bottom button shouldn't say "Stop Test" as the test is already finished at this point.

I assumed this to be a copy-past error and didn't use that in the actual code.

> image
> * Some nitpicky things. I know this is just a mockup, but in the mockup and throughout the app as it exists today, there are a lot of inconsistencies with terms and spellings that I'd like to see us straighten out over the next releases:
>   
>   * "Wi-Fi' is the correct spelling of Wifi/WiFi
>   * I think we should consistently use "Kindness Mode", in some places it's "Kindness mode"/"kindness mode"

D'accord with streamlining to "Wi-Fi" and "Kindness Mode".

bitmold:
@tladesignz 

> IDK what your explanations and the image shown above have to do with each other. Nevertheless:
> We do want to stay as simple as possible. We can sacrifice a little precision here. This benefits both the language and the code. Please refrain from overcomplicating the code again by injecting all sorts of extra texts like you did before.

This isn't over complicated, I think you're misunderstanding something. The image I posted includes a warning label that reads  "Kindness Mode is disabled when Orbot VPN is connected with a bridge". **This is wrong.**

We:
- run a connection test to make sure you can connect to tor without a bridge before enabling kindness mode
- once you pass the test, you can go back to using tor with a bridge


This is implemented this way today. Are we to remove the feature and make it so that, even after passing the connection test, we disable kindenss mode if a user is using obfs4/etc for whatever reason. _Why?_ 

The label just shouldn't be there. It's not a real warning. The only time "using a bridge" is at all relevant is when you are doing the connection test, afterwards users are free to both use bridges with Tor VPN and be a snowflake proxy. Of course, they may have to retake the test in the future. 

If you want to add this label, you would actually have to write more code, aka adding more complexity, to make sure that we turn off kindness mode when the user who passed the test decides to use obfs4/any other transport.

tladesignz:
> If you want to add this label, you would actually have to write more code, aka adding more complexity, to make sure that we turn off kindness mode when the user who passed the test decides to use obfs4/any other transport.

Ok. Remove it.

bitmold:
I did do the thing of splitting the dialog into 3 dialogs. The amount of nesting here was both computationally expensive and  very overcomplicated when dealing with the UI / understanding the code...

Found and a handful of issues with this implementation along the way:

- I couldn't get this to work on a fresh install without no active tor connection. I think some of the initial testing dialog logic had been broken along the way somehow...
- The 2nd dialog, with the confirmation toggle, has its text get cut off by the outer dialog padding. ie It says "connect to the \nnetwork" not "connect to the\nTor network"
- the text next to each icon list item was not vertically aligned inconsistently depending on the text length of each item. this was resolved cleanly with **one** simple `ConstraintLayout` instead of some combo of `LinearLayout`s and `ConstraintLayout`s
-  the new dismiss buttons never unbound the testing service's tor receiver from the Activity...

bitmold:
@tladesignz Give it a go when you can. I want to do a final round of testing later, but I'm satisfied with this as of now. 

Also leaving a note for myself here: weblate will need to be refreshed after this is merged.

tladesignz:
I fixed landscape UI for `fragment_testing_initial` and `fragment_testing_confirm`.

Please review, @bitmold. If you don't find anything problematic anymore, I think we're good to merge!

PR #1748, opened 2026-07-05
request ACCESS_LOCAL_NETWORK when enabling open proxy on all interfaces

### LINKED ISSUE #1743: open proxy on all interface breaks on Android 17
It also can require the `ACCESS_LOCAL_NETWORK` permission. Which we only grant so far in kindness mode... Low priority but needs to be fixed.

munzzyy:
Fixes #1743

The "Open Proxy on All Interfaces" setting makes Tor bind its SOCKS port to `0.0.0.0` (`TorConfig.kt`), which on Android 17 needs the `ACCESS_LOCAL_NETWORK` permission. Right now we only ever request that permission from the kindness mode UPnP dialog, so flipping this setting on an API 37 device just produces a proxy that nothing on the LAN can actually reach - no prompt, no error, it just doesn't work.

What this does:

- Moved `needsAccessLocalNetworkPermission()` out of `UPnPDialogFragment` and into `NetworkUtils`, so it's shared instead of duplicated. `UPnPDialogFragment` and `SnowflakeProxyWrapper` now call the shared version.
- `SettingsPreferenceFragment` now hooks the "Open Proxy on All Interfaces" checkbox: turning it on triggers a permission request first (same grant / deny / "permanently denied -> app settings" flow the kindness dialog already uses), and the checkbox only gets checked once the permission is actually granted.
- `TorConfig.build()` falls back to a localhost-only `SOCKSPort` if the preference is on but the permission isn't (e.g. revoked after the fact in system settings), so we never emit a `0.0.0.0` config nothing can reach.
- One new string for the permission rationale toast.

Verified by reading through the existing kindness-mode permission flow and mirroring it exactly, and by checking every call site of the old `UPnPDialogFragment.needsAccessLocalNetworkPermission` (there were two: the dialog itself and `SnowflakeProxyWrapper`) to make sure both got updated. I don't have a Java/Gradle toolchain in this environment so I couldn't build or run this - I'd appreciate someone testing the actual permission dialog on an API 37 emulator/device before merge. I also haven't tested the LAN end-to-end (a second device actually connecting through the opened proxy), just that the config and permission flow line up.

### DIFF
--- a/app/src/main/java/org/torproject/android/util/NetworkUtils.kt
+++ b/app/src/main/java/org/torproject/android/util/NetworkUtils.kt
@@ -1,10 +1,14 @@
 package org.torproject.android.util
 
+import android.Manifest
 import android.content.Context
+import android.content.pm.PackageManager
 import android.net.ConnectivityManager
 import android.net.NetworkCapabilities
 import android.net.VpnService
+import android.os.Build
 import android.util.Log
+import androidx.core.content.ContextCompat
 import java.net.InetSocketAddress
 import java.net.Socket
 
@@ -76,5 +80,28 @@ object NetworkUtils {
         } catch (_: Exception) {
             return false
         }
+    }
+
+    /**
+     * Checks whether ACCESS_LOCAL_NETWORK still needs to be requested from the user. Any feature
+     * that talks to devices on the LAN (kindness mode's UPnP probing, or the SOCKS/HTTP proxy
+     * listening on all interfaces) needs this permission on Android 17+.
+     *
+     * Returns null on Android 16 and below, where the permission doesn't exist.
+     */
+    fun needsAccessLocalNetworkPermission(context: Context): Boolean? {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.CINNAMON_BUN) {
+            return null
+        }
+        val checkAccessLocalNetworkPerm = ContextCompat.checkSelfPermission(
+            context,
+            Manifest.permission.ACCESS_LOCAL_NETWORK
+        )
+        if (checkAccessLocalNetworkPerm == PackageManager.PERMISSION_DENIED) {
+            Log.d(TAG, "local network permission not granted")
+            return true
+        }
+        Log.d(TAG, "local network permission granted")
+        return false
     }
 }
--- a/app/src/main/java/org/torproject/android/service/tor/TorConfig.kt
+++ b/app/src/main/java/org/torproject/android/service/tor/TorConfig.kt
@@ -24,7 +24,13 @@ object TorConfig {
         val isolate = getIsolation()
         val ipv6Pref = getIpv6()
 
-        if (Prefs.openProxyOnAllInterfaces()) {
+        // on API 37+ this also needs ACCESS_LOCAL_NETWORK. If the pref is on but the permission
+        // got revoked (or was never granted), fall back to a localhost-only SOCKSPort instead of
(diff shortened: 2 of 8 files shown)


app/src/main/java/org/torproject/android/util/NetworkUtils.kt
```diff
@@ -77,4 +81,27 @@ object NetworkUtils {
...
     }
+
+    /**
+     * Checks whether ACCESS_LOCAL_NETWORK still needs to be requested from the user. Any feature
+     * that talks to devices on the LAN (kindness mode's UPnP probing, or the SOCKS/HTTP proxy
+     * listening on all interfaces) needs this permission on Android 17+.
+     *
+     * Returns null on Android 16 and below, where the permission doesn't exist.
+     */
+    fun needsAccessLocalNetworkPermission(context: Context): Boolean? {
```
meenbeese, line 92: Why nullable? Simply return false?

munzzyy: Good question. It's not defensive, the null actually means something different than false here.

null = permission doesn't exist on this OS (pre-API-37), false = permission exists and is granted. `TorConfig` and `SnowflakeProxyWrapper` don't care about the difference and just check `!= true`, but `UPnPDialogFragment` does:

```kotlin
val accessLocalNetworkNeeded =
    NetworkUtils.needsAccessLocalNetworkPermission(requireContext())
        ?: return builder.createWithCurves()
```

On null it bails out early with the plain dialog, no permission section at all. If I collapsed this to false, that same device would instead take the `else` branch and show "permission granted" text for a permission that doesn't exist on that OS version. So the tri-state is intentional, not an oversight, happy to add a comment above the function if that makes the intent clearer.

bitmold: Yeah, it returns `null` in the DialogFragment. Great choice on moving this to the NetworkUtils. We didn't actively think about this on the open proxy on all interfaces setting. Thank you for this PR. Reviewing and merging soon. <3

bitmold:
Rebased with `master` and reviewing now...

bitmold:
@munzzyy **Thank you so much** for implementing this and for making your first contribution to Orbot! This was a very high quality PR and I'm very grateful for it. :star: 

Normally, I'd ask you to make some changes to this PR and provide feedback, but since we have a release coming out very soon - I made them myself. All of your core logic was solid, but I made the following changes/refactors:

- Used an `AlertDialog` instead of a `Toast` when using the open proxy on all interfaces setting. You couldn't see the whole message as a `Toast`, and it disappears before showing the unfortunately very vague and scary "access nearby devices" system prompt Android 17 uses for local network access
- Unset the preference in Orbot's settings UI if the user revoked it 
- `Prefs.openProxyOnAllInterfaces` now takes a `Context` and checks for the permission if API 37+. I moved your logic for this from `TorConfig` to `Prefs`. 
-  Moved the now duplicated logic to open Orbot's settings to one place

I'm definitely open to any future improvements around this if you have further suggestions/opinions. 

Thanks again for making Orbot better :heart:

munzzyy:
Late thanks for this — I only just caught up on the thread. Makes sense on all of it, especially the AlertDialog over the Toast (the message really was too long to read as a toast) and unsetting the pref when the user revokes the permission, which I'd missed entirely. Moving the API 37+ check into `Prefs` with a `Context` is the better home for it too. Appreciate you doing the changes yourself to make the release.

Issue #1695, opened 2026-05-14
[BUG] dnstt bridge lines loose formatting and require a manual edit


ebanaam:
**Describe the Bug**
At Tor, we are sharing dnstt Tor Bridge lines over our email helpdesk and we have received some reports that copying the bridge line from the email into Orbot is resulting in 'Invalid bridge format' error.
For some context, we use html styling (basically enclosing the entire bridge line `<code>`) so as there are no line-breaks. Other bridge lines with obfs4, WebTunnel are not affected.

We have tried a few ways to test if we can optimize how the bridge lines are reaching the user via email, but none of it seems to work (more details on this [bug report](https://gitlab.torproject.org/tpo/community/support/-/work_items/40257)).

**To Reproduce**
1. Get dnstt bridge lines via Email and copy them to your clipboard.
2. Launch Orbot -> tap on 'Set Transport: Custom Bridges' -> Select 'Custom Bridges' -> Tap on 'Next' -> Paste the dnstt bridge lines.
3. See 'Invalid bridge format' error.
4. Delete the existing 'spaces' between characters and re-introduce the spaces manually and the issue disappears.

**Expected Behavior**
Step 4. above should not be required.

**Orbot:**
Version: Orbot for Android v17.9.4 BETA 1 (tor 0.4.9.6.2)

ebanaam:
Example dnstt bridge lines that we are seeing this issue with:

```
dnstt 192.0.2.4:1 A998F319ADB60EE344540EC4B21524CC484F96BE udp=8.8.8.8:53 pubkey=241169008830694749fe96bb070c4855c5bb5b9c47b3833ed7d88521ba30a43f domain=t.ruhnama.net
dnstt 192.0.2.4:2 80EEFA4F4875ED2B7B5A86DF2D7588AD32E29F15 udp=8.8.8.8:53 pubkey=a2fb71077eeaa54a02cda7a90be306af5d299ab21822a8b277d4eacbc9168631 domain=t2.bypasscensorship.org
dnstt 192.0.2.4:3 74D409BED3E2F881F365543A72C8F079CB84FFEB udp=8.8.8.8:53 pubkey=c596c458fc3453dc40903ab235f5854a2609831075640c4c5584f76de05b8271 domain=t.arkadag.org
```

bitmold:
@ebanaam 
[...]

PR #1780, opened 2026-08-02
fix Bridge.doh getter reading the dot parameter

munzzyy:
`Bridge.doh` calls `getPiece("dot")`. Copy paste slip from the `dot` getter below it, shipped in #1629. A `doh=` line parses to null and a Builder round trip drops the resolver. A `dot=` line leaks its value into `doh` and the rebuilt line grows a bogus duplicate.

No Android toolchain here so I compiled the untouched `Bridge.kt` with kotlinc 2.4.10 plus kotlinx-serialization-core 1.11.0 and fed it the first `BuiltInBridges.dnsttBridges` line. On master:

```
doh line -> bridge.doh = null
doh line round trip    = dnstt 192.0.2.4:1 A998F319ADB60EE344540EC4B21524CC484F96BE pubkey=241169008830694749fe96bb070c4855c5bb5b9c47b3833ed7d88521ba30a43f domain=t.ruhnama.net
dot line -> bridge.doh = dns.google:853
dot line round trip    = dnstt 192.0.2.4:1 A998F319ADB60EE344540EC4B21524CC484F96BE doh=dns.google:853 dot=dns.google:853 pubkey=241169008830694749fe96bb070c4855c5bb5b9c47b3833ed7d88521ba30a43f domain=t.ruhnama.net
```

With the patch:

```
doh line -> bridge.doh = https://dns.google/dns-query
doh line round trip    = dnstt 192.0.2.4:1 A998F319ADB60EE344540EC4B21524CC484F96BE doh=https://dns.google/dns-query pubkey=241169008830694749fe96bb070c4855c5bb5b9c47b3833ed7d88521ba30a43f domain=t.ruhnama.net
dot line -> bridge.doh = null
dot line round trip    = dnstt 192.0.2.4:1 A998F319ADB60EE344540EC4B21524CC484F96BE dot=dns.google:853 pubkey=241169008830694749fe96bb070c4855c5bb5b9c47b3833ed7d88521ba30a43f domain=t.ruhnama.net
```

Both lines now round trip byte for byte. To be fair nothing hits this today because `getUdpDnstt()` nulls both params before rebuilding. It only bites the next consumer of `Bridge.doh`. Want a follow up that adds a small parser test around these getters?

### DIFF
--- a/app/src/main/java/org/torproject/android/service/circumvention/Bridge.kt
+++ b/app/src/main/java/org/torproject/android/service/circumvention/Bridge.kt
@@ -188,7 +188,7 @@ class Bridge(var raw: String) {
         get() = getPiece("udp")
 
     val doh
-        get() = getPiece("dot")
+        get() = getPiece("doh")
 
     val dot
         get() = getPiece("dot")


bitmold:
@tladesignz this seems like something we let slip by 

Also, I wonder if this somehow relates to problems we were having in #1695 - an issue I was never able ot reproduce 

thank you @munzzyy

tladesignz:
> @tladesignz this seems like something we let slip by

Obviously a copy-paste error i made. I copied this code from iOS. The bug is in there, too. So, again, @munzzyy, thanks a ton for catching this!

> Also, I wonder if this somehow relates to problems we were having in #1695 - an issue I was never able ot reproduce

I highly doubt that, since as far as I can see, there is no round trip through the `Bridge` class, when using a custom transport:

1. Storing to preferences as `List<String>`:

https://github.com/guardianproject/orbot-android/blob/master/app/src/main/java/org/torproject/android/ui/connect/CustomBridgeBottomSheet.kt#L113-L122

2. Reading from preferences as `List<String>`. No modifications.

https://github.com/guardianproject/orbot-android/blob/master/app/src/main/java/org/torproject/android/service/circumvention/Transport.kt#L256-L260

However, we do a ton of string cleaning (and a validity check, which is… difficult) when reading and writing that `Prefs.bridgesList` property.

munzzyy:
Nice, glad it was useful on both platforms. I see you already patched IPtProxyUI in 5dd649b.

The offer of a parser test still stands. A round trip check over the BuiltInBridges lines would have caught this one and it is cheap to keep around. I can do the Kotlin side here, and the Swift side over in IPtProxyUI if you want it there too.

PR #1791, opened 2026-08-28
normalize unicode spaces in custom bridge input

### LINKED ISSUE #1695: [BUG] dnstt bridge lines loose formatting and require a manual edit

munzzyy:
Fixes #1695

HTML email (Tor's dnstt bridge templates in particular) swaps in a non-breaking space where a regular space belongs once the bridge line gets copied out. validBridgeRegex uses \s. That's ASCII-only in Java/Kotlin regex. Bridge.kt splits on a literal " ". A stray U+00A0 anywhere in the line fails validation, or gets saved and parses wrong later.

This normalizes any Unicode space character to a regular space before we validate or save what's pasted into the custom bridges field. While in there it also puts validation and saving through one cleanBridgeLines path: before this, isValidBridge split with lines() and dropped blanks while the save handler split on a bare \n with no filter, so a CRLF paste could validate fine but store lines with a trailing \r, and blank separator lines got stored as empty entries.

Added unit tests for NBSP-contaminated obfs4 and vanilla lines, CRLF input, blank-line dropping, and a no-regression case for plain ASCII, matching BridgeTest.kt's style.

No Android SDK on this box, so no full gradle build. Verified the regex and normalization logic standalone with kotlinc: the NBSP line fails against the unmodified regex, passes after the fix, plain ASCII behaves exactly as before, and cleaned CRLF input comes out with no \r.

### DIFF
--- a/app/src/main/java/org/torproject/android/ui/connect/CustomBridgeBottomSheet.kt
+++ b/app/src/main/java/org/torproject/android/ui/connect/CustomBridgeBottomSheet.kt
@@ -38,12 +38,23 @@ class CustomBridgeBottomSheet : OrbotBottomSheetDialogFragment() {
                     + """[ \t\f\w\-/+:=.,\[\]]*$""" // Optional bridge arguments (different per bridge type, subject to change)
         )
 
+        // HTML email (e.g. the Tor Project's bridge distribution templates) can substitute a
+        // non-breaking space or another Unicode space variant for a regular space when a bridge
+        // line is copied out. \s in validBridgeRegex above only matches ASCII whitespace, and
+        // Bridge.kt splits on a literal " ", so those bridges get rejected or parsed wrong.
+        private val unicodeSpaceRegex = Regex("\\p{Zs}")
+
+        // One cleaning path for validation and saving: normalize any Unicode space to a regular
+        // space, split into lines (which also handles \r\n from desktop pastes) and drop blanks,
+        // so exactly the lines that were validated are the lines that get stored.
+        fun cleanBridgeLines(input: String): List<String> {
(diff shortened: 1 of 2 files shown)


meenbeese (commented):
No need for two separate `normalizeBridgeInput` and `cleanBridgeLines` here.

munzzyy:
good call, folded it into one cleanBridgeLines. it was only ever called from there anyway. pushed.

bitmold:
This is amazing, thank you!

PR #1800, opened 2026-09-01
Fix ContentProvider update and use consistent APIs

meenbeese:
- Fix per-id update selection typo in ClientAuthContentProvider (id_ → _id)
- Simplify ClientAuthActivity: drop redundant fragment result-forwarding loop and unconditional FLAG_SECURE override, prune unused imports
- Modernize ClientAuthBackupDialogFragment: switch SAF export from deprecated startActivityForResult to ActivityResultContracts.CreateDocument; remove onActivityResult, REQUEST_CODE_WRITE_FILE, and unused imports

### DIFF
--- a/app/src/main/java/org/torproject/android/ui/v3onionservice/clientauth/ClientAuthActivity.java
+++ b/app/src/main/java/org/torproject/android/ui/v3onionservice/clientauth/ClientAuthActivity.java
@@ -12,12 +12,10 @@ import android.os.Looper;
 import android.provider.OpenableColumns;
 import android.view.Menu;
 import android.view.MenuItem;
-import android.view.WindowManager;
 import android.widget.ListView;
 import android.widget.Toast;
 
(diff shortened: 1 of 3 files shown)


bitmold:
The update selection typo was a good catch, but that reassignment of `selection` is never actually run at runtime. Onion service client authorizaiton would always update to be enabled/disabled. Still, I committed a fix crediting you on this

bitmold:
The unconditional FLAG SECURE setting was intentional, I added a comment. 

If you want to rebase this, and just make a PR for using a non-deprecated activity result contract I would review that. Please demonstrate that it works. As I had said before, changing these small onion service java classes is not a priority of mine - these are important files and I don't want to uncarefully modify them at the risk of breaking functionality -  _but,_ I would be open to that small change in isolation .

meenbeese:
I think this PR is resolved, thanks.

PR #1802, opened 2026-09-03
Reset weekly snowflake count lazily by week

meenbeese:
The only use of androidx.work was a PeriodicWorkRequest that reset Prefs.snowflakesServedWeekly to 0 once every 7 days. This is overkill for resetting a single integer, adds a dependency (plus WorkManager's auto-registered initializer/job-service manifest components), and can miss runs due to Doze, force-stop, or the device being offline for a week.


bitmold:
This is a welcome change!  I may not have time to test it for another week or so, but appreciate removing this dependency for something that's simple

app/src/androidTest/java/org/torproject/android/ScreenshotKindnessModeFragment.kt
```diff
@@ -23,7 +23,7 @@ class ScreenshotKindnessModeFragment : BaseScreenshotTest() {
     fun setupPrefs() {
 
         repeat(5) { Prefs.addSnowflakeServed() }
-        Prefs.resetSnowflakesServedWeekly()
+        Prefs.refreshWeeklyServedIfNeeded()
```
bitmold, line 26: This breaks the screengrab test getting a screenshot of the kindness mode screen 

Since the new method doesn't force a rewrite of the weekly prefs, then the screenshot with a a clean weekly (5) and all time total (10) doesn't get taken correctly

Issue #1783, opened 2026-08-04
[Question] Kindness mode turns off after a certain time period


Am0rphous:
**Describe Your Question**
Hi, after enabling Kindness mode, Orbot turns itself off automatically after some period. This happens at irregular intervals and could be after one or several hours. I'm running Orbot on a profile which runs in the background. When checking Orbot I have to enable Kindness mode again. 

**Additional Context**
Fully updated GrapheneOS + Orbot.

Does anyone else experience this?

bitmold:
as far as i can tell, this does not seem to be  a GrapheneOS specific issue

bitmold:
Seemingly closed with #1807 ...

PR #1807, opened 2026-09-07
Kindness mode: stop the stopped proxy, survive being killed, clean up UPnP

### LINKED ISSUE #1183: [BUG] 20% CPU overhead in kindness mode
As soon as the kindness mode is on, Orbox starts a process that uses about 20% of the CPU and cannot be turned off. Not even if the mode is turned off. I don't think that's intended. It helps to end the process or the app.

IMG_20240920_122745
Xiaomi POCO X3 pro

### LINKED ISSUE #1795: Stale UPnP Mappings via In-Memory Tracking
Up to three UDP port mappings are attempted on the local router through UPnP during
normal Snowflake startup.

`SnowflakeProxyWrapper.enableProxy()` selects a random starting port and attempts to
map three consecutive UDP ports through `UPnP.openPortUDP()` when local-network
access is available under the platform permission model.

Mapping cleanup is implemented by `releaseMappedPorts()`, which closes the ports
recorded in `mappedPorts`. The mapped port numbers are retained only in memory. If the
process terminates before cleanup is completed, the mapped port numbers are lost and
no persisted record is available to the shown cleanup logic on the next launch. The
lifetime of a stale mapping depends on the router implementation and the UPnP lease
configuration.
 
The security impact is limited, but router forwarding state can remain after an abnormal
process termination without being tracked for cleanup on the next launch.

```kt
@Synchronized​
fun enableProxy() {​
if (proxy != null) return​
CoroutineScope(Dispatchers.IO).launch {​
val start = Random.nextInt(49152, 65536 - 2)​
if (NetworkUtils.needsAccessLocalNetworkPermission(service) != true) {​
for (port in (start..start + 2)) {​
if (UPnP.openPortUDP(port, OrbotConstants.TAG)) {​
mappedPorts.add(port)​
}​
}​
​
if (mappedPorts.size < 3) {​
releaseMappedPorts()​
}​
}​
[...]​
}​
}​
```

### LINKED ISSUE #1799: [BUG] Kindness mode turning off and no connections
**Describe the Bug**
Since the last update the Snowflake proxy turns off after a while and doesn’t get connections. Battery optimizations are off, I haven’t changed anything and it worked fine for months. The proxy shows up as “restricted” but upnp has always been switched off in my router.

**To Reproduce**
1. Turn Kindness Mode on
2. Wait a few hours
3. The main vpn keeps running, but Kindness mode turns off. Total number of connections doesn’t increase even when it runs.

**Expected Behavior**
Snowflake proxy remains active and provides connections 

No custom configuration 

**Screenshots**
N/A

- Device: moto g9
- OS: Android 11
- Version: v17.9.5-RC-4-tor-0.4.9.11

**Crash Logs (Advanced)**

**Additional Context**

munzzyy:
Kindness mode has a habit of quietly not being on anymore. This traces the three ways that happens and closes each of them.

**The stopped proxy that keeps running (#1183).** `enableProxy()` takes its lock, then does the real work in a coroutine that has no lock at all. A `stopProxy()` that lands before that coroutine runs sees `proxy == null`, does nothing, and the coroutine then starts a proxy nothing owns. The service stops and restarts the proxy on every network callback, so this window gets hit in normal use, and an orphaned proxy is a plausible source of the CPU drain reported there. Starts now carry a generation number: if a stop or a newer start won while the ports were being mapped, the late start closes its own mappings and never calls `start()`.

There is a second copy of the same race one layer down. IPtProxy's `Start()` flips its `isRunning` flag inside a goroutine it spawns, and its `Stop()` is a no-op until that flag is set, so a fast stop is silently ignored and the proxy comes up anyway:

```go
go func() {
    sp.isRunning = true
    err := sp.proxy.Start()
    ...
}()

func (sp *SnowflakeProxy) Stop() {
    if sp.isRunning {
        ...
```

That matches cohosh's shutdown race hypothesis in the issue. The app cannot fix the library, but it can stop dropping its reference too early: after `stop()`, the wrapper watches the instance for two seconds and stops it again if it came up late. The remainder, if any, is upstream in IPtProxy.

**The service that stays dead (#1799, #1783).** START_STICKY is the only thing that brought the service back, and sticky restarts get throttled, abandoned after repeated kills, and never delivered at all on some builds. The only re-arm paths were opening the app or toggling the switch, which is why people find the proxy off hours later. The pref is already the user's standing intent, so it gets honored now: `OnBootReceiver` starts the service after a reboot when kindness mode is on, and a 15 minute WorkManager watchdog (the dependency was already in the app) restarts it when it is wanted, allowed in this region, and not running. The watchdog cancels itself when the user turns kindness mode off. A background start denial on 31+ is caught and left for the next boot or app open. This will not resurrect the service inside a stopped GrapheneOS work profile, nothing can, but within a running profile the watchdog covers the kill-and-forget case.

**The port mappings that outlive the process (#1795).** Mapped UPnP ports lived in one in-memory list, so any abnormal termination left them open on the router until the lease expired, with nothing on the next launch knowing they existed. The mapping is written to prefs before the proxy starts, cleared on clean release, and `releaseStalePorts()` closes whatever a dead run left behind on the next service start.

Fixes #1183, fixes #1799, fixes #1795. Helps #1783 within what the platform allows.

Compatible with #1805: this deliberately stays out of `initNetworkCallbacks()`, so both apply cleanly in either order.

Verified with the unit suite (eleven new tests: the start-commit and watchdog decision tables, and the port record parsing) and `assembleFullpermDebug`, both clean. I do not have a device farm, so the multi-hour soak that #1799 describes is exactly the test this needs from someone who does. Happy to adjust anything.

One thing I noticed but left alone, since it is outside these issues: the preference listener in `SnowflakeProxyService.onCreate` returns early unless a key equals two different constants at once, so the region kill switch and the camo refresh there look unreachable. Can file separately if useful.


meenbeese:
Thanks for the PR, but we are looking to drop the AndroidX Work in #1802. It would be great if you could look at that PR and rework your approach here.

munzzyy:
Thanks @meenbeese, that makes sense for the counter reset. A PeriodicWorkRequest is a lot of machinery to zero one integer, and dropping the dependency for that is fair.

The watchdog here does a different job: re-arming Kindness Mode after the service gets killed while the user still wants it running. START_STICKY on its own wasn't enough in testing (that's the note in KindnessWatchdogWorker), so it ended up as a Worker.

If the plan is to drop androidx.work across the board, what would you want for the periodic re-arm instead, an AlarmManager setExactAndAllowWhileIdle loop? I'll rework this around whatever the project lands on, as long as it keeps surviving a kill. Reading through #1802 now.

meenbeese:
The plan is to drop AndroidX Work at the moment. Can you look at JobScheduler or a simple foreground service as replacement options? I didn't really look at the code that much but one of those should work here.

munzzyy:
Makes sense. JobScheduler is a clean fit, since that's what WorkManager runs on underneath, so I can drop the androidx.work dependency and keep the same survive-a-kill behavior.

One note on the foreground-service option: the watchdog exists to come back after the service is already gone (force-stop, OEM battery killers), and a plain foreground service doesn't get re-armed once it's killed. That's the START_STICKY gap the Worker was covering, so I'll go the JobScheduler route rather than a bare service.

I'll rework it onto JobScheduler and push an update here.

munzzyy:
Pushed the JobScheduler version (2cc1be6). The watchdog is a JobService now, a 15-minute periodic job with setPersisted so it still comes back after a reboot, and the same shouldRestart logic and tests carried over.

I left androidx.work in the build for now, since the weekly snowflake-count reset still uses it. Once #1802 lands and that goes away, the dependency drops out entirely. Compiles clean and the unit tests pass.

meenbeese (commented):
LGTM

bitmold:
This one is exciting, but a bit involved - I'll test and review this soon. If this works as well as you think it does, and touches all of these longstanding issues you referenced above, then this is a very special PR. Thanks again <3

bitmold:
This is great work. I extended it a bit, merged it to `master`, and then added some stability tweaks of my own to kindness mode.

It's easy to see that it fixes:
- the issue with the possible stale UPnP ports (#1795 )
- and it seems to fix both #1783 and #1799 (I'm pretty certain that these are duplicates / GrapheneOS has little to do with that bug...) 

I am inclined to agree with your theory about #1183 above, but I don't feel comfortable closing that issue yet. Until we get a lot more testing and various other people measuring and observing their CPU usage.

munzzyy:
Thanks for testing and merging it. Glad #1183 closed with it, that drain had been open a while. When #1802 lands I will send the small follow-up that drops androidx.work, as we talked about above.

Issue #1404, opened 2025-08-25
[BUG] java.lang.IllegalArgumentException: Service not registered: org.torproject.android.service.OrbotService$2@30897fe


kasia-de:
**Describe the Bug**
`java.lang.IllegalArgumentException: Service not registered: org.torproject.android.service.OrbotService$2@30897fe`

**To Reproduce**

**Expected Behavior**
Not to crash.

**What Custom Configuration Do You Use?**

**Screenshots**

**Smartphone (please complete the following information):**
- Device: Pixel Tablet
- OS: Android 16
- Version: 1750300102

**Crash Logs (Advanced)**
```
java.lang.IllegalArgumentException: Service not registered: org.torproject.android.service.OrbotService$2@30897fe
	at android.app.LoadedApk.forgetServiceDispatcher(LoadedApk.java:2027)
	at android.app.ContextImpl.unbindService(ContextImpl.java:2390)
	at android.content.ContextWrapper.unbindService(ContextWrapper.java:953)
	at org.torproject.android.service.OrbotService.stopTor(OrbotService.java:338)
	at org.torproject.android.service.OrbotService.stopTorAsync(OrbotService.java:241)
	at org.torproject.android.service.OrbotService.stopTorOnError(OrbotService.java:259)
	at org.torproject.android.service.OrbotService.startTor(OrbotService.java:600)
	at org.torproject.android.service.OrbotService.-$$Nest$mstartTor(Unknown Source:0)
	at org.torproject.android.service.OrbotService$IncomingIntentRouter.run(OrbotService.java:1069)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1156)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:651)
	at java.lang.Thread.run(Thread.java:1119)
```

**Additional Context**

kasia-de:
```
07-06 16:22:28.369 22201 22253 E AndroidRuntime: FATAL EXCEPTION: pool-3-thread-1
07-06 16:22:28.369 22201 22253 E AndroidRuntime: Process: org.torproject.android, PID: 22201
07-06 16:22:28.369 22201 22253 E AndroidRuntime: java.lang.IllegalArgumentException: Service not registered: org.torproject.android.service.OrbotService$2@30897fe
[...]

PR #1809, opened 2026-09-09
Fix log lines being stored in duplicate

munzzyy:
Every log line can end up stored two or more times, and the doubling survives closing and reopening the log sheet. n8fr8 mentioned in the Matrix room yesterday that this one was never pinned down, so I went digging.

The duplication is in storage, not display. startTor() guards re-entry with `torServiceConnection != null && conn != null`, but conn stays null for the whole bind window (longer with Smart Connect), and every ACTION_START runs through the cached thread pool. Two start intents inside that window, say the boot receiver plus an always-on VPN system start, both pass the guard, so the service binds twice and each onServiceConnected() adds its own OrbotRawEventListener to the same TorControlConnection. jtorctl's addRawEventListener is a plain List.add, so from then on every tor event is appended to the stored log once per listener. That is why the doubled lines survive the sheet being reopened and survived the UI rewrite. stopTor() also never nulled torServiceConnection, so the stacking can compound across restarts; the "Service not registered: OrbotService$2" crash in #1404 looks like the same stale state.

I proved the listener half on the JVM against jtorctl 0.4.5.7: one control connection, two registered listeners, one 650 NOTICE in, two identical appends out. The new unit tests drive a real TorControlConnection over piped streams and fail on the old logic.

To see it on a device: enable start on boot plus always-on VPN and reboot. "Connected to the Tor control port" shows up twice and every line after it is doubled.

What this changes:

- startTor() is single flight across the bind window. The flag is released on every path that kills the attempt, including a bind that returns false and a binding the system reports dead, so a failed start cannot wedge the service into ignoring later starts.
- stopTor() nulls torServiceConnection.
- onServiceConnected() removes the stale raw event listener before adding the fresh one, so even a bind path the guard does not cover cannot log an event twice.
- The log sheet only appends the stored-log delta when the stored text actually extends what is on screen, and re-reads the stored log when it returns to the foreground. A backgrounded sheet across a service restart used to splice stale text once the new log outgrew the old.

All 28 unit tests pass, and the two new suites fail against the unfixed logic.

### DIFF
--- a/app/src/main/java/org/torproject/android/service/OrbotService.java
+++ b/app/src/main/java/org/torproject/android/service/OrbotService.java
@@ -64,6 +64,7 @@ import androidx.core.app.NotificationCompat;
 import androidx.core.app.ServiceCompat;
 import androidx.core.content.ContextCompat;
 
+import net.freehaven.tor.control.RawEventListener;
 import net.freehaven.tor.control.TorControlCommands;
 import net.freehaven.tor.control.TorControlConnection;
 
@@ -86,6 +87,7 @@ import java.util.Locale;
 import java.util.StringTokenizer;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import kotlin.Unit;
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/torproject/android/service/OrbotService.java
```diff
@@ -232,6 +235,7 @@ private void stopTorAsync(boolean showNotification) {
 
     private void stopTorOnError(String message) {
         //  stopTorAsync(false);
+        torStartRequested.set(false);
```
bitmold, line 238: Is there a reason why you only say `torStartRequest.set(false)` in `stopTorOnError` while not setting the atomic bool in `stopTor`?

bitmold: Nevermind, I see that you do on line 267

PR #1810, opened 2026-09-09
Fix Kindness mode preference listener

### LINKED ISSUE #1808: Kindness mode keeps running after the bridge country changes
`SnowflakeProxyService`'s preference listener returns for every key, so its body never runs:

```kotlin
if (key != Prefs.PREF_BRIDGE_COUNTRY || key != Prefs.PREF_CAMO_APP_PACKAGE) return@OnSharedPreferenceChangeListener
```

A key matches at most one of the two. One side of the `||` always holds, including for those two keys.

So the bridge country branch never reaches `stopSelf()`. A running proxy keeps running after the country is switched to one `Regionalization.isKindnessModeDisabledForCountry` rejects. `refreshNotification()` also never fires on a camo app change, the path added for #1794.

Previous form was `if (key != Prefs.PREF_BRIDGE_COUNTRY) return`. Introduced in f51120b5. Wants `&&`.

Want the one line change with a test, or would you rather take it directly?

vbrekher:
Fixes #1808.

The Snowflake proxy preference listener returned early for every key because the two negative key checks were joined with `||`. This meant bridge-country changes never stopped Kindness mode in countries where it is disabled, and camo-app changes never refreshed the notification.

Use a small predicate with the correct `&&` semantics and add regression coverage for both relevant preference keys as well as unrelated/null keys.

Tests:
- regression test reproduced the current OR-guard failure before the fix
- `./gradlew :app:testFullpermDebugUnitTest`
- `./gradlew :app:assembleFullpermDebug`
- `git diff --check`

### DIFF
--- a/app/src/main/java/org/torproject/android/ui/kindness/SnowflakeProxyPreference.kt
+++ b/app/src/main/java/org/torproject/android/ui/kindness/SnowflakeProxyPreference.kt
@@ -1,0 +1,6 @@
+package org.torproject.android.ui.kindness
+
+import org.torproject.android.util.Prefs
+
+internal fun shouldIgnoreSnowflakePreferenceChange(key: String?): Boolean =
+    key != Prefs.PREF_BRIDGE_COUNTRY && key != Prefs.PREF_CAMO_APP_PACKAGE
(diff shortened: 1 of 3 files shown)


bitmold:
This was a good check, the && over the || - thank you

bitmold:
Thanks again <3

PR #1811, opened 2026-09-09
Handle unavailable preference provider

### LINKED ISSUE #1760: [BUG] Log/Prefs Crash from Google Play
**Describe the Bug**
Logging related crash

**To Reproduce**
N/A

**Expected Behavior**
Don't crash

none

**Screenshots**
None

Version: 1795300402 (17.9.5-RC-4-tor-0.4.9.11)
 
- POCO POCO X4 Pro 5G
Android 13 (SDK 33)
-  Unknown device
Android 14 (SDK 34)
- Redmi Redmi 9C NFC
Android 10 (SDK 29)

**Crash Logs (Advanced)**
Type
java.lang.IllegalArgumentException
Generate crash insights with Gemini
Exception java.lang.IllegalArgumentException:
  at android.content.ContentResolver.update (ContentResolver.java:2399)
  at android.content.ContentResolver.update (ContentResolver.java:2365)
  at org.torproject.android.util.PreferenceProviderKt.putPref (PreferenceProvider.kt:146)
  at org.torproject.android.util.PreferenceProviderKt.putPref (PreferenceProvider.kt:151)
  at org.torproject.android.util.Prefs.orbotServiceLogAppend (Prefs.kt:457)
  at org.torproject.android.service.OrbotService.sendCallbackLogMessage (OrbotService.java:581)
  at org.torproject.android.service.OrbotService.logNotice (OrbotService.java:234)
  at org.torproject.android.service.OrbotRawEventListener.handleDebugMessage (OrbotRawEventListener.java:228)
  at org.torproject.android.service.OrbotRawEventListener.onEvent (OrbotRawEventListener.java:79)
  at net.freehaven.tor.control.TorControlConnection.handleEvent (TorControlConnection.java:240)
  at net.freehaven.tor.control.TorControlConnection.react (TorControlConnection.java:382)
  at net.freehaven.tor.control.TorControlConnection$ControlParseThread.run (TorControlConnection.java:352)

vbrekher:
Fixes #1760.

Preference reads and writes use a ContentProvider because OrbotService runs in the `:tor` process. If Android cannot resolve the provider URI, `ContentResolver` throws `IllegalArgumentException`; the service-log path currently lets that exception escape and crash the Tor control thread.

Treat that specific resolver failure as an unavailable preference provider: reads fall back to their existing defaults and writes become no-ops. Other runtime exceptions still propagate.

Tests:
- regression test reproduces the provider failure before the fallback
- `./gradlew :app:testFullpermDebugUnitTest`
- `./gradlew :app:assembleFullpermDebug`
- `git diff --check`

### DIFF
--- a/app/src/main/java/org/torproject/android/util/PreferenceProvider.kt
+++ b/app/src/main/java/org/torproject/android/util/PreferenceProvider.kt
@@ -106,17 +106,30 @@ class PreferenceProvider: ContentProvider() {
     }
 }
 
-private fun <T> ContentResolver.getPref(key: String, converter: (Cursor, Int) -> T?): T? {
-    val cursor = query(Uri.withAppendedPath(PreferenceProvider.CONTENT_URI, key),
-        null, null, null, null) ?: return null
-
-    cursor.use {
-        if (it.moveToFirst()) {
-            return converter(it, it.getColumnIndex(PreferenceProvider.ROW_VALUE))
-        }
+internal fun <T> preferenceProviderCall(defaultValue: T, action: () -> T): T =
+    try {
+        action()
+    } catch (_: IllegalArgumentException) {
+        defaultValue
     }
 
-    return null
+private fun <T> ContentResolver.getPref(key: String, converter: (Cursor, Int) -> T?): T? {
+    return preferenceProviderCall<T?>(null) {
+        val cursor = query(
+            Uri.withAppendedPath(PreferenceProvider.CONTENT_URI, key),
+            null, null, null, null
+        ) ?: return@preferenceProviderCall null
+
+        cursor.use {
+            if (it.moveToFirst()) {
+                return@preferenceProviderCall converter(
+                    it, it.getColumnIndex(PreferenceProvider.ROW_VALUE)
+                )
+            }
+        }
+
+        null
+    }
 }
 
 fun ContentResolver.getPrefString(key: String, default: String? = null): String? {
@@ -143,8 +156,13 @@ fun ContentResolver.getPrefFloat(key: String, default: Float? = null): Float? {
 }
 
 private fun ContentResolver.putPref(key: String, values: ContentValues) {
-    update(Uri.withAppendedPath(PreferenceProvider.CONTENT_URI, key),
-        values, null, null)
+    preferenceProviderCall(Unit) {
+        update(
+            Uri.withAppendedPath(PreferenceProvider.CONTENT_URI, key),
+            values, null, null
+        )
+        Unit
+    }
 }
 
 fun ContentResolver.putPref(key: String, value: String?) {
--- a/app/src/test/java/org/torproject/android/util/PreferenceProviderTest.kt
+++ b/app/src/test/java/org/torproject/android/util/PreferenceProviderTest.kt
@@ -1,0 +1,28 @@
+package org.torproject.android.util
+
+import org.junit.Assert.assertEquals
+import org.junit.Test
+
+class PreferenceProviderTest {
+    @Test
+    fun providerCallReturnsSuccessfulResult() {
+        assertEquals("value", preferenceProviderCall("fallback") { "value" })
+    }
+
+    @Test
+    fun providerCallFallsBackWhenProviderIsUnavailable() {
+        assertEquals(
+            "fallback",
+            preferenceProviderCall("fallback") {
+                throw IllegalArgumentException("Unknown URI")
+            }
+        )
+    }
+
+    @Test(expected = IllegalStateException::class)
+    fun providerCallDoesNotHideUnrelatedRuntimeFailures() {
+        preferenceProviderCall(Unit) {
+            throw IllegalStateException("unrelated failure")
+        }
+    }
+}


bitmold:
@n8fr8 @vbrekher my guess was that #1760 was happening on some devices when Orbot is started on boot. 

Looking at the stack trace from #1760 we see:
```
java.lang.IllegalArgumentException
Generate crash insights with Gemini
Exception java.lang.IllegalArgumentException:
at android.content.ContentResolver.update (ContentResolver.java:2399)
at android.content.ContentResolver.update (ContentResolver.java:2365)
at org.torproject.android.util.PreferenceProviderKt.putPref (PreferenceProvider.kt:146)
at org.torproject.android.util.PreferenceProviderKt.putPref (PreferenceProvider.kt:151)
at org.torproject.android.util.Prefs.orbotServiceLogAppend (Prefs.kt:457)
at org.torproject.android.service.OrbotService.sendCallbackLogMessage (OrbotService.java:581)
at org.torproject.android.service.OrbotService.logNotice (OrbotService.java:234)
at org.torproject.android.service.OrbotRawEventListener.handleDebugMessage (OrbotRawEventListener.java:228)
at org.torproject.android.service.OrbotRawEventListener.onEvent (OrbotRawEventListener.java:79)
at net.freehaven.tor.control.TorControlConnection.handleEvent (TorControlConnection.java:240)
at net.freehaven.tor.control.TorControlConnection.react (TorControlConnection.java:382)
at net.freehaven.tor.control.TorControlConnection$ControlParseThread.run (TorControlConnection.java:352)
```

In `OrbotRawEventListener` we call the offending `handleDebugMessage(String, String) on these events  
```java
            case TorControlCommands.EVENT_DEBUG_MSG, TorControlCommands.EVENT_INFO_MSG,
                 TorControlCommands.EVENT_NOTICE_MSG, TorControlCommands.EVENT_WARN_MSG,
                 TorControlCommands.EVENT_ERR_MSG -> handleDebugMessage(keyword, data);
```

and only then, the bug could occur if the severity of the log message isn't DEBUG/debug
```java
 private void handleDebugMessage(String severity, String message) {
        if (severity.equalsIgnoreCase("debug")) Log.d("Orbot", severity + ": " + message);
        else mService.logNotice(severity + ": " + message);
    }
```

So that may be why this was hard to test/surfaced inconsistently.


In any case @vbrekher this was an excellent PR, and I appreciate how it handles this case in a general sense. I've merged this so we can test it in our forthcoming release...

bitmold:
@vbrekher I do wonder though with the offending call in #1760 - in that crash, the call to 
`at org.torproject.android.util.Prefs.orbotServiceLogAppend (Prefs.kt:457)` is a call to upate a value in the preference datastore. 

With your fix here, instead of crashing with an `IllegalArgumentException` an empty Kotlin `Unit` is returned:

```kt
private fun ContentResolver.putPref(key: String, values: ContentValues) {
    preferenceProviderCall(Unit) {
        update(
            Uri.withAppendedPath(PreferenceProvider.CONTENT_URI, key),
            values, null, null
        )
        Unit
    }
}

internal fun <T> preferenceProviderCall(defaultValue: T, action: () -> T): T =
    try {
        action()
    } catch (_: IllegalArgumentException) {
        defaultValue
    }
```

In this case, when we are able to call `update` without an exception, we return `Unit`. But if the exception is thrown, like this crash used to do, we'd still return `Unit`. So there's no good way to meaningfully differentiate a preference write that was successful from a no-op failed write that would've thrown the exception reported in #1760

As a quick starting point, I updated your `putPref` implementation to default with `null` 
```kt
private fun ContentResolver.putPref(key: String, values: ContentValues) {
    val result: Unit? = preferenceProviderCall(null) {
        update(
            Uri.withAppendedPath(PreferenceProvider.CONTENT_URI, key),
            values, null, null
        )
        Unit
    }
    if (result == null) {
        Log.e(PreferenceProvider.TAG, "Couldn't update preference $key")
    }
}
```
so at the very least, on a write, if that call to `update` was gonna throw the exception, there's a debug log and a way for us programmers to meaningfully hook into and respond to that case.


I'm curious if you have thoughts on an implementation of something like this for the corresponding call to  `getPref`, since in that case we return a value of type `T?` where `T` could be `Int`, `String`, etc ( the possible preference types). With reading preferences it gets paritcularly murky here, since a valid call to get a preference may return `null` if that value doesn't exist in the preferences store:
```kt

fun ContentResolver.getPrefString(key: String, default: String? = null): String? {
    return getPref(key) { c, i -> c.getStringOrNull(i) } ?: default
}

fun ContentResolver.getPrefBoolean(key: String, default: Boolean = false): Boolean {
    val v = getPref(key) { c, i -> c.getStringOrNull(i) } ?: return default

    return v == "true"
}

fun ContentResolver.getPrefInt(key: String, default: Int? = null): Int? {
    return getPref(key) { c, i -> c.getIntOrNull(i) } ?: default
}

fun ContentResolver.getPrefLong(key: String, default: Long? = null): Long? {
    return getPref(key) { c, i -> c.getLongOrNull(i) } ?: default
}

@Suppress("unused")
fun ContentResolver.getPrefFloat(key: String, default: Float? = null): Float? {
    return getPref(key) { c, i -> c.getFloatOrNull(i) } ?: default
}
```

It'd be nice if, on a preference read, we could meainginfully distinguish between a `getPref` call that returns `null` (when the preference value doesn't exist) VS a call to `getPref` that returns `null` because an `IllegalArguementException` was thrown.  

I'd be curious to hear if you have any thoughts surrounding this, and If you're motivated to adding some kind of result type here, that could be really helpful when debugging future would-be crashes.

I'd imagine you could use some type of result type, wherein there's a specific value returned when the `IllegalArguementException` is thrown (or in the case of `cursor == null`) - this indicates that the call _didn't work_, and then a success type, which is your nullable `T?` (`String?`, `Int?`, `Long?`, etc...). This indicates conclusively that if a `getPref` call returns `null` it's because the preference doesn't exist, instead of today where `null` ambiguously means the pref doesn't  exist ***or*** that it _might_ exist but we weren't able to learn of it's (non)existence since we handled an `IlleglaArguementException`

PR #1812, opened 2026-09-10
Show configured proxy ports independently

### LINKED ISSUE #1134: UI: No ports are shown if HTTP port is not set
**Describe the Bug**
When the SOCKS port is set to a port, and the HTTP port is set to 0 (disabled), the ports are shown as "Proxy Ports : Ports Not Set"

**To Reproduce**
1. Open Orbot. It should not be running.
2. Go to settings. Set SOCKS to auto. Set HTTP to 0. Set transparent to 0.
3. Go back to the initial view. Start Orbot.
4. Orbot should have successfully started, and is properly routing user application data through Tor.
5. Go to the view that shows the open ports. By default this might say "SOCKS: 9050 HTTP: 8118."
6. Observe that it says no ports are open.

**Expected Behavior**
It should show only the SOCKS port with it's value. The HTTP port is not open.

- OS: Android 14

vbrekher:
Fixes #1134.

The port callback was only saved and displayed when both SOCKS and HTTP ports were enabled. With HTTP disabled (`0`), a valid SOCKS port was therefore shown as unset.

Store and render the two ports independently so disabling one proxy does not hide the other.

Adds regression coverage for the one-port-disabled cases.

### DIFF
--- a/app/src/main/java/org/torproject/android/OrbotActivity.kt
+++ b/app/src/main/java/org/torproject/android/OrbotActivity.kt
@@ -282,11 +282,11 @@ class OrbotActivity : BaseActivity() {
                 }
 
                 OrbotConstants.LOCAL_ACTION_PORTS -> {
-                    val socks = intent.getIntExtra(OrbotConstants.EXTRA_SOCKS_PROXY_PORT, -1)
-                    val http = intent.getIntExtra(OrbotConstants.EXTRA_HTTP_PROXY_PORT, -1)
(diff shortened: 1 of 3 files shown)


bitmold:
This was also great @vbrekher - I always deeply appreciate people clearing out some of our more longstanding (UI) issues. I made a few tweaks to your work in `ad313a6558cc570589148345af169217129b5091` but it's all just housekeeping things. 

Thanks so much!!!

PR #1814, opened 2026-09-13
Remove dead handler callback interface

meenbeese:


bitmold:
This was sitting there for almost a decade, nice catch !

PR #1815, opened 2026-09-14
Reenable Webtunnel bridges in the UI

meenbeese:
WebTunnel was removed from the UI because Moat advertised no default Webtunnel bridges. That is no longer true: the live `/moat/circumvention/settings` endpoint now serves webtunnel bridges (via `bridgedb`) — it's the first advertised transport for China and Iran, and it's also present for Russia. Verified against production during development.


bitmold:
@tladesignz @n8fr8 let's talk about this tomorrow

app/src/main/res/values/strings.xml
```diff
@@ -175,6 +175,8 @@
     <string name="bridges_via_email_subtitle">Cloaks your traffic. Gets around some Tor blocking. Requires an email sent from a Gmail or Riseup account.</string>
     <string name="bridge_meek_azure" translatable="false">Meek Azure</string>
     <string name="bridge_meek_azure_subtitle">Cloaks your traffic. Gets around some Tor blocking. May be very slow.</string>
+    <string name="bridge_webtunnel" translatable="false">WebTunnel</string>
+    <string name="bridge_webtunnel_subtitle">Tunnels your traffic through a web server. Gets around some Tor blocking. May be slow.</string>
```
bitmold, line 179: If we are using webtunnel, we may want to rewrite this ...

tladesignz:
Uh, nice catch! I'll look into this and update on the iOS side, too, if everything's fine.

bitmold:
> Uh, nice catch! I'll look into this and update on the iOS side, too, if everything's fine.

Yeah, very curious about what the scoop is. There's was a minor release earlier today 
https://gitlab.torproject.org/tpo/anti-censorship/pluggable-transports/webtunnel

app/src/main/res/values/strings.xml
bitmold: > WebTunnel is pluggable transport that attempt to imitate web browsing activities based on [HTTPT](https://censorbib.nymity.ch/#Frolov2020b).

that's from their readme, it be cool in some way to adopt that phrasing in some way, somehow conveying ""it cloakes your traffic by imitating web browsing activities"

tladesignz:
> > Uh, nice catch! I'll look into this and update on the iOS side, too, if everything's fine.
> 
> Yeah, very curious about what the scoop is. There's was a minor release earlier today https://gitlab.torproject.org/tpo/anti-censorship/pluggable-transports/webtunnel

Hm. With IPtProxy, we rely on Lyrebird to provide the Webtunnel implementation.

However, [tha still points to v0.0.3](https://github.com/tladesignz/IPtProxy/blob/master/IPtProxy.go/go.mod#L80).

tladesignz:
Aaaah... waitasec.

Sorry, this is all a misunderstanding. I looked into this now. Actually, nothing has changed.

All is fine, as it is.

There is the expectation, that, maybe, one day, MOAT will advertise a *builtin* webtunnel config. In that case, we would offer a direct *Webtunnel* option, which would use these offerings.

However, this is not the case:

```
GET https://bridges.torproject.org/moat/circumvention/builtin

{
    "meek": […],
    "meek-azure": […],
    "obfs4": […],
    "snowflake": […]
}
```

Anyway, *this does *not* mean*, that IPtProxy/Orbot currently do not support Webtunnel at all.

In fact, the code picks up Webtunnel bridge lines and adds them to the "custom" bridge line collection. If any Webtunnel bridge lines are used, Orbot will start the Webtunnel client and make Tor use it. (You can even mix them with the "custom" bridge type.)


To be clear: `/moat/circumvention/settings` provides Webtunnel bridge lines since a year or two, but only, if you tell them the "right" country. They're tagged `source=bridgedb`, which means, they're coming from a bucket which is provided (in slices) to users-in-need only and are constantly replaced, when it's noticed, that they're becoming blocked.

You can see the difference more clearly, when you use the `/moat/circumvention/defaults` endpoint, where `source=builtin` and `source=bridgedb` are mixed.


# To drive this home finally:

- All bridges which are `builtin` are available for everyone, and in Orbot they will be selectable by transport type.
- All bridges which come from `bridgedb` are available for certain users only, and in Orbot they will be mangled into the "custom" "bridge type". You can mix transports there and Orbot makes sure to start/stop all respective clients and configure Tor correctly.


This already happens as it is. The patch you provided messes up that logic, so we cannot merge it.

PR #1824, opened 2026-09-17
Encrypted settings

tladesignz:
This is quite a chunk to digest. And it's only the beginning! :-)

It uses the very latest version of `androidx.datastore:datastore-preferences` which allows encrypted settings using tink through `androidx.datastore:datastore-tink`.

-> datastore preferences is the latest "thing". Google thinks it's hot shit and all XML preferences are actually deprecated to some extent, also secure preferences.

This will automatically create a key in the secure element, which is ironically referenced using an XML preferences file.

That key is used to decrypt a file which contains a JSON-serialized settings store object.

Since this is all directed towards the latest Compose UI, the interface to that data is a `Flow` object. Hence all access is asynchronous.

To make this work with our old codebase without too much wild refactoring, I created accessors which block the current thread on access.

Additionally, I added a `suspend set` method, which can (and should) be used in asynchronous circumstances, where we already have them.

I already migrated all settings, which are purely accessed through normal code.

Additionally, I created `SecureEditTextPreference` and `SecureListPreference`, so we can slowly migrate any preference UX which is currently managed by `preferences.xml` and `SettingsPreferenceFragment`.

The final picture would be, that all settings UX is done through Compose UI which can make use of the data `Flow`ing around. (Meaning, UI updates get implicitly triggered by changes in the underlying data, async/await is fully honoured.)

And we should aim to get rid of the `Prefs` object and the `PreferenceProvider`, which means, that all items are actually 
stored in the encrypted `Settings` object instead of unencrypted `SharedPreferences`.

Although it is said that one should not use `runBlocking` it actually doesn't have noticeable impact. In fact, there shouldn't be much difference to what we did before where code was implicitly blocking on all IO regarding preferences. So, YOLO.

Let me know how you feel about this!


bitmold:
OK this is going to take me some days to get thru but will dialog with you here as we go.

I haven't looked too deeply at this work yet, but one initial thought comes to mind. 

To limit the blast radius of all these refactors can we keep our old prefs implementation and switch to using this new implementation for the socks proxy password and username. 

We can have less chaotic of a PR, verify that things work in a controlled way, and then migrate other preferences over. I'm very hesitant to change the behavior of Prefs across the whole app because its so integral to every aspect of the project. Its accessed across different processes and so on. There could be some gotchas that break things in ways we don't immediately realize. 

All in all tho this is exciting and I'm excited to review it.

tladesignz:
Honestly, no, I'm not too fond ripping out almost everything again. `Prefs` is still there for all the things which are written via `preferences.xml`. Stuff which is only written and read through `Prefs` already moved to `Settings`.

It's not *that* many items. We can smoke test all of them. If there are subtle differences, you'll only notice them in live action anyway. Get it over with, instead of dragging this out.

Besides, since everything is synchronised, there should be no race condition surprises.

But, hey, we could ask some AI agent to look it over.

bitmold:
How does this implementation handle an existing setting the user has set in shared prefs? With this new settings API are old settings values respected ?

meenbeese:
Thank you for this PR @tladesignz. I'm glad that we're finally able to move to more modern technologies for this app. For now, the first thing I noticed that you didn't implement proper suspend styles APIs for the settings fields, which is a fair compromise for now. However, I think we should at least make the init function non-blocking as I have made here below. And in the future we should try to remove as much of the `runBlocking` calls as we can.

```diff
diff --git a/app/src/main/java/org/torproject/android/OrbotApp.kt b/app/src/main/java/org/torproject/android/OrbotApp.kt
index d5be0da7d..521f92f7a 100644
--- a/app/src/main/java/org/torproject/android/OrbotApp.kt
+++ b/app/src/main/java/org/torproject/android/OrbotApp.kt
@@ -9,7 +9,6 @@ import androidx.lifecycle.ProcessLifecycleOwner
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.runBlocking
 import org.torproject.android.localization.Languages
 import org.torproject.android.localization.LocaleHelper
 import org.torproject.android.service.circumvention.Transport.Companion.stateLocation
@@ -24,9 +23,7 @@ class OrbotApp : Application() {
     override fun onCreate() {
         super.onCreate()
 
-        runBlocking {
-            Settings.init(this@OrbotApp)
-        }
+        Settings.init(this@OrbotApp)
 
         // set state dir for IPtProxy
         try {
@@ -60,11 +57,12 @@ class OrbotApp : Application() {
 
         setLocale()
 
-        // this code only runs on first install and app updates
-        if (Settings.currentVersionForUpdate < BuildConfig.VERSION_CODE) {
-            // Don't do anything resource intensive here, instead set a flag to do the task later.
+        // One-time migration from the legacy SharedPreferences store, plus the first-install/update
+        // bookkeeping. Both run off the main thread so app startup isn't blocked by filesystem I/O.
+        CoroutineScope(Dispatchers.IO).launch {
+            Settings.migrate(this@OrbotApp)
 
-            CoroutineScope(Dispatchers.IO).launch {
+            if (Settings.currentVersionForUpdate < BuildConfig.VERSION_CODE) {
                 Settings.set(
                     currentVersionForUpdate = BuildConfig.VERSION_CODE,
 
diff --git a/app/src/main/java/org/torproject/android/util/Settings.kt b/app/src/main/java/org/torproject/android/util/Settings.kt
index 51ee6ce16..94e4b88a3 100644
--- a/app/src/main/java/org/torproject/android/util/Settings.kt
+++ b/app/src/main/java/org/torproject/android/util/Settings.kt
@@ -52,7 +52,7 @@ object Settings {
 
     private lateinit var dataStore: DataStore<SettingsStore>
 
-    suspend fun init(context: Context) {
+    fun init(context: Context) {
         val settingsFile = context.preferencesDataStoreFile(SETTINGS_FILE_NAME)
 
         try {
@@ -84,8 +84,6 @@ object Settings {
                 ReplaceFileCorruptionHandler { SettingsStoreSerializer.defaultValue },
                 produceFile = { settingsFile })
         }
-
-        if (dataStore.data.first().migrated < CURRENT_MIGRATION_STEP) migrate(context)
     }
 
     fun set(key: String, value: String) {
@@ -390,7 +388,9 @@ object Settings {
             }
         }
 
-    private suspend fun migrate(context: Context) {
+    suspend fun migrate(context: Context) {
+        if (dataStore.data.first().migrated >= CURRENT_MIGRATION_STEP) return
+
         val prefs = PreferenceManager.getDefaultSharedPreferences(context)
 
         dataStore.updateData { store ->
```

tladesignz:
> How does this implementation handle an existing setting the user has set in shared prefs? With this new settings API are old settings values respected ?

The core of the meat is, to no ones suprise, [Settings.kt](https://github.com/tladesignz/orbot/blob/encrypted_settings/app/src/main/java/org/torproject/android/util/Settings.kt).

I would strongly suggest to check my branch out and look at it in Android Studio, that is way easier than trying to figure out what's going on in the Github diff view.

It contains a migration on init:

https://github.com/tladesignz/orbot/blob/encrypted_settings/app/src/main/java/org/torproject/android/util/Settings.kt#L88

https://github.com/tladesignz/orbot/blob/encrypted_settings/app/src/main/java/org/torproject/android/util/Settings.kt#L393-L453

It will read the shared preferences, copy the values, which moved to `Settings` to the `SettingsStore` object and remove them from the shared preferences file.

I added migration versioning, so we are able to migrate one or more values later on, until shared preferences are finally done.

tladesignz:
> Thank you for this PR @tladesignz. I'm glad that we're finally able to move to more modern technologies for this app. For now, the first thing I noticed that you didn't implement proper suspend styles APIs for the settings fields, which is a fair compromise for now. 

There is already when you want to write them:

https://github.com/tladesignz/orbot/blob/encrypted_settings/app/src/main/java/org/torproject/android/util/Settings.kt#L114-L156

For reading, it would be a lot of boilerplate, which will stay unused for a very long time. And some might never, because they're used in a synchronous context, and changing that context to asynchronously handle everything would, in certain circumstances, make that code really ugly and complicated for no real gain.

So, I suggest to just change the respective accessor method of a specific setting, when it is really worthwhile.

> However, I think we should at least make the init function non-blocking as I have made here below. And in the future we should try to remove as much of the `runBlocking` calls as we can.

That actually adds a race condition I would seriously like to avoid.

Code accessing the settings is completely detached from each other, it is even in different processes. Trying to do the migration asynchronously *will* break things. On a standard phone, even a rather cheap one, it will not add much overhead. And it will only do it once. With every successive start, it's a no-op.

bitmold:
> I would strongly suggest to check my branch out and look at it in Android Studio, that is way easier than trying to figure out what's going on in the Github diff view.

I _always_ do, for every PR, just was looking for an explanation on this PR...

> It contains a migration on init:
> 
> https://github.com/tladesignz/orbot/blob/encrypted_settings/app/src/main/java/org/torproject/android/util/Settings.kt#L88
> 
> https://github.com/tladesignz/orbot/blob/encrypted_settings/app/src/main/java/org/torproject/android/util/Settings.kt#L393-L453
> 
> It will read the shared preferences, copy the values, which moved to Settings to the SettingsStore object and remove them from the shared preferences file.
> 
> I added migration versioning, so we are able to migrate one or more values later on, until shared preferences are finally done.

My point was that it would be helpful to have comments like this in the PR to begin with

app/src/main/java/org/torproject/android/ui/more/camo/CamoConfirmationDialogFragment.kt
```diff
@@ -27,8 +30,14 @@ class CamoConfirmationDialogFragment : DialogFragment() {
                 if (altIconValue != -1) key += altIconValue
                 val activePackageName = mapping[key]!!
                 Prefs.setCamoAppPackage(activePackageName)
```
bitmold, line 32: Can you please migrate `Prefs.setCamoAppPackage` to the new `Settings`. It doesn't make sense to migrate some of these camo fields but not all. 

That would then eliminate `isCamoEnabled` from `Prefs` and not have this situation where these values are tightly associated, but some are stored in the old `Prefs` and others are stored in the new `Settings`.

```kt
val isCamoEnabled: Boolean
        get() {
            val app = cr?.getPrefString(PREF_CAMO_APP_PACKAGE, DEFAULT_CAMO_DISABLED_ACTIVITY) ?: ""
            if (camoAppAltIconIndex != -1) return false
            if (Settings.camoAppAltIconIndex != -1) return false
            return app != DEFAULT_CAMO_DISABLED_ACTIVITY
        }
```

Plus, if you did this all in a commit, there'd be a clear reference commit on how to migrate over something from `Prefs` -> `Settings`

app/src/main/java/org/torproject/android/util/Settings.kt
```diff
@@ -0,0 +1,552 @@
...
+                // https://github.com/guardianproject/orbot-android/issues/1563
+                // https://www.rfc-editor.org/rfc/inline-errata/rfc3986.html
+                Pair(null, url.toString())
+            }
+        }
+
+    private suspend fun migrate(context: Context) {
+        val prefs = PreferenceManager.getDefaultSharedPreferences(context)
+
+        dataStore.updateData { store ->
```
bitmold, line 451: For each of these newly migrated `Setting`s, please define a constant somewhere at the top of the file away from the migration logic. 

As it stands right now, someone could fat-finger a change to one of these keys in such a way where it might not get caught. 

I especially want this, because these same keys are then written out a second time when you remove them from the old `Prefs` backing:

```kt
  dataStore.updateData { store ->
            store.copy(
                migrated = CURRENT_MIGRATION_STEP,
                smartConnect = prefs.getBoolean("pref_use_smart_connect", store.smartConnect),
                smartConnectTimeout = prefs.getInt(
                    "pref_smart_connect_timeout",
                    store.smartConnectTimeout
                ),
// ...
```

```kt
 prefs.edit {
            remove("pref_use_smart_connect")
            remove("pref_smart_connect_timeout")
            remove("pref_connection_pathway")
            remove("pref_bridges_list")
// ...
```

app/src/main/java/org/torproject/android/util/Settings.kt
```diff
@@ -0,0 +1,556 @@
...
+            dataStore = DataStoreFactory.create(
+                SettingsStoreSerializer,
+                ReplaceFileCorruptionHandler { SettingsStoreSerializer.defaultValue },
+                produceFile = { settingsFile })
+        }
+
+        if (dataStore.data.first().migrated < CURRENT_MIGRATION_STEP) migrate(context)
+    }
+
+    fun set(key: String, value: String) {
```
bitmold, line 118: Can this be renamed to be explicitly about setting/getting proxy configuration settings. I'm afraid someone might later use it when they're thinking of the other `set` function used for most every other `Setting`. If someone mistakenly tried to set a preference this way, there'd be no compile error / some future developer may waste a lot of time not realizing that their non-proxy-related preference write was a no-op

bitmold:
There are at least two big regressions here.


I Upgraded from having built in bridges set (using obfs4). Then, when I switched to the new settings code, I set direct connection. 

In the UI I see: 
 
image


But in my logs, `obfs4` is clearly still in the `torrc`, IPtProxy is running, etc. 
image


Further, you'll notice that in the screenshot above, the VPN key doesn't appear. Something with the `Prefs.putUseVpn()` migration seems to have not been implemented correctly etiher, at least when coming from a device that's been configured already.

Earlier I had suggested keeping this very simple:

> To limit the blast radius of all these refactors can we keep our old prefs implementation and switch to using this new implementation for the socks proxy password and username.

**_This_** is why I suggested starting small with just the few proxy preferences. It's easier to verify that things work/nothing regressed, in a controlled environment. 

> Honestly, no, I'm not too fond ripping out almost everything again. Prefs is still there for all the things which are written via preferences.xml. Stuff which is only written and read through Prefs already moved to Settings.
> 
> It's not that many items. We can smoke test all of them. If there are subtle differences, you'll only notice them in live action anyway. Get it over with, instead of dragging this out.

I'd wager that many of the other items may not have been migrated correctly either. We don't know, and we'll have to figure them all out...

To my earlier point though. You don't have to rip out almost everything again. You can make a new branch, implement a small isolated migration of the settings that we actually really want encrypted, make sure it works, then incrementally cherry-pick the myriad of other changes you introduced here for other settings.

app/src/main/java/org/torproject/android/OrbotApp.kt
bitmold, line 55: I suppose we also don't need this call to `Prefs.setContext` anymore if `Settings.init` calls it with the application context above

tladesignz:
> I _always_ do, for every PR, just was looking for an explanation on this PR...

Good!

> My point was that it would be helpful to have comments like this in the PR to begin with

Oh, you want more comments! Sure can do! Just say so.

app/src/main/java/org/torproject/android/ui/more/camo/CamoConfirmationDialogFragment.kt
tladesignz: Sure!

tladesignz:
Aaah, oh, shit. Your comments made me remember why `PreferenceProvider` was actually there. That's needed to work around dual process limitations.

Gotta rethink this completely.

Thank you for your support until here. I'll get back to you, when I fixed that.

